PR #34, opened 2019-07-22
catalog: support "ALTER TABLE ... DROP CONSTRAINT ..."

srenatus:
Before, this would result in

    column "venues_id_pkey" of relation "venues" does not exist


kyleconroy:
Thanks @srenatus!

Issue #7, opened 2019-07-02
String arrays


kyleconroy:
```sql
create table rooms (
            id varchar not null,
            members varchar[] not null default '{}'
);

-- name: GetMembersByRoom :one  
SELECT members
FROM rooms
WHERE id = $1;
```

srenatus:
Just to clarify, the following issue is related to _this issue_, isn't it? ⬇️ 

Right now, dinosql will generate this code for me:
```go
const deleteToken = `-- name: DeleteToken :exec
DELETE FROM chef_authn_tokens cat
WHERE cat.id=$1
AND projects_match(cat.project_ids, $2::TEXT[])
`

type DeleteTokenParams struct {
        ID string
        ID string
}

func (q *Queries) DeleteToken(ctx context.Context, arg DeleteTokenParams) error {
        _, err := q.db.ExecContext(ctx, deleteToken, arg.ID, arg.ID)
        return err
}
```
In the SQL statement, `$2` is supposed to be an array, and the generate function should be something like
```go
type DeleteTokenParams struct {
        ID string
        Foo []string
}

func (q *Queries) DeleteToken(ctx context.Context, arg DeleteTokenParams) error {
        _, err := q.db.ExecContext(ctx, deleteToken, arg.ID, pq.Array(arg.Foo))
        return err
}
```

kyleconroy:
@srenatus You are correct. To generate the correct method for the `DeleteToken` query, a few things need to be implemented:
- Support for the cast (`::`) operator
- Scanning into and from slices using `pq.Array`
- Generating the correct field name for the DeleteTokenParams struct

Once that's complete, your query should generate the following code:

```go
type DeleteTokenParams struct {
        ID string
        ProjectIDs []string
}

func (q *Queries) DeleteToken(ctx context.Context, arg DeleteTokenParams) error {
        _, err := q.db.ExecContext(ctx, deleteToken, arg.ID, pq.Array(arg.ProjectIDs))
        return err
}
```

Is the `projects_match` function a custom function? Or just a placeholder you're using?

srenatus:
> Is the projects_match function a custom function? Or just a placeholder you're using?

Yup, it's a custom SQL function deployed with the schema. Would dinosql need to know about that?

kyleconroy:
>  Would dinosql need to know about that?

[...]

PR #35, opened 2019-07-23
Initial PostgreSQL array support

kyleconroy:
This is a partial solution to #7, and should work for most basic cases.


srenatus:
Hooray. Thanks for pushing this all forward. 🎉

PR #36, opened 2019-08-01
Add support for cast operations

kyleconroy:
This change gets us closer to generating the ideal code for the query @srenatus posted in #7. Currently, the output looks like:

```go
const deleteToken = `-- name: DeleteToken :exec
DELETE FROM chef_authn_tokens cat
WHERE cat.id=$1
AND projects_match(cat.project_ids, $2::TEXT[])
`

type DeleteTokenParams struct {
        ID      string
        Dollar2 []string
}

func (q *Queries) DeleteToken(ctx context.Context, arg DeleteTokenParams) error {
        _, err := q.db.ExecContext(ctx, deleteToken, arg.ID, pq.Array(arg.Dollar2))
        return err
}
```

### DIFF
--- a/internal/catalog/build.go
+++ b/internal/catalog/build.go
@@ -379,3 +379,14 @@ func isNotNull(n nodes.ColumnDef) bool {
 	}
 	return false
 }
+
+func ToColumn(n *nodes.TypeName) pg.Column {
+	if n == nil {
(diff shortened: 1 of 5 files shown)


srenatus:
Thank you! That's a nice step forward. :smiley:

PR #43, opened 2019-08-02
capture CREATE FUNCTION statements into catalog

srenatus:
Trying to contribute a little myself here 😃 

- turned the `Funcs` field of `Schema` into a map, having a slice there was difficult to work with in `catalog.Update()`
- sprinkled a few TODOs that I think might need addressing for proper function support

Happy to contribute more, this is a fun project 😃

### DIFF
--- a/internal/catalog/build_test.go
+++ b/internal/catalog/build_test.go
@@ -325,6 +325,32 @@ func TestUpdate(t *testing.T) {
 				},
 			},
 		},
+		{
+			`
+			CREATE FUNCTION foo(bar TEXT) RETURNS bool AS $$ SELECT true $$ LANGUAGE sql;
+			CREATE FUNCTION foo(bar TEXT, baz TEXT) RETURNS TEXT AS $$ SELECT "baz" $$ LANGUAGE sql;
+			`,
+			pg.Catalog{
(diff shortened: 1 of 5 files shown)


internal/catalog/build_test.go
```diff
@@ -325,6 +325,26 @@ func TestUpdate(t *testing.T) {
...
 			},
 		},
+		{
+			`
+			CREATE FUNCTION foo(bar TEXT) RETURNS bool AS $$ SELECT true $$ LANGUAGE sql;
+			`,
+			pg.Catalog{
+				Schemas: map[string]pg.Schema{
+					"public": {
+						Enums:  map[string]pg.Enum{},
```
kyleconroy: For future tests, cmp.Diff is called with cmpopts.EquateEmpty(), so you don't need to create an empty map for the Enums and the Tables.

srenatus:
Thanks for the review. I'll get to it soon, hope this doesn't block you.

srenatus:
@kyleconroy I've updated the branch -- still changed the `Funcs` field, but to `map[string][]Function`, since alternatively, we'd have to go with `*[]Function` to allow updating the slice in `catalog.Update`. This seems like a decent compromise. What do you think? 😃

PR #51, opened 2019-08-08
catalog: differentiate functions with different argument types

srenatus:
This should get another small step closer to _nice_ argument struct field names. Only the recording bits to supported the added test cases, nothing more yet.


kyleconroy:
Thanks!

PR #68, opened 2019-08-19
A few quality of life improvements

### LINKED ISSUE #67: Quality of life improvements
- [x] sqlc.json: "name" in the config should be optional. if unspecified should just take the package basename
- [x] sqlc.json: "schema" should be able to be a file or directory
- [x] automatically create the output directory

kyleconroy:
- `name` in the package config is now optional
- return an error if package path is empty
- `schema` and `queries` can point to individual files
- the package path is automatically created if it does not exist

Fixes #67

### DIFF
--- a/README.md
+++ b/README.md
@@ -283,7 +283,7 @@ in the directory where the `sqlc` command is run.
 
 Each package document has the following keys:
 - `name`:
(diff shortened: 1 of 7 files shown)


kyleconroy:
cc @inconshreveable

PR #81, opened 2019-09-12
Update gen.go

### LINKED ISSUE #80: Not recognizing some types
The tool is unable to properly generate some structs, with the columns of the tables have the following types:

- character of any size, should map to string or array of rune (string is my favorite option);
- real, should map to float32;
- double precision, should map to float64
- numeric and decimal are tricky, so I don't think they should be mapped to any kind of float.

I will try to send a PR this weekend. In the meantime, I will investigate some other types that may be missed by the tool.

euller88:
Partially fixes #80


kyleconroy:
Thanks!

PR #95, opened 2019-10-04
internal/dinosql: Prepare() with no GoQueries still valid

### LINKED ISSUE #94: Prepare with no queries does not compile
If I don't have any queries, Prepare generates this:

```
func Prepare(ctx context.Context, db dbtx) (*Queries, error) {
	q := Queries{db: db}
	var err error
	return &q, nil
}
```

I understand this is invalid. But I'm just trying to test this out and it's not a great experience.

kevinburkemeter:
Assign the error value to _ if no queries exist, so the file still
compiles. This is an edge case so I'm not sure it's worth thinking
more about.

Fixes #94.

### DIFF
--- a/internal/dinosql/gen.go
+++ b/internal/dinosql/gen.go
@@ -688,7 +688,11 @@ func New(db dbtx) *Queries {
 {{if .EmitPreparedQueries}}
 func Prepare(ctx context.Context, db dbtx) (*Queries, error) {
 	q := Queries{db: db}
-	var err error{{range .GoQueries}}
+	var err error
+	{{- if eq (len .GoQueries) 0 }}
+	_ = err
+	{{- end }}
+	{{- range .GoQueries }}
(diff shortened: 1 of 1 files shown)


internal/dinosql/gen.go
```diff
@@ -688,7 +688,11 @@ func New(db dbtx) *Queries {
 {{if .EmitPreparedQueries}}
 func Prepare(ctx context.Context, db dbtx) (*Queries, error) {
 	q := Queries{db: db}
-	var err error{{range .GoQueries}}
+	var err error
+	{{- if eq (len .GoQueries) 0 }}
+	_ = err
+	{{- end }}
+	{{- range .GoQueries }}
```
kevinburkemeter: With the hyphen at the front I'm pretty sure you don't need this on the same line as `var err error`.

PR #109, opened 2019-10-23
Bytea support

euller88:
Support to bytea added, passed the test in gen_test.go, not sure if it will work in all situations, please review


kyleconroy:
Thanks @euller88. I'm also not sure if this will work in all situations, but having this preliminary support seems better than no support at all.

PR #124, opened 2019-11-04
wrap prepare errors with query name for easier debugging

inconshreveable:
Without this, you'll get prepare errors but have no idea which query failed to prepare.


kyleconroy:
thanks!

PR #127, opened 2019-11-05
gen: Add Close method to Queries struct

### LINKED ISSUE #101: add Close method
At the very least we can Close all of the prepared Stmt's, if that is something we are doing.

kyleconroy:
When using prepared queries, generate a method to close all the prepared
queries.

Fixes #101


kyleconroy:
After writing this, I thought of another way to expose this functionality. We could add a `Stmts` method to the Queries struct that returns a `[]*sql.Stmt`. I think that's a bit too much, so I'm just going to go with the `Close` method for now.

PR #138, opened 2019-11-28
proper generated code to handle byte slices

### LINKED ISSUE #137: Scanning fails for bytea into []byte
When scanning and from bytea into []byte, the generated code is unable to properly read  or write the data. This is due to the fact that the generated code for slices in general(wrapping the slices in the pq.Array() function), does not work for bytea when reading, and probably (not tested) is not writing the correct data.
I already know how to solve this, and I'm working on a PR.

euller88:
fixes #137

### DIFF
--- a/internal/dinosql/gen.go
+++ b/internal/dinosql/gen.go
@@ -95,14 +95,14 @@ func (v GoQueryValue) Params() string {
 	}
 	var out []string
 	if v.Struct == nil {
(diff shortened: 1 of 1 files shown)


kyleconroy:
@joshuarubin this should fix your first issue

joshuarubin:
looks like there were still log statements in it?

kyleconroy:
Yep, fixing that in #146

PR #141, opened 2019-12-05
Add a few types, export DBTX

joshuarubin:
This change includes several enhancements:
1. Add the `blob` type
2. Add the `inet` type and import for the `"net"` package as necessary
3. Fixes the "uses" functions so that they work with array columns
4. Export the `DBTX` interface. The unexported type didn't play well with [wire](https://github.com/google/wire) which needs concrete types which requires them to be exported. That said, since this interface is always the same, maybe there should be a type directly exported by sqlc, e.g. `"github.com/kyleconroy/sqlc.DBTX"` instead.
5. Fixes some minor lint issues

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,19 +203,19 @@ func (q *Queries) ListAuthors(ctx context.Context) ([]Author, error) {
 	return items, nil
 }
 
-type dbtx interface {
+type DBTX interface {
 	ExecContext(context.Context, string, ...interface{}) (sql.Result, error)
 	PrepareContext(context.Context, string) (*sql.Stmt, error)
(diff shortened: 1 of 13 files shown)


kyleconroy:
Thanks for the pull request. I knew that name looked familiar when I got the email about this PR :)

The changes look good. As for adding the interface to `sqlc`, I'd like to keep the generated code dependency-free most of the time. If it becomes an issue, we can discuss a setting that pulls in the type from another package.

joshuarubin:
Great, thanks for responding so quickly! I'm not certain if this contains all the changes I need for this project, so there might be more coming.

PR #144, opened 2019-12-11
[]byte isn't a pq.Array. Fix scanning into enums.

joshuarubin:
This fixes two issues:
1. Fields with type `[]byte` were being considered as `pq.Array(byte)`
1. Couldn't Scan into enum values. They are strings, but with a custom type, so I added the `sql.Scanner` interface to enum types.

### DIFF
--- a/internal/dinosql/gen.go
+++ b/internal/dinosql/gen.go
@@ -894,7 +894,8 @@ type {{.Ret.Type}} struct { {{- range .Ret.Struct.Fields}}
 {{end}}
 
 {{if eq .Cmd ":one"}}
-{{range .Comments}}//{{.}}{{end}}
+{{range .Comments}}//{{.}}
+{{end -}}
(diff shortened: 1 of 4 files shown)


kyleconroy:
@joshuarubin I'm surprised that enums were broken, as I use them in production in other code bases. Could you split out the enum change into a separate pull request?

A fix for the `[]byte` issue also exists in #138. It looks like there are more cases than you're covering here? Sadly #138 has logging statements. If you add the if check to a few more places I can merge this one.

joshuarubin:
The enum issue may only be because I am using an array of enums

joshuarubin:
Also, `Scan()` can panic if it's not passed a `[]byte`. I can have it return an error instead, but that requires importing `fmt` or `errors`, wasn't sure if it was worth the trouble. Thoughts?

PR #145, opened 2019-12-11
fix scanning into enums

joshuarubin:
The enum issue may only be because I am using a column with an array of enums.
Also, `Scan()` can panic if it's not passed a `[]byte`. I can have it return an error instead, but that requires importing `fmt` or `errors`, wasn't sure if it was worth the trouble. Thoughts?

### DIFF
--- a/internal/dinosql/gen.go
+++ b/internal/dinosql/gen.go
@@ -851,6 +851,11 @@ const (
 	{{.Name}} {{.Type}} = "{{.Value}}"
 	{{- end}}
 )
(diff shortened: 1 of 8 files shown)


joshuarubin:
OK, I've added a test that failed before this change.

kyleconroy:
Thanks! I know update the `db_test.go` file isn't the easiest. Really appreciate adding the tests.

joshuarubin:
Thanks!

PR #164, opened 2019-12-13
Avoid duplicate import statements

mrnugget:
Before this change it was possible to generate duplicate import statements.

Take this schema:

```json
  {
    "version": "1",
    "packages": [{
        "schema": "schema.sql",
        "queries": "queries.sql",
        "name": "queries",
        "path": "."
    }],
    "overrides": [
      { "column": "toy.owner_id", "go_type": "database/sql.NullInt64" },
      { "column": "toy.manufacturer_id", "go_type": "database/sql.NullInt64" }
    ]
  }
```

That would generate a `models.go` file with two imports for `database/sql`.

What the change here does is to make all imports, in the queries and the
models file, unique.

### DIFF
--- a/internal/dinosql/gen.go
+++ b/internal/dinosql/gen.go
@@ -195,22 +195,22 @@ func (r Result) Imports(settings PackageSettings) func(string) [][]string {
 }
 
 func (r Result) ModelImports() [][]string {
-	var std []string
+	std := make(map[string]struct{})
 	if r.UsesType("sql.Null") {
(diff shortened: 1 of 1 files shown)


kyleconroy:
Thanks! This does have a slight problem in that import order could change, as map key iteration is randomized. I'm going to open a follow-up pull request that sorts the imports before returning them.

mrnugget:
Thanks for merging and doing the follow-up fix!

PR #166, opened 2019-12-13
internal/catalog: Parse unnamed function arguments

kyleconroy:
Arguments to SQL functions are not required to have a name.

Also add a test for `DROP FUNCTION [IF EXISTS]`


kyleconroy:
🤦‍♂ I accidentally merged this before the tests, sorry

Issue #174, opened 2019-12-15
How to get start? I think it should be very straight forward but I only get panic error.


AllenDang:
Here is how I start.
1. go get github.com/kyleconroy/sqlc
2. cd to sqlc/cmd/sqlc
3. go build .
4. copy compiled sqlc to $GOPATH/bin

And then, I'm trying to run the sample.

1. Mkdir ~/Desktop/sqlctest
2. Run sqlc init and edit the sqlc.json in project folder.
```
{
  "version": "1",
  "packages": [
    {
      "name": "db",
      "emit_json_tags": true,
      "emit_prepared_queries": false,
      "path": "internal/db",
      "queries": "./sql/query/",
      "schema": "./sql/schema/"
    }
  ]
}
```
3. Create  folders in project folder like below.
![截屏2019-12-15下午1 57 46](https://user-images.githubusercontent.com/18550/70858748-e3e59d00-1f42-11ea-9827-ee8d2169387a.png)

4. Edit the author.sql.
```
CREATE TABLE authors (
  id   BIGSERIAL PRIMARY KEY,
  name text      NOT NULL,
  bio  text
);

-- name: GetAuthor :one
SELECT * FROM authors
WHERE id = $1 LIMIT 1;

-- name: ListAuthors :many
SELECT * FROM authors
ORDER BY name;
```
5. Run sqlc generate ./sql/query/author.sql
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x88 pc=0x42b6669]

goroutine 1 [running]:
github.com/kyleconroy/sqlc/internal/dinosql.ParseQueries(0xc00017c090, 0x480c471, 0x1, 0xc00011e300, 0x1, 0x4, 0x0, 0x0, 0x0, 0x0, ...)
        /Users/allendang/Documents/Gocode/src/github.com/kyleconroy/sqlc/internal/dinosql/parser.go:226 +0x689
github.com/kyleconroy/sqlc/internal/cmd.glob..func4(0x4820620, 0xc000148a60, 0x1, 0x1)
        /Users/allendang/Documents/Gocode/src/github.com/kyleconroy/sqlc/internal/cmd/cmd.go:156 +0x9aa
github.com/spf13/cobra.(*Command).execute(0x4820620, 0xc000148a30, 0x1, 0x1, 0x4820620, 0xc000148a30)
[...]

Issue #176, opened 2019-12-15
`sqlc generate` runs into nil pointer dereference


dominikbraun:
My `db.sql`:

```sql
CREATE TABLE authors (
  id   BIGSERIAL PRIMARY KEY,
  name text      NOT NULL,
  bio  text
);

-- ...
```
My `sqlc.json`:

```json
{
  "version": "1",
  "packages": [
    {
      "name": "db",
      "emit_json_tags": true,
      "emit_prepared_queries": false,
      "path": "db",
      "queries": "./db.sql",
      "schema": "./db.sql"
    }
  ]
}
```
Running `sqlc generate` causes a panic:

```shell
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x88 pc=0x6b9249]

goroutine 1 [running]:
github.com/kyleconroy/sqlc/internal/dinosql.ParseQueries(0xc00012ffb0, 0xe038f1, 0x1, 0xc000116300, 0x1, 0x4, 0x0, 0x0, 0x0, 0x0, ...)
        /home/runner/work/sqlc/sqlc/internal/dinosql/parser.go:226 +0x689
github.com/kyleconroy/sqlc/internal/cmd.glob..func4(0xdfaea0, 0xe2d4d8, 0x0, 0x0)
        /home/runner/work/sqlc/sqlc/internal/cmd/cmd.go:156 +0x9aa
github.com/spf13/cobra.(*Command).execute(0xdfaea0, 0xe2d4d8, 0x0, 0x0, 0xdfaea0, 0xe2d4d8)
...
```
The eror occurs here - [parser.go:221](https://github.com/kyleconroy/sqlc/blob/master/internal/dinosql/parser.go#L221):

```go
query, err := parseQuery(c, stmt, source)
if err != nil {
	merr.Add(filename, source, location(stmt), err)
	continue
}
query.Filename = filepath.Base(filename)
```
The code assumes that after checking for `err != nil` it is safe to access `query`. This is not correct, since `parseQuery` may also return `nil` for both the query and the error - [parser.go:384](https://github.com/kyleconroy/sqlc/blob/master/internal/dinosql/parser.go#L384):

```go
default:
	return nil, nil
```
The other cases do not match because the statement seems to be an instance of `pg_query.CreateStmt`.

[...]

PR #178, opened 2019-12-16
parseQuery: Return either a query or an error

dominikbraun:
This fixes https://github.com/kyleconroy/sqlc/issues/174 and https://github.com/kyleconroy/sqlc/issues/176.

`parseQuery` used to return a `nil` query and a `nil` error, however the code relies on the assumption that the query is always valid if the error is not `nil` (as explained [here](https://github.com/kyleconroy/sqlc/issues/176#issue-538122582)).

This assumption gets correct with this change. `parseQuery` returns either a query _or_ and error. There's no `return nil, nil` anymore.

### DIFF
--- a/internal/dinosql/parser.go
+++ b/internal/dinosql/parser.go
@@ -371,7 +371,7 @@ func parseQuery(c core.Catalog, stmt nodes.Node, source string) (*Query, error) 
 	}
 	raw, ok := stmt.(nodes.RawStmt)
 	if !ok {
-		return nil, nil
+		return nil, errors.New("node is not a statement")
 	}
 	switch n := raw.Stmt.(type) {
 	case nodes.SelectStmt:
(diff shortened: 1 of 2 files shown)


internal/dinosql/parser.go
```diff
@@ -382,7 +382,7 @@ func parseQuery(c core.Catalog, stmt nodes.Node, source string) (*Query, error)
 		}
 	case nodes.UpdateStmt:
 	default:
-		return nil, nil
+		return nil, fmt.Errorf("parseQuery: unsupported statement type: %T", n)
```
kevinburkemeter: Maybe add a note that here and above you should file an issue if you hit this error because it's easy to see someone generating code that (should be) valid but hitting an edge case here because we haven't thought about it.

dominikbraun: That makes sense because with https://github.com/kyleconroy/sqlc/pull/179, these edge cases shouldn't be hit at all. Now the user gets the following error if it happens though:

```
parseQuery: unsupported statement type: pg_query.CreateStmt - feel free to file an issue on GitHub
```

dominikbraun:
Done, the error messages are uniformly again now.

PR #179, opened 2019-12-16
Don't allow equal query and schema paths

dominikbraun:
At the moment (without PR https://github.com/kyleconroy/sqlc/pull/178), this config file causes sqlc to panic:

```json
{
  "version": "1",
  "packages": [
    {
      "name": "db",
      "emit_json_tags": true,
      "emit_prepared_queries": false,
      "path": "db",
      "queries": "db.sql",
      "schema": "db.sql"
    }
  ]
}
```

sqlc can't handle the situation where the `queries` value is equal to the `schema` value. 

While https://github.com/kyleconroy/sqlc/pull/178 fixes the panic from a technical point of view and prevents sqlc from panicking in other situations, this PR checks if the paths are the same and informs the user if necessary.

### DIFF
--- a/internal/cmd/cmd.go
+++ b/internal/cmd/cmd.go
@@ -127,6 +127,11 @@ var genCmd = &cobra.Command{
 		output := map[string]string{}
 
 		for i, pkg := range settings.Packages {
+			if pkg.Schema == pkg.Queries {
+				fmt.Fprintf(os.Stderr, "package[%d]: schema and query path must not be identical\n", i)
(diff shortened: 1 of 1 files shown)


kyleconroy:
See #180 for an alternative to this change

dominikbraun:
I'd prefer https://github.com/kyleconroy/sqlc/pull/180 allowing the same path/file.

kyleconroy:
Great, closing this in favor of merging #180

PR #191, opened 2019-12-17
internal/dinosql: Add COMMENT ON support

### LINKED ISSUE #187: Add schema comments to models.go
If I attach a comment to a table or column in the schema it should appear as a comment in the generated Go code.

```sql
CREATE TABLE foo ( bar INT );
COMMENT ON TABLE foo IS 'Foo is a table';
COMMENT ON COLUMN foo.bar IS 'Bar is a column';
```

```go
// Foo is a table
type Foo struct {
  // Bar is a column
  Bar int
}
```

kyleconroy:
Comments are supported for SCHEMA, TABLE, COLUMN, and TYPE.

Fixes #187

### DIFF
--- a/internal/catalog/build.go
+++ b/internal/catalog/build.go
@@ -370,6 +370,105 @@ func Update(c *pg.Catalog, stmt nodes.Node) error {
 			Arguments:  args,
 			ReturnType: join(n.ReturnType.Names, "."),
 		})
+
+	case nodes.CommentStmt:
+		switch n.Objtype {
+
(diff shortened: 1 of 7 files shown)


maxhawkins:
Related to #193:

This breaks if one of your comments is on a view.

```sql
CREATE TABLE users (name TEXT);

CREATE VIEW harrys AS
  SELECT name
  FROM users
  WHERE name ILIKE '%harry%';

COMMENT ON harrys.name IS 'Harry's name';
```

`relation "harrys" does not exist`

PR #198, opened 2019-12-18
internal/dinosql: Filter out invalid characters

### LINKED ISSUE #197: Fails when ':' is included in ENUM name
This enum declaration is valid sql:
```sql
CREATE TYPE enum AS ENUM ('a:');
```

But running it through sqlc produces this error:
```
error generating code: source error: 22:7: expected ';', found ':' (and 10 more errors)
```

kyleconroy:
Ensure that enum identifiers are valid

Fixes #197

### DIFF
--- a/internal/dinosql/gen.go
+++ b/internal/dinosql/gen.go
@@ -7,6 +7,7 @@ import (
 	"go/format"
 	"log"
 	"path/filepath"
+	"regexp"
 	"sort"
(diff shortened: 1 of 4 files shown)


kyleconroy:
After this change, it's possible to have two enum values that map to the same Go identifier. This will need to be fixed in a future pull request.

Issue #201, opened 2019-12-18
LIMIT, OFFSET don't work when subquery is present


maxhawkins:
```sql
CREATE TABLE foo (bar TEXT NOT NULL);

-- name: GetFoo :one
SELECT
    (SELECT 1) AS baz
FROM foo
WHERE
	bar = $1
LIMIT $2
OFFSET $3;
```

Produces
```go
type GetFooParams struct {
	Bar   string `json:"bar"`
	Bar_2 string `json:"bar_2"`
	Bar_3 string `json:"bar_3"`
}
```

kyleconroy:
#209 is a hot fix for this issue, but it doesn't fix it completely

kyleconroy:
This has been fixed on `main` and will be fixed in v1.10.0. The query now generates the following input parameters.

```
// Code generated by sqlc. DO NOT EDIT.
// source: query.sql

package querytest

import (
	"context"
)

const getFoo = `-- name: GetFoo :one
SELECT
    (SELECT 1) AS baz
FROM foo
WHERE
	bar = $1
LIMIT $2
OFFSET $3
`

type GetFooParams struct {
	Bar    string
	Limit  int32
	Offset int32
}

func (q *Queries) GetFoo(ctx context.Context, arg GetFooParams) (interface{}, error) {
	row := q.db.QueryRowContext(ctx, getFoo, arg.Bar, arg.Limit, arg.Offset)
	var baz interface{}
	err := row.Scan(&baz)
	return baz, err
}
```

PR #204, opened 2019-12-18
Generating correct LIMIT and OFFSET field names

dominikbraun:
Fixes https://github.com/kyleconroy/sqlc/issues/201.

A query like this caused incorrect field names for `LIMIT` and `OFFSET`:

```sql
-- name: GetFoo :one
SELECT
    (SELECT 1) AS baz
FROM foo
WHERE
	bar = $1
LIMIT $2
OFFSET $3;
```
The generated struct used to look like this:

```go
type GetFooParams struct {
	Bar   string `json:"bar"`
	Bar_2 string `json:"bar_2"`
	Bar_3 string `json:"bar_3"`
}
```

This happened because `LIMIT` and `OFFSET` are `nil` for the subquery but they've been assigned to the `paramSearch` though.

### DIFF
--- a/internal/cmd/cmd.go
+++ b/internal/cmd/cmd.go
@@ -127,11 +127,6 @@ var genCmd = &cobra.Command{
 		output := map[string]string{}
 
 		for i, pkg := range settings.Packages {
-			if pkg.Schema == pkg.Queries {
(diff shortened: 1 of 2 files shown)


dominikbraun:
Oops, looks like my previous commits from https://github.com/kyleconroy/sqlc/pull/178 were in my local history.

PR #209, opened 2019-12-19
Generating correct LIMIT and OFFSET field names

dominikbraun:
Fixes https://github.com/kyleconroy/sqlc/issues/201.

A query like this caused incorrect field names for `LIMIT` and `OFFSET`:

```sql
-- name: GetFoo :one
SELECT
    (SELECT 1) AS baz
FROM foo
WHERE
	bar = $1
LIMIT $2
OFFSET $3;
```
The generated struct used to look like this:

```go
type GetFooParams struct {
	Bar   string `json:"bar"`
	Bar_2 string `json:"bar_2"`
	Bar_3 string `json:"bar_3"`
}
```

This happened because `LIMIT` and `OFFSET` are `nil` for the subquery but they've been assigned to the `paramSearch` though.

### DIFF
--- a/internal/dinosql/parser.go
+++ b/internal/dinosql/parser.go
@@ -990,8 +990,12 @@ func (p *paramSearch) Visit(node nodes.Node) Visitor {
 		p.parent = node
 
 	case nodes.SelectStmt:
-		p.limitCount = n.LimitCount
-		p.limitOffset = n.LimitOffset
+		if n.LimitCount != nil {
+			p.limitCount = n.LimitCount
+		}
+		if n.LimitOffset != nil {
+			p.limitOffset = n.LimitOffset
+		}
 
 	case nodes.TypeCast:
 		p.parent = node


kyleconroy:
@dominikbraun Thanks for the pull request. I think the existing limit / offset code needs to be totally rewritten. For example I don't think the following query will compile correctly:

```sql
-- name: GetFoo :one
WITH cte as (
    SELECT foo FROM bar LIMIT $1
)
SELECT
foo
FROM cte
OFFSET $2;
```

dominikbraun:
Yap, and while `(SELECT 1) AS baz` works as a subquery, `(SELECT 1 FROM foo) AS baz` runs into an error:

```
column reference "bar" is ambiguous
```

You may merge this PR as a temporary hotfix, or just rewrite the affected code from scratch.

kyleconroy:
> You may merge this PR as a temporary hotfix

Since I'm not sure about the long-term fix, adding a hot fix seems like a fine approach.

PR #212, opened 2019-12-23
internal/dinosql: Implement advisory locks

### LINKED ISSUE #211: Acquire advisory lock
I'm trying to add this to a query file:

```
-- name: AcquireLock :exec
SELECT pg_advisory_xact_lock($1);
```

However, I get the following error:

```
 $ make sql
/Users/kevin/bin/sqlc generate
# package db
db/query/locks.sql:1:1: nodes.ResTarget has nil name
```

It would be nice if this could generate a function call for me.

kyleconroy:
Fixes #211

### DIFF
--- a/internal/dinosql/gen.go
+++ b/internal/dinosql/gen.go
@@ -589,6 +589,11 @@ func (r Result) goInnerType(col core.Column) string {
 	case "inet":
 		return "net.IP"
 
+	case "void":
+		// A void value always returns NULL. Since there is no built-in NULL
+		// value into the SQL package, we'll use sql.NullBool
+		return "sql.NullBool"
+
 	case "any":
 		return "interface{}"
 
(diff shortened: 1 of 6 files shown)


kyleconroy:
This still isn't perfect, mainly due to the void return type

```
$ sqlc generate
2019/12/22 19:12:29 unknown Postgres type: void
2019/12/22 19:12:29 unknown Postgres type: void
2019/12/22 19:12:29 unknown Postgres type: void
```

kyleconroy:
Okay, it's looking good now

```go
const acquireLock = `-- name: AcquireLock :exec
SELECT pg_advisory_xact_lock($1)
`

func (q *Queries) AcquireLock(ctx context.Context, key int64) error {
    _, err := q.db.ExecContext(ctx, acquireLock, key)                                                                                                                                                       
    return err
}
```

PR #215, opened 2019-12-25
dinosql/internal: Add lower and upper functions

### LINKED ISSUE #214: regression: "relation "lower" does not exist"
On tip, this query:

```
SELECT * FROM company_users WHERE company_slug = $1 AND email = LOWER($2);
```

generates the following error:

```
db/query/company_users.sql:9:1: relation "lower" does not exist
```

This did not generate an error with previous versions of sqlc.

kyleconroy:
Fixes #214 

Adding support for advisory locks in #212 broke code generation for the queries that used functions with parameter references. This change restores some of that functionality. The only difference is the name of the field on the parameter struct.

```sql
SELECT * FROM company_users WHERE company_slug = $1 AND email = LOWER($2);
```

used to generate

```
type Params struct {
    CompanySlug string
    Email string
}
```

but will now generate

```
type Params struct {
    CompanySlug string
    Lower string
}
```

I know that's not ideal, but I thought it would be better to provide a hot fix.

### DIFF
--- a/internal/dinosql/gen.go
+++ b/internal/dinosql/gen.go
@@ -577,7 +577,7 @@ func (r Result) goInnerType(col core.Column) string {
 		}
 		return "sql.NullTime"
 
-	case "text", "pg_catalog.varchar", "pg_catalog.bpchar":
+	case "text", "pg_catalog.varchar", "pg_catalog.bpchar", "string":
 		if notNull {
 			return "string"
 		}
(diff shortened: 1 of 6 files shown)


landbed:
https://github.com/kyleconroy/sqlc/commit/ba4c9bfc0bfdcb1616ca5ef9b1299f35b313c254

@kyleconroy  this version with **postgresql** still have this issue, we have to using sqlc.arg like below to fix:

SELECT * FROM company_users WHERE company_slug =sqlc.arg(company_slug) AND email = LOWER(sqlc.arg(email));

PR #225, opened 2019-12-31
MySQL Support

### LINKED ISSUE #150: MySQL support
Love this project idea! All of my tools are built on mysql 5.x and it'd be great to have that supported.

cmoog:
This PR contains a near feature-complete implementation of MySQL support. Although it's missing a few features, I do think it is wise to merge to prevent more significant merge conflicts down the line. 

Most notably, the refactoring of the `dinosql/gen.go` file into parser agnostic Go code generation functions could potentially cause difficulty merging if we wait until MySQL support is fully stable.

Much of the remaining implementation work involves supporting all types and functions of the MySQL spec completely. This should not drastically alter the structure of the current implementation.

Critiques and contributions are welcome!

Fixes #150


internal/cmd/cmd.go
```diff
@@ -21,6 +22,7 @@ import (
 func Do(args []string, stdin io.Reader, stdout io.Writer, stderr io.Writer) int {
 	rootCmd := &cobra.Command{Use: "sqlc", SilenceUsage: true}
 	rootCmd.AddCommand(checkCmd)
+	rootCmd.AddCommand(unstable__mysql)
```
dan-compton: experimental

dan-compton:
Amazing.  I'm going to try this out.

kyleconroy:
🏆 🏆 🏆 

@cmoog thank you so much for this pull request! I didn't think MySQL support was going to land any time soon and I've been proven wrong. I'm still on Christmas / NYE vacation, so I won't get a chance to review this until January 2nd. I just wanted to let you know that I've seen the PR and will be reviewing it soon.

> Although it's missing a few features, I do think it is wise to merge to prevent more significant merge conflicts down the line.

Agreed! There's another open PR which I will wait to merge until this lands.

internal/dinosql/config.go
```diff
@@ -10,11 +10,27 @@ import (
...
 type GenerateSettings struct {
-	Version   string            `json:"version"`
-	Packages  []PackageSettings `json:"packages"`
-	Overrides []Override        `json:"overrides,omitempty"`
-	Rename    map[string]string `json:"rename,omitempty"`
+	Version    string            `json:"version"`
+	Packages   []PackageSettings `json:"packages"`
+	Overrides  []Override        `json:"overrides,omitempty"`
+	Rename     map[string]string `json:"rename,omitempty"`
+	PackageMap map[string]PackageSettings
```
kyleconroy: I'm not sure I understand the use case for `PackageMap`. Is it just for convenience?

internal/dinosql/gen.go
```diff
@@ -159,8 +180,15 @@ type GoQuery struct {
 	Arg          GoQueryValue
 }
 
-func (r Result) UsesType(typ string) bool {
-	for _, strct := range r.Structs() {
+type Generateable interface {
```
kyleconroy: As sqlc adds support for more database engines, I'd like this to b a concrete type instead of an interface. The `Query` struct generated by the parser should be general enough to use across engines. I understand that's not the case today.

kyleconroy:
@cmoog I've changed the pull request's base branch, it now merges into the `mysql` branch. I'm going to accept this PR as is and work on the configuration changes in that branch.

If you'd rather we work on this another way, please let me know. It's easy to do this differently.

internal/dinosql/config.go
kyleconroy: Answered my own question: The package map means we can pass around the settings struct by itself, instead of a global and package settings pair

cmoog: Exactly, my original thinking was that it would make lookups by package name easier, but I suppose the argument could be made that it made things a bit more indirect and error prone than they should be.

internal/dinosql/gen.go
cmoog: Great point. Although `generatable` worked as a POC, I do think it's appropriate to pay careful consideration to where the line of abstraction should lie. Good thinking with generalizing `Query` to be usable by multiple engines.

cmoog:
@kyleconroy Sounds good to me. Thanks for taking a look at the PR. Let me know how you'd like to proceed with improving the interface between parsing and code generation.  When you think it's appropriate, I'd be happy to make the necessary changes to the `mysql` package.

kyleconroy:
I've added a set of TODO items to #230. I don't think the mysql package will need to change too much before being merged in. The interface between parsing and code generation is a work in progress, so there's no point in letting this code sit around before it gets figured out.

PR #228, opened 2020-01-06
internal/dinosql: Add support for numeric types

### LINKED ISSUE #227: Numeric (p, s) type is not supported
Numeric type (p, s) (sometimes referenced as decimal) is not supported, despite being part of the standard postgresql syntax. 

https://www.postgresql.org/docs/9.3/datatype-numeric.html

kyleconroy:
Since the Go standard library does not have a decimal type, lib/pq returns decimals and numerics as strings. You'll need to use a decimal library to properly support these values.

https://github.com/lib/pq/issues/648

Fixes #227

### DIFF
--- a/internal/dinosql/gen.go
+++ b/internal/dinosql/gen.go
@@ -551,6 +551,14 @@ func (r Result) goInnerType(col core.Column) string {
 		}
 		return "sql.NullFloat64" // TODO: Change to sql.NullFloat32 after updating the go.mod file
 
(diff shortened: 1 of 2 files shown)


alexgille:
Hi, I just want to understand better the root cause of having to deal with strings. For example, why wouldn't a `float32`/`float64` be sufficient to convey `NUMERIC` and `DECIMAL` values?

PR #232, opened 2020-01-07
Set stdout correctly

akshayjshah:
The `cmd.Do` function takes a writer to use as standard out, but doesn't
use it. Instead, there's a duplicate call to `cobra.Command.SetErr`.
This patch replaces one of the duplicates with `SetOut`.


kyleconroy:
Thanks!

PR #237, opened 2020-01-08
examples: Add the jets and booktest examples

kyleconroy:
Move the end-to-end test into its own package: internal/endtoend
Add a new package for creating temporary PostgreSQL schemas:
internal/sqltest


kyleconroy:
Once this lands, the https://github.com/kyleconroy/sqlc-examples/ repository will be deleted.

PR #244, opened 2020-01-09
MySQL support improvements

cmoog:
This PR contains additional improvements to MySQL support

- now supports unqualified column expressions in the `SELECT` expression for the "left-most" table in the `FROM` expression. This is consistent with the MySQL specification.
- adds support for more MySQL datatypes and function return types
- adds logic to reuse `structs` when all fields are selected from a table

On a separate note, _please consider not squashing the commit history_. As the project grows and new collaborators seek to contribute, a rich git history along with its associated metadata is very helpful! I also personally find it incredibly useful as I navigate a project.

### DIFF
--- a/internal/dinosql/gen.go
+++ b/internal/dinosql/gen.go
@@ -52,7 +52,7 @@ func (gf GoField) Tag() string {
 }
 
 type GoStruct struct {
-	Table   Comparable
+	Table   core.FQN
 	Name    string
 	Fields  []GoField
 	Comment string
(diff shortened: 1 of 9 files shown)


jmhodges:
Are there tickets for what it fixes? I'm trying to figure if it fixes some bugs I've just run into. This change includes a lot of stuff and lot of it seems irrelevant to the things discussed (like the README changes?).

cmoog:
@jmhodges I apologize if the PR is difficult to follow, my intention was to add a few fixes that were left behind from the last merge. Given the instability of MySQL support, I figured a monolithic PR would be acceptable, but I suppose you're correct that each fix here should be separated and have its own ticket.

kyleconroy:
> On a separate note, please consider not squashing the commit history.

@cmoog I'm fine rebasing the commits onto master when I merge, but that will require commits being self-contained. It also means there should be no merge commits. In it's current form, I can't merge this PR without squashing commits, as it includes many commits which have already landed on master.

cmoog:
@kyleconroy Ok I see – squash away 😄 . I'll be more careful in the future.

internal/mysql/example/sqlc.json
```diff
@@ -2,8 +2,8 @@
   "version": "1",
   "packages": [
     {
-      "name": "teachersDB",
-      "database": "mysql",
+      "name": "db",
```
kyleconroy: This example should be moved into the top-level `examples` directory.

internal/dinosql/gen.go
```diff
@@ -52,7 +52,7 @@ func (gf GoField) Tag() string {
 }
 
 type GoStruct struct {
-	Table   Comparable
+	Table   core.FQN
```
kyleconroy: Very happy that the `FQN` type can be re-used.

👍

internal/mysql/example/sqlc.json
cmoog: Do you object to removing the example altogether? I'm thinking it's best to have consistent examples across engines. So the `booktest` works great for this purpose.

PR #246, opened 2020-01-09
log filename of unparsable mysql schema file

### LINKED ISSUE #245: mysql: generate command error doesn't include enough info to help debug
Running `sqlc` in mysql mode gets me the error:

```
$  sqlc generate
# package sqlcmodels
error parsing file: schema failure: syntax error at position 1362
```

But it's not clear what file wasn't parsed correctly.

jmhodges:
Updates #245

### DIFF
--- a/internal/mysql/parse.go
+++ b/internal/mysql/parse.go
@@ -43,7 +43,7 @@ func parsePath(sqlPath string, inPkg string, s *Schema, settings dinosql.Generat
 		}
 		queries, err := parseContents(filename, contents, s, settings)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("Failed to parse contents of file [%v]: %w", filename, err)
 		}
 		parsedQueries = append(parsedQueries, queries...)
 	}


internal/mysql/parse.go
```diff
@@ -44,7 +44,7 @@ func parsePath(sqlPath string, inPkg string, s *Schema, settings dinosql.Generat
 		}
 		queries, err := parseContents(filename, string(contents), s, settings)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("Failed to parse contents of file [%v]: %s", filename, err)
```
kevinburkemeter: Think you want to swap `%v` and `%s` here. (Also maybe use the new `%w` thing?)

jmhodges: That's up to @kyleconroy! I was using the same pattern of arguments from the errors right above it.

kevinburkemeter: Ah fair!

kyleconroy: Both of you are correct! sqlc requires Go 1.13, so we should be using `%w` for all wrapped error.

kyleconroy:
Thanks @jmhodges! Once you update` %s` to `%w I'll merge this in.

internal/mysql/parse.go
jmhodges: beep boop all done!

jmhodges:
For reasons, I also fixed the other places that could use %w in https://github.com/kyleconroy/sqlc/pull/250

PR #248, opened 2020-01-10
removes goose commands before parsing

### LINKED ISSUE #247: mysql: goose comment commands aren't removed from schema files
`RemoveRollbackStatements` is called in dinosql, but not in the mysql version of the parser, and so the mysql schemas are not aware of the goose commands.

cmoog:
Fixes #247


internal/mysql/example/queries.sql
```diff
@@ -1,3 +1,4 @@
+-- +goose Up
```
kyleconroy: This isn't how goose migrations work. There should be a single `Up` and a single `Down` comment per migration. You'll need to remove the second Up and Down comment.

kyleconroy:
Thanks for the quick fix. If you'd like me to merge your commits as-is, please rebase them into a single commit. Thanks!

cmoog:
> Thanks for the quick fix. If you'd like me to merge your commits as-is, please rebase them into a single commit. Thanks!

Ran into some git issues but I think everything should be good now unless I'm mistaken. Thanks for the guidance. 😄

PR #249, opened 2020-01-10
Improved MySQL errors

cmoog:
Addresses  #245 

## Example outputs:
```
❯ sqlc generate
# package booktest
booktest/mysql/query.sql:14:7: syntax error at or near 'SELEC'

❯ sqlc generate
# package booktest
booktest/mysql/query.sql:13:1: column "id" not found in table "books"
```

## Implementation notes

Note that the MySQL parser does not offer the same interface as the `pg` parser for inspecting the context of syntax errors. With this restraint, this PR uses regex matches on the `error` messages to parse the information. I don't feel great about the dependence on the contents of the `sqlparser` error message but I'm not sure how else to extract the required information...

### DIFF
--- a/internal/cmd/cmd.go
+++ b/internal/cmd/cmd.go
@@ -139,7 +139,13 @@ var genCmd = &cobra.Command{
 				q, err := mysql.GeneratePkg(name, pkg.Schema, pkg.Queries, settings)
 				if err != nil {
 					fmt.Fprintf(os.Stderr, "# package %s\n", name)
-					fmt.Fprintf(os.Stderr, "error parsing file: %s\n", err)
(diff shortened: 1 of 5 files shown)


jmhodges:
(Yeah, this seems required but maybe I missed something. I made a ticket with vitess so in the long-run, so we don't have to do this. https://github.com/vitessio/vitess/issues/5682 )

kyleconroy:
>  I don't feel great about the dependence on the contents of the sqlparser error message but I'm not sure how else to extract the required information...

Agreed! Matching errors via regex is gross, but it doesn't appear there is a better way currently. Also, thank you for adding tests.

Issue #261, opened 2020-01-12
type overrides: use actual package name instead of import name


kevinburkemeter:
If I have this:

```json
    "overrides": [
      {
          "go_type": "github.com/kevinburke/go-types.PrefixUUID",
          "postgres_type": "uuid"
      }
    ]
```

The generated Go source code is, e.g.

```
Get(ctx context.Context, id go-types.PrefixUUID) (Thing, error)
```

Which is a syntax error. In theory it would be good to determine the package import name "package types" and then use that instead.

kyleconroy:
Fixed partially by #262, closing for now

PR #262, opened 2020-01-12
internal/dinosql: strip leading "go-" or trailing "-go" from import

kevinburkemeter:
If I try to import from a package named "go-x" or "x-go" it will
generate a syntax error. We should do the right thing and determine
what the package name is and set it to that, but in lieu of that,
a good guess for the import name is just stripping that prefix or
suffix.

Updates #261.

### DIFF
--- a/README.md
+++ b/README.md
@@ -342,7 +342,7 @@ Each override document has the following keys:
 - `go_type`:
   - A fully qualified name to a Go type to use in the generated code.
 - `null`:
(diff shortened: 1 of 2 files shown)


kyleconroy:
> We should do the right thing and determine what the package name is and set it to that

Agreed

> A good guess for the import name is just stripping that prefix or suffix

Also agreed.

PR #266, opened 2020-01-13
improves MySQL query metadata parsing

cmoog:
## Current behaviour on `master` (defect)
```sql
/* naame: CreateAuthor :exec */
INSERT INTO authors (name) VALUES (?);
```
This query fails silently and does not generate a `go` query.

## Implementation Notes
Given the similarity of metadata syntax between `mysql` and `pg`, it seems appropriate to consolidate the logic.

### DIFF
--- a/internal/dinosql/parser.go
+++ b/internal/dinosql/parser.go
@@ -340,20 +340,35 @@ func validateQueryName(name string) error {
 		isLetter := unicode.IsLetter(c) || c == '_'
 		isDigit := unicode.IsDigit(c)
 		if i == 0 && !isLetter {
-			return fmt.Errorf("invalid query name: %q", name)
+			return fmt.Errorf("invalid query name %q", name)
 		} else if !(isLetter || isDigit) {
-			return fmt.Errorf("invalid query name: %q", name)
+			return fmt.Errorf("invalid query name %q", name)
 		}
 	}
 	return nil
 }
 
-func parseMetadata(t string) (string, string, error) {
+type CommentSyntax int
+
+const (
+	CommentSyntaxDash CommentSyntax = iota
+	CommentSyntaxStar               // Note: this is the only style supported by the MySQL sqlparser
+	CommentSyntaxHash
+)
+
+func ParseMetadata(t string, commentStyle CommentSyntax) (string, string, error) {
 	for _, line := range strings.Split(t, "\n") {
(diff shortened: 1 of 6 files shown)


kyleconroy:
MySQL [supports](https://dev.mysql.com/doc/refman/8.0/en/comments.html) three styles of comments:

>

cmoog:
When I first implemented this functionality, I ran into some issues with `vitess` support for `--` comments. I agree that supporting the same comment style is much preferred to the current proposal. Let me visit the parsing problem and see if I missed something.

cmoog:
It looks like the available methods exposed by `sqlparser` only support `/* */` comments for extraction. I'm not sure what the best option is from here... we could write out own custom "parser" for comments or use just use the comment style above. @kyleconroy, any thoughts?

internal/dinosql/parser.go
```diff
@@ -340,20 +340,34 @@ func validateQueryName(name string) error {
...
 
-func parseMetadata(t string) (string, string, error) {
+type CommentStyle int
+
+const (
+	MySQLCommentStyle CommentStyle = iota
+	PGCommentStyle
+)
+
+func ParseMetadata(t string, commentStyle CommentStyle) (string, string, error) {
```
kyleconroy: > It looks like the available methods exposed by sqlparser only support /* */ comments for extraction. 

That's fine for now. Can we add a comment to this function explaining that the MySQL parser doesn't support `--` and `#` comments?

internal/dinosql/parser.go
```diff
@@ -340,20 +340,34 @@ func validateQueryName(name string) error {
...
 		}
 	}
 	return nil
 }
 
-func parseMetadata(t string) (string, string, error) {
+type CommentStyle int
+
+const (
+	MySQLCommentStyle CommentStyle = iota
```
kyleconroy: Instead of naming these after the database engine, let's name them after the comment syntax

```go
type CommentSyntax int

const (
    CommentSyntaxDash CommentSyntax = iota
    CommentSyntaxStar
    CommentSyntaxHash
)
```

Also, note that the type name comes first in the constant name. This is consistent with the [HTTP status constants](https://golang.org/src/net/http/status.go) and the errors in the [os package](https://golang.org/pkg/os/#pkg-variables).

cmoog: Nice catch.

Issue #71, opened 2019-08-19
Customize parameter names


kyleconroy:
Certain queries end up with generic (read: terrible) names in parameter structs. For example

```sql
CREATE TABLE foo (bar text not null);

-- name: ListBar :many
SELECT bar FROM foo WHERE $1:bool;
```

There is no way to generate a good name for lone parameter in this query. Maybe another special comment? `-- param` could work.

```sql
-- name: ListBar :many
-- param: 1 IsTrue
SELECT bar FROM foo WHERE $1:bool;
```

maxhawkins:
+1 this would be helpful.

I encountered this situation:
```sql
CREATE TABLE foo (bar TEXT);

-- name: SetBar :exec
UPDATE foo
SET bar = $1::TEXT;
```

Casting `$1` to `TEXT` avoids a `sql.NullString` in the parameter of `SetBar`, but the parameter doesn't infer the name `bar`. It would be convenient if I could set it explicitly.

maxhawkins:
Adding named parameters a la HugSQL could solve this problem too:

```sql
-- name: ListBar :many
SELECT bar FROM foo WHERE :is_true::bool;
```

kyleconroy:
@maxhawkins I've avoided adding support for the HugSQL syntax because it's nonstandard. Passing that query to the PostgreSQL parser returns an error.

maxhawkins:
OK, agreed. Avoiding nonstandard syntax makes sense to me.

I want to add automatic SQL formatting to my query files and nonstandard syntax would break it.

kyleconroy:
[SQLAlchemy](https://docs.sqlalchemy.org/en/13/core/tutorial.html#using-textual-sql) supports named parameters in the form `:param`:

```python
stmt = text("SELECT * FROM users WHERE users.name BETWEEN :x AND :y")
stmt = stmt.bindparams(x="m", y="z")
```

[Sequel](http://sequel.jeremyevans.net/rdoc/files/doc/sql_rdoc.html), a Ruby database toolkit, supports the same format:

```ruby
DB.fetch("SELECT * FROM albums WHERE name LIKE :pattern", pattern: 'A%') do |row|
  puts row[:name]
end
```

While this format isn't part of the SQL standard, I think it solves the problem nicely.

maxhawkins:
psql [uses a similar format](https://stackoverflow.com/questions/36959/how-do-you-use-script-variables-in-psql/3588796#3588796) for its \set command:

[...]

PR #275, opened 2020-01-16
adds mysql support for custom `sqlc.arg(MyParam)` params

cmoog:
Adds functionality described in #71 

As is, the current implementation supports both `sqlc.arg(ParamName)` and `:ParamName`.

### DIFF
--- a/examples/booktest/mysql/query.sql
+++ b/examples/booktest/mysql/query.sql
@@ -43,6 +43,9 @@ WHERE book_id = ?;
 
 /* name: UpdateBookISBN :exec */
 UPDATE books
-SET title = ?, tags = ?, isbn = ?
+SET title = ?, tags = :book_tags, isbn = ?
 WHERE book_id = ?;
 
+/* name: DeleteAuthorBeforeYear :exec */
+DELETE FROM books
+WHERE yr < sqlc.arg(min_publish_year) AND author_id = ?;
--- a/internal/mysql/param_test.go
+++ b/internal/mysql/param_test.go
@@ -75,6 +75,16 @@ func TestSelectParamSearcher(t *testing.T) {
 				},
 			},
 		},
+		{
+			input: "select first_name, id FROM users LIMIT sqlc.arg(UsersLimit)",
+			output: []*Param{
+				&Param{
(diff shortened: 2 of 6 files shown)


examples/booktest/mysql/query.sql
```diff
@@ -46,3 +46,6 @@ UPDATE books
 SET title = ?, tags = ?, isbn = ?
 WHERE book_id = ?;
 
+/* name: DeleteAuthorBeforeYear :exec */
+DELETE FROM books
+WHERE yr < sqlc.arg(MinPublishYear) AND author_id = ? ;
```
kyleconroy: Can we change this to `min_publish_year` to make sure that the snake case to camel case conversion works?

internal/mysql/param_test.go
```diff
@@ -118,20 +128,20 @@ func TestInsertParamSearcher(t *testing.T) {
 
 	tests := []testCase{
 		testCase{
-			input: "INSERT INTO users (first_name, last_name) VALUES (?, ?)",
+			input: "INSERT INTO users (first_name, last_name) VALUES (?, sqlc.arg(UserLastName))",
```
kyleconroy: One thing I didn't mention in #71. I'm not sure we should allow mixing `?` and `sqlc.arg()`. Thoughts?

cmoog:
Just added the requests test changes. I also included an example in `booktest` for `:param`. Support for `:param` was included in the initial `mysql` PR.

kyleconroy:
Thoughts about [mixing parameter types](https://github.com/kyleconroy/sqlc/pull/275#discussion_r367528056)? I think it will only lead to confusion.

cmoog:
> Thoughts about [mixing parameter types](https://github.com/kyleconroy/sqlc/pull/275#discussion_r367528056)? I think it will only lead to confusion.

I can certainly imagine times when I'd want to only have one or two-argument name overrides in complex queries with 5+ arguments. In those cases, mixing `?` and `sqlc.arg` would be very convenient. With properly documented examples, I don't think it would be too confusing.

cmoog:
Just added proper error messages for the last four examples. The first example already returns a reasonably helpful message from the column schema validation.

PR #276, opened 2020-01-17
Named parameter support for PostgreSQL

### LINKED ISSUE #71: Customize parameter names

### LINKED ISSUE #189: Go Param Names in Complex Statements
I've got a query that looks something like this (though there are multiple columns being set):

```sql
-- name: UpdateFoo :one
UPDATE foo
SET
  bar = CASE WHEN $1::bool
    THEN $2::bar_type
    ELSE bar
    END
RETURNING *;
```

This produces a params struct like this:

```go
type UpdateFooParams struct {
	Column1         bool           `json:""`
	Column2_2       bar_type       `json:"_2"`
}
```

The naming here isn't very helpful. Do you have any suggestions on how this can be improved? Either by using a different kind of query (that allows updating multiple fields, each one optional) or by somehow annotating things such that the params struct uses more helpful naming?

kyleconroy:
This change supports for `sqlc.arg(name)` and `@name` parameter styles. It's powered by a new `postgresql/ast` package which support rewriting query ASTs.

This pull request is a work in progress and is not ready for review. Soon!

Fixes #71
Fixes #189


Cyberax:
Hi! Can I help with this PR?

PR #281, opened 2020-01-19
mysql type overrides

### LINKED ISSUE #273: add support for type overrides when using mysql
Creating this ticket to track support for type overrides with `mysql`. 

I think this will require changing the `postgres_type` field in `sqlc.json` into something that makes sense for both database engines. Possibly `db_type` instead?

cmoog:
Fixes #273 . 

This implements package-level and global type overrides for `mysql` usage.

# Implementation notes
I adjusted a few data structures to ease the passing of package settings and database schema throughout the parsing and generation process.
The primary new structures are shown below. 

```go 
type PackageGenerator struct {
	*Schema
	dinosql.CombinedSettings
	packageName string
}

type Result struct {
	PackageGenerator
	Queries []*Query
}
```
By separating the schema and package settings from the constructed queries, we're able to keep many of the parsing functions _more_ functionally pure.


internal/dinosql/config.go
```diff
@@ -59,43 +59,43 @@ type Override struct {
 	GoType string `json:"go_type"`
 
 	// fully qualified name of the Go type, e.g. `github.com/segmentio/ksuid.KSUID`
-	PostgresType string `json:"postgres_type"`
+	DbType string `json:"db_type"`
```
kyleconroy: This change is backwards incompatible and will break existing sqlc users. Configuration changes should be backwards compatible unless we're bumping the configuration version number[0]. There are a few options we have at our disposal.

- Add `db_type`, deprecate `postgres_type`, and complain when both are set. When `postgres_type` is set, we set it's value to `db_type`. 
- Add `mysql_type` and error `postgres_type` and `mysql_type` are both set.

I'm not very happy with either of those options. First, keep your first change, renaming `postgres_type` to `db_type`. However, we'll still allow `postgres_type` to be set as long as `db_type` is unset. 

We will need to add an optional `engine` field to global overrides. This field must be set when two different database engines are in use. Otherwise, we won't know which type to override.

[0] Changing the version number is actually more work because we want to be able to automatically upgrade older versions of the configuration files to new versions.

kyleconroy: `DbType` should be renamed to `DBType`

cmoog: Good thinking with the `engine` field for global overrides. I'll go ahead and make those changes now. Hopefully, these special cases won't make the documentation too dense.

examples/booktest/mysql/models.go
```diff
@@ -26,6 +28,7 @@ type Author struct {
 
 type Book struct {
 	BookID        int
+	BookUUID      uuid.UUID
```
kyleconroy: The booktest examples is taken from [xo/xo](https://github.com/xo/xo/tree/master/examples/booktest). The PostgreSQL and MySQL versions should be the same as that project.

Can you back out your multiple changes to this file and add tests for overrides in the `internal/endtoend/testdata` directory?

internal/dinosql/config.go
```diff
@@ -58,43 +59,78 @@ type Override struct {
...
+		}
+	}
+	return nil
+}
+
 func (o *Override) Parse() error {
+
+	// validate deprecated postgres_type field
+	if o.Deprecated_PostgresType != "" {
+		color.Yellow(`WARNING: "postgres_type" is deprecated. Instead, use "db_type" to specify a type override.`)
```
kyleconroy: These errors messages are great. Explicit and helpful.

I appreciate the thought of adding color output, but I'd like to do that in a single, holistic pass in the future. Can you just write this to stderr without color?

internal/dinosql/config.go
```diff
@@ -58,43 +59,78 @@ type Override struct {
...
-	table       pg.FQN
-	goTypeName  string
+	ColumnName  string
+	Table       pg.FQN
+	GoTypeName  string
 	goPackage   string
 	goBasicType bool
 }
 
+func (c *GenerateSettings) GatherEngines() map[Engine]struct{} {
```
kyleconroy: This function is only used in a single spot. Can you please inline into the `ValidateGlobalOverrides` function? I like to only break out functions into separate pieces when needed.

examples/booktest/mysql/models.go
cmoog: @kyleconroy How do you think we should organize the new `testdata` directory with multiple database engines?

kyleconroy: Individual tests should be in a subdirectory of testdata. Unless we're testing multi-engine configuration, we probably want to keep tests for each engine separate.

PR #285, opened 2020-01-22
internal/dinosql: Import needed types for Querier

### LINKED ISSUE #283: Emitted Querier interface does not include uuid import
I have a select query that uses a postgres uuid arg as the criteria.  The generated function signature on the Querier interface is correct but the emitted source file does not include the uuid package import.

kyleconroy:
Before this change, sqlc would generate a Querier interface without
importing the correct types. This is now fixed.

The `Querier` interface now lives in it's own file, querier.go.

Fixes #283


myles-mcdonnell:
We are now running this version in our build chain, all good +1

kyleconroy:
@myles-mcdonnell Great to hear, I'll merge it in now

PR #295, opened 2020-01-24
internal: Remove the PackageMap from settings

### LINKED ISSUE #293: Overrides are only applied for the last package defined in the configuration file
Given a configuration file with *multiple* packages, only the overrides for the *last package defined* will be overriden. In this example, `bar.a` and `bar.b` will be overridden but `foo.a` will not. If you delete the second package definition, then `bar.a` and `bar.b` will be overridden properly. Seems like somewhere some memory space is being re-used for overrides.

I've run git bisect and narrowed it down to commit 92ad2e33de2991703c8b8924684083ec5341be88 where MySQL support was first added

```json
{
        "version": "1",
        "packages": [
                {
                        ...
                        "overrides": [
                                {
                                    "go_type": "*A",
                                    "column": "foo.a"
                                }
                        ],
                },
                {
                        ...
                        "overrides": [
                                {
                                        "column": "bar.a",
                                        "go_type": "*A"
                                },
                                {
                                        "column": "bar.b",
                                        "go_type": "*B"
                                }
                        ]
                }
        ]
}
```

kyleconroy:
Create a new `CombinedSettings` struct that includes the global settings
and the current package settings.

Fixes #293

### DIFF
--- a/internal/dinosql/gen_test.go
+++ b/internal/dinosql/gen_test.go
@@ -1,7 +1,6 @@
 package dinosql
 
 import (
-	"path/filepath"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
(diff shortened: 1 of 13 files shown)


internal/dinosql/gen_test.go
```diff
@@ -94,33 +86,8 @@ func TestColumnsToStruct(t *testing.T) {
 	}
 }
 
-var mockSettings GenerateSettings
-
-func init() {
-	mockSettings = GenerateSettings{
```
kyleconroy: We should also remove the mockSettings global variable from the MySQL tests. Individual tests should instead create the necessary settings themselves.

PR #306, opened 2020-01-28
parser: optionally rewrite numbered params to positional params

mightyguava:
JDBC seems to only support positional parameters like `?`. I had to write a HACK to convert numbered parameters like $`1 `to `?`. This is buggy and terrible. It also means that you can't use `$1` multiple times in the same query to reference the same value.

This is a better change that examines the AST and edits the parser in the parsing step. It's cherry-picked from https://github.com/mightyguava/sqlc/pull/5 because I'd like to merge changes to core parsing code to master rather than risk complicated merge conflicts later.

When the option is off, nothing changes.

When the option is on, the parameters are no longer sorted by their `Number`, since we will rely on position as the binding index instead. Repeated parameters are no longer ignored. All parameters are included in the same order they appear in the query string. This means that there may be multiple `Columns` for the same query with the same `Number`. The generator can choose to de-dupe or not. Since this option will be tied to the generator, likely only Java/Kotlin generators, this is fine.

### DIFF
--- a/internal/dinosql/config.go
+++ b/internal/dinosql/config.go
@@ -51,6 +51,8 @@ type PackageSettings struct {
 	EmitJSONTags        bool       `json:"emit_json_tags"`
 	EmitPreparedQueries bool       `json:"emit_prepared_queries"`
 	Overrides           []Override `json:"overrides"`
+	// HACK: this is only set in tests, only here till Kotlin support can be merged.
+	rewriteParams       bool
 }
 
 type Override struct {
--- a/internal/dinosql/parser.go
+++ b/internal/dinosql/parser.go
@@ -229,7 +229,8 @@ func ParseQueries(c core.Catalog, pkg PackageSettings) (*Result, error) {
 			continue
 		}
 		for _, stmt := range tree.Statements {
-			query, err := parseQuery(c, stmt, source)
+			rewriteParameters := pkg.rewriteParams
+			query, err := parseQuery(c, stmt, source, rewriteParameters)
 			if err == errUnsupportedStatementType {
 				continue
 			}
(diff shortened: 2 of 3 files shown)


internal/dinosql/parser.go
```diff
@@ -958,7 +987,8 @@ type paramRef struct {
 type paramSearch struct {
 	parent   nodes.Node
 	rangeVar *nodes.RangeVar
-	refs     map[int]paramRef
+	refs     *[]paramRef
```
mightyguava: So I tried to make `Visit` take a pointer receiver: `(p *paramSearch)`, rather than doing a pointer to a slice here. This broke the `DeleteVenue` query parsing somehow. The parent for the second parameter because a `FunctionCall` instead of an `Expr`. I'm a compilers noob so I decided to not dive down that rabbit hole.

internal/dinosql/config.go
```diff
@@ -51,6 +51,8 @@ type PackageSettings struct {
 	EmitJSONTags        bool       `json:"emit_json_tags"`
 	EmitPreparedQueries bool       `json:"emit_prepared_queries"`
 	Overrides           []Override `json:"overrides"`
+	// HACK: this is only set in tests, only here till Kotlin support can be merged.
```
mightyguava: I promise to remove this when Kolint support is merged :)

mightyguava:
@kyleconroy what do you think about merging this PR? I split it out of the kotlin branch as it's in the parser code and I'm nervous about it running into more merge conflicts (it already has).

kyleconroy:
I believe this can be closed now that the Kotlin branch has been merged.

Issue #282, opened 2020-01-20
Support the PostgreSQL engine on Windows


dharmjit:
Hi, I just started exploring this project. But not able to build the project on windows. Is windows support in the road map. Thanks

euller88:
Can you specify the error message?

dharmjit:
I am getting below error.
```
github.com\lfittl\pg_query_go@v1.0.0\parser\include/port.h:17:19: fatal error: netdb.h: No such file or directory                                                                                                                                  compilation terminated.
```

euller88:
Ah, the dreaded "missing C header files" error. First, this seems like a problem within pg_query_go and the C header files available on your system. I have two recommendations for you:

1. if you are in Windows 10, activate WSL2, and use sqlc within the Linux subsystem. Be sure to install gcc on it.
2. Try the vendoring instructions present in the [pg_query_go repo](https://github.com/lfittl/pg_query_go#usage-with-go-modules), and see if it works.

I will research a little bit more in order to see if there is some other options, but those two are the ones that were on the top of my mind.

landbed:
@kyleconroy @euller88 how about providing a binary file on every releases? user can download directly?

kyleconroy:
@landbed that is already the case. You can download macOS and Linux binaries for every commit on [master](https://github.com/kyleconroy/sqlc#tip-releases). Each [release](https://github.com/kyleconroy/sqlc/releases/tag/v1.0.0) also has binaries.

The issue is that out PostgreSQL parser [does not support Windows](https://godoc.org/github.com/lfittl/pg_query_go). Once MySQL support is complete, we'll offer a Windows version that does not support PostgreSQL.

dharmjit:
Hi @kyleconroy, windows binaries then will not be useful for people using postgresql.

766b:
Just FYI, you can run `sqlc` on Windows by installing [Windows Subsystem for Linux](https://docs.microsoft.com/en-us/windows/wsl/install-win10) then using compiled Linux binary. Probably can add this to readme as temporary workaround for Windows users.

kyleconroy:
Initial Windows support is now live (#886). 

```bash
GOOS=windows GOARCH=amd64 go build ./cmd/sqlc/
```

These builds have a huge limitation: the PostgreSQL engine does not work. While that's obviously not ideal, it's good for those people that only need MySQL support.

arif2009:
[...]

PR #307, opened 2020-01-29
Attempt to build on Windows

kyleconroy:


kyleconroy:
https://github.com/kyleconroy/sqlc/issues/282

The naive approach of building on Windows does not work. I believe this has something to do with netdb.h being included instead of [winsock2.h](https://stackoverflow.com/questions/4432989/c-header-files-for-udp-in-windows).

Issue #302, opened 2020-01-26
Update configuration structure to support multiple languages


kyleconroy:
The current version of the configuration file can't support multiple languages. Packages have Go-specific settings that do not apply to other languages.

kyleconroy:
Here's an example configuration file using all the existing options, sans picking an engine. Notice that Go-specific information is contained in each section.

```json
{
  "version": "1",
  "packages": [
    {
      "name": "db",
      "path": "src/lib/db",
      "queries": "sql/db",
      "schema": "src/migrations",
      "emit_prepared_queries": true,
      "emit_json_tags": true,
      "overrides": [{
        "go_type": "github.com/foo/bar.Baz",
        "column": "foo"
      }]
    }
  ],
  "overrides": [{
    "go_type": "github.com/foo/bar.Baz",
    "postgres_type": "integer"
  }]
  "rename": {
    "sid": "SID"
  }
}
```

Here's a rough idea based on [uber/prototool](https://github.com/uber/prototool/blob/dev/example/proto/prototool.yaml). I'm not sure what to do about global overrides or renaming support.

```json
{
  "version": "1",
  "packages": [
    {
      "queries": "sql/db",
      "schema": "src/migrations",
      "plugins": {
        "go": {
          "name": "db",
          "path": "src/lib/db",
          "emit_prepared_queries": true,
          "emit_json_tags": true,
          "overrides": [{
            "go_type": "github.com/foo/bar.Baz",
            "column": "foo"
          }]
        }
      }
    }
  ]
}
```

kyleconroy:
It's also getting complicated enough that we may want to switch to YAML or TOML

mightyguava:
Switching to YAML for discussion:

```yaml
version: 1
packages:
  - queries: sql/db
    schema: src/migrations
    plugins:
      go:
        name: db
        path: src/lib/db
        emit_prepared_queries: true
        emit_json_tags: true
        overrides:
          - go_type: github.com/foo/bar.Baz
            column: foo
```

So I guess also generating Kotlin output would look like the below? Changing `go_type` => `gen_type` and `postgres_type` => `db_type` to allow the names to be consistent across backends/versions. Add maybe we add globals, also by engine, since it's unlikely the same overrides and renames will apply across languages.

```yaml
version: 2
packages:
  - queries: sql/db
    schema: src/migrations
    plugins:
      go:
        name: db
        out: src/lib/db
        emit_prepared_queries: true
        emit_json_tags: true
[...]

PR #315, opened 2020-01-31
internal/mysql: Port tests to endtoend package

kyleconroy:
As part of my work on #302, I wanted to avoid having to update the MySQL tests to use the new configuration format. When tests are written using the end-to-end framework, none of the internal structures leak into the tests.


kyleconroy:
@cmoog Care to take a look?

internal/endtoend/testdata/mysql_enums/go/models.go
```diff
@@ -0,0 +1,47 @@
...
+
+package querytest
+
+import ()
+
+type FirstNameType string
+
+const (
+	john   FirstNameType = "john"
+	albert FirstNameType = "albert"
```
kyleconroy: We're going to have to fix the naming here. These should be

```go
const (
	FirstNameJohn   FirstNameType = "john"
	FirstNameAlbert FirstNameType = "albert"
)
```

kyleconroy: It's fine for now, will address it in another PR.

cmoog: Good find. I must have written some unfortunately narrow tests that didn't catch that before.

cmoog:
Looks good to me 👍

PR #320, opened 2020-02-06
internal/dinosql: ignore golang-migrate rollbacks

treuherz:
We use [golang-migrate](https://github.com/golang-migrate/migrate) for migration management, and would like to be able to use it with sqlc. The obstacle here is that it doesn't indicate up/down migrations through special comments, but by naming the SQL files things like 
```
1481574547_create_users_table.up.sql
1481574547_create_users_table.down.sql
```
so I've had to put the check in `ReadSQLFiles` and test the functionality differently.

Let me know if you'd like any of this done differently, I'm happy to keep tweaking it.


kyleconroy:
@treuherz Thanks! Especially for the tests

treuherz:
Happy to contribute, thanks for the speedy merge!

Issue #161, opened 2019-12-13
SQLite support


RadhiFadlillah:
Hi all, great works on this project. Is it possible to add SQLite support ? Right now I'm working on embedded device and often uses SQLite, so it will be awesome for me.

kyleconroy:
Agreed! I'd love to add SQLite support.

The biggest blocker right now is access to a high-quality SQLite query parser. The SQLite C API doesn't expose a parsing function as far as I can tell, which means we'll probably need to create a project similar to [pg_query_go](https://github.com/lfittl/pg_query_go). 

We'll make this the official issue for tracking SQLite.

jhaungs:
Just a suggestion. The amalgamated sqlite source file can be used as a static or dynamic library. If you wanted to write to the C interface, you could get a parser for free if you're willing to track their development changes over time. Even if all you did was separate out the lemon parser generator and the generated parser, you'd have a leg up on writing your own, and a simpler path to future compliance with their changes.

kyleconroy:
I spent a few days looking into the feasibility of reusing the parser present in the sqlite source code. I don't think it's going to work. The parse.y file is littered with C / C++ calls, making using the existing parser very difficult. The internal structs are not cleaning exposed during parsing.

Instead, my current plan is to use the ANTLR, as they have an existing [SQLite](https://github.com/antlr/grammars-v4/blob/master/sql/sqlite/SQLite.g4) grammar. I played around with the generated parser today. It's not perfect, but it should fit the bill.

douglas-plezentek:
Hi @kyleconroy, has there been any progress on this?  I'm close to publishing build rules for sqlc in bazel, and it would be useful for writing and running tests that don't require a backend.

mvdan:
In case it helps, note that someone has a pretty advanced cgo-free implementation of sqlite at https://pkg.go.dev/modernc.org/sqlite. I seem to recall that it's translated from C, so the internal APIs will likely not be nicer than the C ones, but perhaps you can reuse some of the code at least.

The author is not on GitHub, but they are on Gitlab if you want to start a conversation: https://gitlab.com/cznic/sqlite

mvdan:
In particular, the internal parser seems to be at https://pkg.go.dev/modernc.org/sqlite/lib#Xsqlite3Parser.

kalafut:
[...]

PR #331, opened 2020-02-11
Add three new experimental engines, including SQLite

kyleconroy:
This will be a long-running pull request implementing SQLite support. SQLite support is being implemented using a different strategy than MySQL and PostgreSQL. Right now type checking, type inference, and code generation can't be shared among database engines.

For SQLite, the plan is use the parser (generated by ANTLR) to output a generic SQL AST. Think of this a form of compiler [IR](https://en.wikipedia.org/wiki/Intermediate_representation). The goal would then be to port both the MySQL and PostgreSQL engines to the same IR, and then do all (or most) of the work on that IR.

This is all a WIP and may fail horribly. I won't be changing any existing packages with this pull request.

#161


mightyguava:
> The goal would then be to port both the MySQL and PostgreSQL engines to the same IR, and then do all (or most) of the work on that IR.

+1 This seems much more maintainable than to try to use 3 different parsers with 3 different AST representations.

kyleconroy:
This is in a state where it's ready for a first set of eyes. The architecture of this change is more important than the specifics.

The best place to start is `internal/endtoend/parsers/parser_test.go`. Each database engine has a new package that exposes a `Parser` struct implementing this interface:

```go
type Parser interface {
    Parse(io.Reader) ([]ast.Statement, error)
}
```

An `ast.Statement` is the entry point into the new `internal/sql/ast` package. This ast package contains a generic set of SQL AST nodes. Each database parser is responsible for turning a query string into a slice of statements.

The new `internal/sql/catalog` package takes those statements and builds a catalog. It's very similar to the existing `internal/catalog` / `internal/pg` packages, but not specific to a single engine.

The `internal/sql/info` package was just a proof-of-concept, feel free to ignore it.

Again, all of this is subject to change, but I'm happy with the initial progress.

cc @mightyguava @cmoog

kyleconroy:
I'm happy enough with this progress that I'm going to merge this in. The experimental engines, prefixed with an underscore, are not enabled in releases. Once they get to a better state, we'll enable them for testing purposes.

kyleconroy:
I'm also going to rebase and split this change into concrete commits before merging

internal/sql/info/info.go
```diff
@@ -0,0 +1,59 @@
+package info
+
+import (
+	"github.com/kyleconroy/sqlc/internal/sql/catalog"
+)
+
+// Provide a read-only view into the catalog
+func Newo(c *catalog.Catalog) InformationSchema {
```
mightyguava: Newo?

kyleconroy: Oops, as I said, that package is experimental and not used.

mightyguava:
@kyleconroy any thoughts on my long comment above?

kyleconroy:
> I'm a bit concerned about the maintainability of the engines though. It seems like it'll be quite a lot of effort to adapt each parser's output to a single fully featured IR. 

So far this has not been the case. I've re-implemented the catalog using the new strategy and it's worked well.

> I wonder if there's a potential alternative (potentially insane) of using a single grammar that's a superset of all languages. 

The issue with a single grammar is that you have to write a filter per engine that knows which features are allowed for each engine. Right now this is handled by the individual parsers. For example, I have a generate `CreateTypeStmt` node. Since SQLite does not support `CREATE TYPE`, I don't need to do anything. The SQLite parser will fail to parse `CREATE TYPE` statements.

As I've done more work (https://github.com/kyleconroy/sqlc/pull/402, https://github.com/kyleconroy/sqlc/pull/401, #400, #399, #398, #397, #391) I'm confident in the current approach.

mightyguava:
Great to hear! I’d like to port the Kotlin code over for basic MySQL support as I have no use for it with PostgreSQL. Are the new parsers in a state for me to code against?

PR #336, opened 2020-02-12
config: Add support for YAML

### LINKED ISSUE #330: Add support for YAML
Now that version 2 of the configuration format has been merged, we should add support for reading from `sqlc.yaml`. The newer format is a bit of a chore to write in JSON.

kyleconroy:
Both version 1 and version 2 of the config formats can now be expressed
in YAML. sqlc will look for `sqlc.yaml` defined in the working directory.

Fixes #330


internal/config/config_test.go
```diff
@@ -46,7 +46,8 @@ func TestBadConfigs(t *testing.T) {
 		},
 		{
 			"unknown fields",
-			"json: unknown field \"foo\"",
+			`yaml: unmarshal errors:
```
kyleconroy: I'm not happy about this error changing. Going forward, we shouldn't be exposing an error message directory from a dependency.

PR #342, opened 2020-02-18
parser: Attach range vars to insert params

### LINKED ISSUE #340: a CTE problem?
the following code works just fine when executed on Postgres, however `sqlc generate` results in `column "code" does not exist`:
```sql
-- name: InsertCode :one
WITH cc AS (
    INSERT INTO td3.codes(created_by, updated_by, code, hash, is_private)
    VALUES ($1, $1, $2, encode(digest($2, 'sha1'), 'hex'), false)
    RETURNING hash
)
INSERT INTO td3.test_codes(created_by, updated_by, test_id, code_hash)
VALUES(
    $1, $1, $3, (select hash from cc)
)
RETURNING *;
```

kyleconroy:
Also process CTEs for insert statements

Fixes #340


internal/endtoend/testdata/insert_cte/go/query.sql.go
```diff
@@ -0,0 +1,51 @@
...
+)
+INSERT INTO td3.test_codes(created_by, updated_by, test_id, code_hash)
+VALUES(
+            $1, $1, $4, (select hash from cc)
+)
+RETURNING id, ts_created, ts_updated, created_by, updated_by, test_id, code_hash
+`
+
+type InsertCodeParams struct {
+	CreatedBy string
```
kyleconroy: This should be a `time.Time`, not a string.

kyleconroy: Ooops, I misread this code

PR #351, opened 2020-02-22
Allow sqlc.arg('argname') form for named params

Cyberax:
This pacifies static SQL analyzers in IDEs (like GoLand).


kyleconroy:
Thanks for the PR.

Can you explain your use of go run? I don’t understand why the Placeholder is needed.

internal/catalog/build.go
```diff
@@ -304,7 +304,9 @@ func Update(c *pg.Catalog, stmt nodes.Node) error {
 
 	case nodes.CreateSchemaStmt:
 		name := *n.Schemaname
-		if _, exists := c.Schemas[name]; exists {
+		// It's possible for the sqlc-internal scheme to be created by the user
+		// explicitly to allow static analysis of code with named params.
+		if _, exists := c.Schemas[name]; exists && name != "sqlc" {
```
kyleconroy: I think is better solved a different way. If the user needs to create the `sqlc` schema, they should do it in such a way that it does not conflict. Specifically:

```sql
CREATE SCHEMA IF NOT EXISTS sqlc; 
```

Cyberax:
> Can you explain your use of go run? I don’t understand why the Placeholder is needed.
My trick is to import codegen utilities like sqlc as modules. For this to work I need to make sure they are not optimized away by `go mod tidy`. This can be done by creating a dummy file (like `use.go`) that contains `import (_ "github.com/kyleconroy/sqlc")`. 

However right no sqlc can't be imported because Golang compiler detects that there are no Go files in the package ("github.com/kyleconroy/sqlc: module github.com/kyleconroy/sqlc@latest found (v1.0.0), but does not contain package github.com/kyleconroy/sqlc")

A simple placeholder file tricks Go into accepting the "package".

internal/catalog/build.go
Cyberax: Thanks!

I knew I should have read documentation first.

PR #358, opened 2020-02-24
internal/dinosql: support Postgres macaddr type in Go

kevinburkemeter:
Tests not passing, just wanted to submit


kevinburkemeter:
ok added

PR #361, opened 2020-02-25
Add support for named arguments in INSERTs

Cyberax:
This is the first draft of the named args support in INSERTs. 

But perhaps this "navigation expression" should be refactored into a full-blown structure and used instead of `cursor.name`.


kyleconroy:
Thanks for the patch. After looking into the bug, I settled on a different solution in #372. Thanks for finding the bug and submitting the initial fix.

PR #362, opened 2020-02-25
Add support for create-if-not-exists for 'create schema'

Cyberax:
Actually support "create schema if not exists sqlc"


kyleconroy:
Sorry about this! I realized I bit after I mentioned "CREATE SCHEMA IF NOT EXISTS" that the catalog would need to be updated. Thanks for the patch.

PR #368, opened 2020-02-28
kotlin: add Query class to support timeout and cancellation

mightyguava:
There's now a `sqlc.runtime` package containing a few abstract classes implemented by the each query method. The query methods now return these intermediate classes rather than executing the query directly. You have to now explicitly call `.execute()` on the return value. I don't really like this level of indirection, but it now makes it possible to set timeouts at the per-query level, and allows queries to be canceled. It also opens up the possibility of "interceptors" for query objects that can be used to apply additional common code.

The runtime library is at https://github.com/kyleconroy/sqlc-kotlin-runtime. I haven't published it to maven yet. It's kinda complicated and will take some time. So the runtime is checked in to this repo in the meantime to let tests pass.


mightyguava:
cc: @ryanhall07 @tirsen @swankjesse PTAL

mightyguava:
@kyleconroy fyi i'm setting up the artifact hosting, let me know if you want to make an account for yourself https://issues.sonatype.org/browse/OSSRH-55459

PR #391, opened 2020-03-10
sql/ast: Port support for COMMENT ON

kyleconroy:


internal/endtoend/testdata/comment_on/query.sql
```diff
@@ -0,0 +1,12 @@
+CREATE SCHEMA foo;
+
+CREATE TABLE foo.bar (
+        baz text NOT NULL 
+);
+
+CREATE TYPE foo.mood AS ENUM ('sad', 'ok', 'happy');
+
+COMMENT ON SCHEMA foo IS 'this is the foo schema';
```
kyleconroy: We don't put schema comments anywhere yet

Issue #365, opened 2020-02-27
Incorrect "ambiguous column reference" error


farwydi:
Input: (PostgreSQL)
```sql
-- name: FetchContentLocalization :many
SELECT c.id, --- Error here
       tx1.value AS name,
       tx2.value AS description,
       tx3.value AS text1,
       tx4.value AS text2,
       tx5.value AS text3,
       tx6.value AS text4
FROM (
         SELECT c.*
         FROM contents c
                  INNER JOIN content_group_links cgl ON c.id = cgl.content_id
                  INNER JOIN group_category_links gcl ON cgl.group_id = gcl.group_id
                  INNER JOIN categories ct ON gcl.category_id = ct.id
                  INNER JOIN services s ON ct.service = s.id
         WHERE s.id = 1
         UNION
         SELECT c.*
         FROM contents c
                  INNER JOIN content_includes ci ON c.id = ci.content_id
                  INNER JOIN categories ct ON ci.category_id = ct.id
                  INNER JOIN services s ON ct.service = s.id
         WHERE s.id = 1
     ) AS c
         LEFT JOIN override_content_params ocp ON ocp.content_id = c.id
         LEFT JOIN i18n_texts tx1 ON (tx1.key = c.name_i AND tx1.lang_id = 1)
         LEFT JOIN i18n_texts tx2 ON (tx2.key = c.description_i AND tx2.lang_id = 1)
         LEFT JOIN i18n_texts tx3 ON (tx3.key = c.text_field_1_i AND tx3.lang_id = 1)
         LEFT JOIN i18n_texts tx4 ON (tx4.key = c.text_field_2_i AND tx4.lang_id = 1)
         LEFT JOIN i18n_texts tx5 ON (tx5.key = c.text_field_3_i AND tx5.lang_id = 1)
         LEFT JOIN i18n_texts tx6 ON (tx6.key = c.text_field_4_i AND tx6.lang_id = 1)
WHERE c.id NOT IN
      (
          SELECT content_id
          FROM content_excludes ce
                   INNER JOIN categories ct ON ce.category_id = ct.id
                   INNER JOIN services s ON ct.service = s.id
          WHERE s.id = 1
      )
  AND (TRUE OR COALESCE(ocp.enable, c.enable) IS TRUE)
ORDER BY c.id;
```

Got:
localization.sql:2:8: column reference "id" is ambiguous

But is not true :)

### Playground link

[...]

PR #393, opened 2020-03-10
Failing test for failure to parse as

rf:
Copied an existing testcase and wrote a failing test. Turns out it happens when `as` is used in a subquery.

Re #365 as I'm pretty sure it's the same issue


kyleconroy:
Added the failing test to the issue (#365) via this [playground link](https://play.sqlc.dev/p/9359639085584325d3811942133ff00081d5eac367bd3fd9b80df2113c8ba136)

PR #425, opened 2020-03-30
Move hyphen-minus to the right place

rjc:
While there, don't terminate a list item with a full stop ('.').


kyleconroy:
Thanks for the typo fix!

rjc:
No worries! :^)

Issue #193, opened 2019-12-17
View support


maxhawkins:
```sql
CREATE TABLE foo (bar INT);

CREATE VIEW baz AS
  SELECT *
  FROM foo;
```

Should generate:
```go
type Foo struct {
  Bar int
}
type Baz struct {
  Bar int
}
```

Currently no structs are generated for views.

ereyes01:
I just found this package and it looks super useful. One of the first things I did was to search the issues to see if there was any insight on how views are handled, and found this.

Is the DDL code given ever actually run in the database? If it's only used to generate the structs, would a valid workaround be:

```sql
CREATE TABLE foo (bar INT);

CREATE TABLE baz (bar INT);
```

And then design your queries of `baz` as normal, since the DDL is only for deriving and generating the `Baz` Go struct?

I acknowledge that it's not as convenient or robust as copy/pasting the same thing you run in the database, and sometimes it's extra work to determine the equivalent `CREATE TABLE` command from a complex view query.

Thanks!

kyleconroy:
@ereyes01 That's a good work around for the time being. The eventual goal is to fully support view creation and querying.

randomlogin:
Hi!
I'm using 1.5.0, am I right, that views are not supported there yet? 
The only workaround is to construct appropriate table?

ScottyFillups:
Hi @kyleconroy! Really love sqlc; I'm wondering if there's any plans to tackle views any time soon? I'd willing to help out :smiley:

PR #426, opened 2020-04-01
A list of paths can be past in for schemas and queries

jskelcy:
This adds support for passing in a list of files or directories for queries and schemas.

This came about because the project currently recommends workarounds for SQL views, where users create a table definition that is the would-be table schema for the view. We did not want this file living alongside our production schema so I created an adjacent director called `workarounds`, this was going to make scripting fresh DB migrations much simpler as I did not have to regex out workaround files. However, because I could only pass in one path this did not work. This adds that support.

It is kind of massive and breaks config backwards compatibility but I can look into options for fixing that.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -2,6 +2,7 @@ package config
 
 import (
 	"bytes"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"go/types"
@@ -35,6 +36,39 @@ type versionSetting struct {
 
 type Engine string
 
+type Paths []string
+
+func (p *Paths) UnmarshalJSON(data []byte) error {
+	if string(data[0]) == `[` {
+		var out []string
+		if err := json.Unmarshal(data, &out); err != nil {
+			return nil
+		}
+		*p = Paths(out)
+		return nil
+	}
+	var out string
+	if err := json.Unmarshal(data, &out); err != nil {
+		return nil
+	}
+	*p = Paths([]string{out})
+	return nil
+}
+
+func (p *Paths) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	out := []string{}
+	if sliceErr := unmarshal(&out); sliceErr != nil {
+		var ele string
+		if strErr := unmarshal(&ele); strErr != nil {
+			return strErr
+		}
+		out = []string{ele}
+	}
+
+	*p = Paths(out)
+	return nil
+}
+
 const (
 	EngineMySQL      Engine = "mysql"
 	EnginePostgreSQL Engine = "postgresql"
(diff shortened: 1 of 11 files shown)


jskelcy:
So I think I can create a new type with a custom yaml and json umarshaler. I did most of this with find and replace but this is way too massive of a PR.

kyleconroy:
@jskelcy thanks for the patch!

Since this is a backwards-incompatible change, we'd need to bump the configuration file version. The current version is '2', though it's not in heavy use. Version '1' is what most people use.

That said, I think we can solve this in a backwards compatible way using glob patterns. 

```yaml
version: "1"
packages:
  - name: "db"
    path: "internal/db"
    schema: "./migrations/*"
    queries: "./sql/query/"
```

Here the migrations directory would have two directories in it, one with the production schema and one with the extra one for views.

Would that work for you?

Also I'm sorry that I haven't made any progress on #193, view support is still pretty far off

jskelcy:
@kyleconroy I made the change backward compatible let me know what you think.

jskelcy:
I can also mess around with the glob patterns, it might work out fine for me. But I figured I had already got this thing in some kind of shape, so why I would make it backwards compatible and see how we felt about it.

internal/config/config.go
```diff
@@ -35,6 +36,39 @@ type versionSetting struct {
 
 type Engine string
 
+type Paths []string
+
+func (p *Paths) UnmarshalJSON(data []byte) error {
```
jskelcy, line 41: I am not sure that I need this function. When the tests were failing they contained yaml parsing error even when the tests appeared to use json. Still trying to figure that out, but until I understand that there is some conversion or something going on then I will leave this here for now.

jskelcy:
Realizing now that while you suggested the glob approach to backwards compatibility, I went with a different approach but did not explicitly say what it is. 

I used a custom unmarshaler that accepts both a list or a single value. I suppose this is more flexible then the glob pattern because you could have SQL files anywhere and pull them in, but in reality I dont know if that is a need anyone has. If we dont think this is useful then I an implement the glob version, but I figured it is worth discussing.

jskelcy:
@kyleconroy Bumping the thread here. I have a backwards-compatible change where I use a custom unmarshaler that allows `queries` and `schema` to be a single string or a list. It is a little bit more flexible than using a wildcard, but I might be solving a problem that does not exist. Either way, let me know, I can try the wildcard strategy if you are less into this.

kyleconroy:
I was using docker compose a few days and encountered this pattern in the wild. `docker-compose.yml` has a field named `env_file` that be a string or a list of strings:

```
env_file: ./foo.env
---
env_file:
  - ./foo.env
  - ./bar.env
```

The glob syntax I proposed would not have helped me, so I think this is a good strategy.

internal/config/config.go
kyleconroy: We technically don't need this, as we use the yaml package to parse both JSON and YAML. However, I like keeping this so that we ensure that the configuration can still be parsed with the `encoding/json` package.

PR #433, opened 2020-04-05
Implement ALTER TYPE RENAME VALUE / ADD VALUE

### LINKED ISSUE #428: Renaming enum values via migration doesn't update the generated code
migration 0:  
```
DROP TYPE IF EXISTS e_my_enum_type CASCADE;
CREATE TYPE e_my_enum_type AS ENUM (
    'green',
    'red',
    'black'
    );
```

generated code for migration 0:  
```
type EMyEnumType string

const (
	EMyEnumTypeGreen EMyEnumType = "green"
	EMyEnumTypeRed   EMyEnumType = "red"
	EMyEnumTypeBlack EMyEnumType = "black"
)
```

**Problem**

migration 5:
`ALTER TYPE e_my_enum_type RENAME VALUE 'black' TO 'yellow';`

generated code after migration 5 is applied is the same as migration 0 for type `EMyEnumType`

Expected that it would be: 
```
type EMyEnumType string

const (
	EMyEnumTypeGreen EMyEnumType = "green"
	EMyEnumTypeRed   EMyEnumType = "red"
	EMyEnumTypeBlack EMyEnumType = "yellow"
)
```

kyleconroy:
This pull requests actually includes two parallel implementations of ALTER TYPE, one for the legacy `internal/catalog` package, and one for the new `internal/sql/catalog` package. I really need to port the `internal/dinosql` package over to the new catalog so I don't need to waste time doing things twice.

Fix #428

### DIFF
--- a/internal/catalog/build.go
+++ b/internal/catalog/build.go
@@ -99,6 +99,56 @@ func Update(c *pg.Catalog, stmt nodes.Node) error {
 
 	switch n := raw.Stmt.(type) {
 
+	case nodes.AlterEnumStmt:
+		fqn, err := ParseList(n.TypeName)
+		if err != nil {
+			return err
+		}
+		schema, exists := c.Schemas[fqn.Schema]
+		if !exists {
+			return wrap(pg.ErrorSchemaDoesNotExist(fqn.Schema), raw.StmtLocation)
+		}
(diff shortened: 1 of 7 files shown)


kyleconroy:
Fixes #428

PR #450, opened 2020-04-09
internal/dinosql: add map containing all enums

### LINKED ISSUE #447: Test for enum membership
An enum type in Postgres generates the following Go code:

```sql
CREATE TYPE company_role AS ENUM ('member', 'admin', 'guest');
```

```go
type CompanyRole string

const (
	CompanyRoleMember CompanyRole = "member"
	CompanyRoleAdmin  CompanyRole = "admin"
	CompanyRoleGuest  CompanyRole = "guest"
)
```

Given a random string, I would like to test whether it is a member of the enum or not without going to the database. I can create a function, but `s == CompanyRoleMember || s == CompanyRoleAdmin || s == CompanyRoleGuest` breaks without warning as soon as a new field was added to the enum.

Is there a way to provide a map, or `func ValidCompanyRole(s string) bool` or `func (c CompanyRole) Valid() bool`, or anything I could call to automatically check for enum membership?

kevinburkemeter:
Currently if I want to test for enum membership, I need to write a
custom function and then remember update it every time I add a new
field to an enum.

By placing all of the enums in a map, it is possible to write a
function in a user file in the same package that tests for membership
and does not require updating (since it can just check the map). This
also avoids the need to change the public API.

Fixes #447.

### DIFF
--- a/internal/dinosql/gen.go
+++ b/internal/dinosql/gen.go
@@ -1145,6 +1145,14 @@ const (
 	{{- end}}
 )
 
+{{ if $.EmitEnumMapping }}
+var Every{{.Name}} = map[{{.Name}}]bool{
+	{{ range .Constants -}}
+	{{.Name}}: true,
+	{{ end -}}
+}
+{{ end -}}
+
 func (e *{{.Name}}) Scan(src interface{}) error {
 	switch s := src.(type) {
 	case []byte:
@@ -1299,6 +1307,7 @@ type tmplCtx struct {
 	EmitJSONTags        bool
 	EmitPreparedQueries bool
 	EmitInterface       bool
+	EmitEnumMapping     bool
 }
 
 func (t *tmplCtx) OutputQuery(sourceName string) bool {
@@ -1330,6 +1339,7 @@ func Generate(r Generateable, settings config.CombinedSettings) (map[string]stri
 		EmitInterface:       golang.EmitInterface,
 		EmitJSONTags:        golang.EmitJSONTags,
 		EmitPreparedQueries: golang.EmitPreparedQueries,
+		EmitEnumMapping:     golang.EmitEnumMapping,
 		Q:                   "`",
 		Package:             golang.Package,
 		GoQueries:           r.GoQueries(settings),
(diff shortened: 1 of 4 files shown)


internal/mysql/gen.go
```diff
@@ -63,7 +63,7 @@ func (pGen PackageGenerator) enumNameFromColDef(col *sqlparser.ColumnDefinition)
 		dinosql.StructName(col.Name.String(), pGen.CombinedSettings))
 }
 
-// Structs marshels each query into a go struct for generation
+// Structs marshals each query into a go struct for generation
```
kyleconroy, line 66: My spelling strikes again!

internal/dinosql/gen.go
```diff
@@ -1145,6 +1145,12 @@ const (
 	{{- end}}
 )
 
+var every{{.Name}} = map[{{.Name}}]bool{
```
kyleconroy, line 1148: I don't want to add this map by default. Can you put it behind a feature flag? `emit_enum_mapping` or something like that. Also, I think this should be public, as by default I don't encourage people to add code to the generated package.

kevinburkemeter: Done

internal/dinosql/gen.go
```diff
@@ -1145,6 +1145,14 @@ const (
 	{{- end}}
 )
 
+{{ if $.EmitEnumMapping }}
+var Every{{.Name}} = map[{{.Name}}]bool{
```
kyleconroy, line 1149: Looking over this, don't you want these to be the string values of the enums? Not the Enum values themselves?

kevinburkemeter: Either way... can just as easily do e.g `if db.EveryCompanyRole[CompanyRole("unknownstring")]`

kyleconroy: I'll leave it up to you 👍

kyleconroy:
@kevinburkemeter still interested in this feature landing? A release is going out on Thursday

kyleconroy:
@kevinburkemeter I’m closing this for now, please feel free to reopen if you have time to update to the master branch.

PR #455, opened 2020-04-12
Support different size integers for MySQL

Xeoncross:
Different MySQL column types like bigint or smallint are not given the correct Go equivalent. This is a non-perfect start to correctly matching the sizes.


kyleconroy:
Thanks for the PR. To get the tests passing, you'll need to run `make regen`.

kyleconroy:
@Xeoncross MySQL support is now provided by a different package, which handles integers sizes correctly.

Issue #462, opened 2020-04-26
Fully qualified name to Go type not applied, no erros shown in console


Maxiride:
Due to the heavy involvement of marshalling and unmarshalling to JSON needed in my application I wanted to use the _zero._ types provided by [gopkg.in/guregu/null.v3/zero](https://gopkg.in/guregu/null.v3/zero).

I've tried to do a test run to substitute strings with the _zero.String_ type provided by the package without success, here is the relevant configuration.

```
overrides:
  - go_type: "gopkg.in/guregu/null.v3/zero.String"
    db_type: "string"
    "null": true
```
I am using  `"null"` thanks to the report made in #413

The type applied to the generated models is still the default `sql.NullString`.

Given that I never heard of "fully qualified name to a go type" I suppose I might be wrong in the configuration 😅 and I'd appreciate to shed some light on the subject.

---
Upon further browsing the issue I found #255 which I suppose is related if not the same issue I am encountering.

Issue #255, opened 2020-01-10
JSON Parser fails for type override on complex package names


rupertevill:
The standard syntax for defining a type override in the lib at the moment is:

` "overrides": [
      {
          "go_type": "github.com/gofrs/uuid.UUID",
          "postgres_type": "uuid"
      }
  ]`

This results in the following import statement and type (in the struct) in the auto gen code:

```
import (
    "github.com/gofrs/uuid"
)

type test struct {
    columnOne uuid.UUID
}
```
However, the parser fails if the package name containing the desired type override is more complex, for example "gopkg.in/guregu/null.v3".

Have looked at the parser as it is at the moment and the bug fix doesn't look to bad, let me if you want any contribution 👍

Maxiride:
I've closed #462 since I'm pretty sure it was a duplicate issue.
Is the bug caused by the presence of a dot in the package name?

rupertevill:
> I've closed #462 since I'm pretty sure it was a duplicate issue.
> 
> Is the bug caused by the presence of a dot in the package name?

Correct. The presence of the dot in the package name causes the issue.

Maxiride:
Too bad... those packages are my holy grail.

Anyway,
>Have looked at the parser as it is at the moment and the bug fix doesn't look to bad, let me if you want any contribution 👍

@rupertevill did you eventually started working on a PR?

kyleconroy:
This has been fixed in main and will be released in the next version. Here's what it looks like:

```json
{
  "version": "1",
  "packages": [
    {
      "path": "go",
      "name": "override",
      "engine": "postgresql",
      "schema": "schema.sql",
      "queries": "query.sql",
      "overrides": [
        {
          "column": "foo.bio",
          "nullable": true,
          "go_type": {
            "import": "gopkg.in/guregu/null.v4",
            "type": "String"
          },
        }
      ]
    }
  ]
}
```

```
import (
    null_v4 "gopkg.in/guregu/null.v4"
)

type test struct {
    columnOne null_v4.String
}
```

Issue #413, opened 2020-03-22
Field `null` in Override configuration is ignored.


ipsusila:
Hi, thank you for the nice software. I'm facing the following issue, in which the `overrides` settings was not applied properly.

When the configuration is:
```
overrides:
  - go_type: "sql.NullInt64"
    db_type: "integer"
    null: true
```
the `null` config value is not decoded properly. However, if I changed the configuration to (use `"null"` as field name)
```
overrides:
  - go_type: "sql.NullInt64"
    db_type: "integer"
    "null": true
```
`sqlc` will correctly override type as expected. In the former case, `YAML` decoder is failed to decode the configuration (the value is always `false`), perhaps because the field name is `null`.
Probably this issue is related to #412 too?

kyleconroy:
Oof, bitten again by the strangeness of the YAML spec. We'll probably rename the field to be something like `not_null` or `is_null`.

kyleconroy:
Closing in favor of #570

Issue #177, opened 2019-12-16
Can't override with type in same package


mightyguava:
I wanted to create a type for an enum, e.g. `State`

```sql
# job.sql
create table job (
  id bigserial primary key,
  state int
)
```

```go
# job.sql.go
type Job struct {
  ID string `json:"id"`
  State State `json:"state"`
}
```

```go
# enums.go
type State string
const (
  StateRunning State = iota
  StateFailed
  StateDone
)
```

It's convenient to have `enums.go` in the same package as the generated code, but there doesn't seem to be a way to tell the generator to not create an import. I end up with the package importing itself which Go reports as an import cycle.

alibitek:
Indeed, for some reason the enum type is set as the struct field type in some cases. 
Don't know exactly why. 
I tried to set an override, but unfortunately I can't, because the enum from db is generate in the same package as the code using it.

**First test**:   
```
        {
          "go_type": "gitlab.com/company/project/database.ECustomEnum",
          "db_type": "e_custom_enum"
        }
```
this gives error: `import cycle not allowed` , because the package where the code is generated, namely: `database` is trying to import itself via the override.  

**Second test**:  
```
        {
          "go_type": "gitlab.com/company/project/ECustomEnum",
          "db_type": "e_custom_enum"
        }
```
this gives error: `cannot load gitlab: cannot find module providing package gitlab` 
Inside `querier.go` I get a weird import like:  
```
import (
   "context"
   "github.com/google/uuid"
   "gitlab"
)
```

kyleconroy:
This should now on main. You'll need to specify the `go_type` using an object instead of just a string

```json
        {
          "go_type": {"type": "ECustomEnum"},
          "db_type": "e_custom_enum"
        }

```

PR #463, opened 2020-04-27
Fix complex packages import paths

maxiride:
**EDIT**
_Please skip directly to [this](https://github.com/kyleconroy/sqlc/pull/463#issuecomment-620020148) reply as none of what I made matters anymore since I found another import path case which fails with this current PR._
---

See #462 and #255

Attempting to fix the parsing of complex package names by leveraging regex capabilities to capture the relevant parts.

After smashing my head for an hour I asked for help on Reddit, details of the regex expression can be found [here](https://www.reddit.com/r/regex/comments/g8zhjv/every_time_that_i_came_here_i_say_to_myself_that/foqd3qh/), credits for the regex go to /u/Lee_Dailey.

I've also added the relevant test entries which used to fail and now they pass.

---
### Notes
I've left the original code as commented out for reference, it should be removed if the PR is to be accepted to avoid polluting the code with unnecessary lines.

The regex expression can be easily batch tested [here](https://regex-golang.appspot.com/assets/html/index.html) to ensure it satisfies package import names, I am not aware of any other strange import path other than `gopkg.in/guregu/null.v3/zero`. To nitpick the issue we should find other packages with complex import paths. I can't guarantee that this will work with **every package ever**.

**Two assumptions on fully qualified names have been made to build the regex:**

1. A type name is always in the form of pkgName.TypeName
1. The import path ends with the rightmost forward slash

This led to formulate these two regex:

- To match a package type: match everything backwards from the end of the string to the first encountered forward slash /
- To match a package name: match everything forward from the start of the string till the last dot found


internal/config/config.go
```diff
@@ -218,7 +219,20 @@ func (o *Override) Parse() error {
 		if lastSlash == -1 {
 			return fmt.Errorf("Package override `go_type` specifier %q is not the proper format, expected 'package.type', e.g. 'github.com/segmentio/ksuid.KSUID'", o.GoType)
 		}
-		typename = o.GoType[lastSlash+1:]
+
+		// Upstream typename assignment
+		// typename = o.GoType[lastSlash+1:]
```
Maxiride, line 224: I left the upstream code as commented out while fiddling around. Those lines should be removed upon an accepted PR to avoid polluting the code.

internal/config/config.go
```diff
@@ -229,7 +243,18 @@ func (o *Override) Parse() error {
 		if strings.HasSuffix(typename, "-go") {
 			typename = typename[:len(typename)-len("-go")]
 		}
-		o.GoPackage = o.GoType[:lastDot]
+		// Upstream package name assignment
+		//o.GoPackage = o.GoType[:lastDot]
```
Maxiride, line 247: I left the upstream code as commented out while fiddling around. Those lines should be removed upon an accepted PR to avoid polluting the code.

Maxiride:
I now noticed that the same author (guregu) has another package with the import path `gopkg.in/guregu/null.v3`.

This import path, with the implementation in this PR will fail, given the go_type `gopkg.in/guregu/null.v3.String` this will yield:

- Captured type: `null.v3.String ` ❌, working type is `null.String`
- Captured package name: `gopkg.in/guregu/null.v3 ` ✅



---
image

.gitignore
```diff
@@ -0,0 +1,4 @@
+﻿.idea
```
kevinburkemeter, line 1: A .gitignore should only ignore project artifacts. Files that are specific to your editor can be ignored by editing `.git/info/exclude` on your laptop. You can also configure Git to ignore these files automatically when you create a new repo by adding a template.

.gitignore
```diff
@@ -0,0 +1,4 @@
...
+go.sum
+go.sum
+*.sum
```
kevinburkemeter, line 4: This makes the rules on lines 2 and 3 extraneous. In any event, I'm not sure it's what we want.

.gitignore
Maxiride: Thanks for the input 🙌

.gitignore
Maxiride: I don't understand, maybe should I remove the .gitignore altogether from the versioning?
I now notice that this repo doesn't have one.

kyleconroy:
> If @kyleconroy can chime in I might attempt to implement this approach. However, being it a bit of an overhaul on several layers of the package I'd prefer to settle down with an accepted line of work before commencing anything.

I'm going to need to think about the way forward a bit, so I suggest you pause work on the PR.

The issue is that we're trying to parse three parts of information from a single string: the import path, the package name, and the type identifier. I think you're correct that we need to use a struct for this instead of a string. 

I also want to solve https://github.com/kyleconroy/sqlc/issues/177 at the same time.

My current proposal looks like this:

```yaml
overrides:
  - go_type:
      import: "gopkg.in/guregu/null.v3"
      package: "null"
      type: "String"
    db_type: "string"
    "null": true
```

Maxiride:
> I also want to solve #177 at the same time.
> 
> My current proposal looks like this:
> 
> ```yaml
> overrides:
>   - go_type:
>       import: "gopkg.in/guregu/null.v3"
>       package: "null"
>       type: "String"
>     db_type: "string"
>     "null": true
> ```

Having explicit declaration of the three entries surely pinpoints what the user wants and helps in generating proper code. However, as of now, I'm not sure #177 could be solved by such approach, if I understood correctly the issue some kind of additional entry would be needed to mark the override to not be imported. Nonetheless seems to me that the structure you proposed is a good base starting point.

#413 on the other hand seems to me more easily addressable, if the config gets changed it could be an occasion to rethink the naming of the entry `"null"`. As you already proposed either `not_null` or `is_null` is perfectly fine IMHO.

Since the null entry [defaults to false](https://github.com/kyleconroy/sqlc#type-overrides) I personally find that `is_null` is a more clearer take on the phraseology to use.

Here the same struct with the null field addressed.

```yaml
overrides:
  - go_type:
      import: "gopkg.in/guregu/null.v3"
      package: "null"
      type: "String"
    db_type: "string"
    is_null: true
```
---
Overall, from my understanding of the package, addressing #255 and #413 at the same time is more easily done in this PR. #177 requires more brainstorming beyond the config struct itself.

kyleconroy:
I think this solves #177 quite well. You just leave out the import and package declarations

```yaml
overrides:
  - go_type:
      type: "State"
    column: "job.state"
```

Maxiride:
I've attempted to implement a solution, unit tests pass however, I am unable to proceed further in the testing of the new implementation due to an upstream issue #476.

I've made a custom Dockerfile to build the package in a Linux environment and while the package does build without errors I am unable to run any further [integration test](https://github.com/kyleconroy/sqlc#running-tests) as debugging into a running docker container is a PITA.

I'll update when I have either a workaround or a working solution.

internal/config/config.go
```diff
@@ -114,7 +114,7 @@ type SQLGen struct {
 type SQLGo struct {
 	EmitInterface       bool              `json:"emit_interface" yaml:"emit_interface"`
 	EmitJSONTags        bool              `json:"emit_json_tags" yaml:"emit_json_tags"`
-	EmitPreparedQueries bool              `json:"emit_prepared_queries" yaml:"emit_prepared_queries":`
```
Maxiride, line 117: This was a typo in the master branch

examples-complex-import/authors/sqlc.yaml
```diff
@@ -0,0 +1,17 @@
+version: "1"
```
kyleconroy, line 1: The entire `examples-complex-import` directory should be moved to the `internal/endtoend/testdata` directory and renamed to `complex_go_type`.

internal/config/config.go
```diff
@@ -114,7 +114,7 @@ type SQLGen struct {
...
+	EmitPreparedQueries bool              `json:"emit_prepared_queries" yaml:"emit_prepared_queries"`
```
kyleconroy, line 117: Thanks for catching this

internal/config/config.go
```diff
@@ -127,27 +127,40 @@ type SQLKotlin struct {
 }
 
 type Override struct {
-	// name of the golang type to use, e.g. `github.com/segmentio/ksuid.KSUID`
-	GoType string `json:"go_type" yaml:"go_type"`
+	// Import path, package name and type name as you would type them in the IDE
+	GoTypeParam GoTypeParams `json:"go_type" yaml:"go_type"`
```
kyleconroy, line 131: This is a backwards incompatible change. The `GoTypeParams` struct needs to work if `go_type` is a string. This will require a custom `UnmarshalJSON` method.

internal/config/config.go
```diff
@@ -207,35 +220,49 @@ func (o *Override) Parse() error {
...
-		if lastDot == -1 {
-			return fmt.Errorf("Package override `go_type` specifier %q is not the proper format, expected 'package.type', e.g. 'github.com/segmentio/ksuid.KSUID'", o.GoType)
-		}
-		if lastSlash == -1 {
-			return fmt.Errorf("Package override `go_type` specifier %q is not the proper format, expected 'package.type', e.g. 'github.com/segmentio/ksuid.KSUID'", o.GoType)
-		}
-		typename = o.GoType[lastSlash+1:]
-		if strings.HasPrefix(typename, "go-") {
+		// TODO need to implement checks for #177, is an import path is set, the package is external otherwise it doesn't need to be imported
+		/*		This checks shouldn't be needed anymore, proper Go type definition responsibility is to the end user given the explicitly set fields.
```
kyleconroy, line 228: This comment isn't correct since we need to continue to support a world where `go_type` is a string.

Maxiride:
> As I mentioned in a few comments, we can't change the type of the `go_type` parameter in a backwards-incompatible way. 

I didn't think about it but I stand corrected, it is indeed necessary.

>I'm happy to take over the PR from here if you don't want to implement backwards compatibility. Let me know.

I will happily hand over the PR to be taken care of! =) Would have loved to contribute more but at this point I wouldn't even know how to proceed to implement backwards compatibility and I believe it's best left to you.

Checked the "Allow edits from maintainers" so you should be good to go.

Maxiride:
I'm not understanding why after moving the example authors folder to the endtoend directory the build is failing.

go fmt applies tabs

kyleconroy:
> I'm also sorry for the tons of commits I made which undoubtedly pollute the commit history, I've been more prone to a "commit often,

No need to apologize. We squash commits for PRs, so all your commits just turn into one :)

Maxiride:
Any update about on which milestone we can expect this feature to be implemented?

kyleconroy:
I'm hoping for the next release, 1.5.0. I've been focused on some large architecture changes, so this fell to the wayside. Sorry for the delay.

Maxiride:
No troubles!
Just wanted an update :) glad to hear it might land in 1.5.0


Il lun 22 giu 2020, 17:28 Kyle Conroy <notifications@github.com> ha scritto:

tv42:
You could just load the import path with https://godoc.org/golang.org/x/tools/go/packages and see what the name of the package is (the source is going to be available at sqlc generate time).

Or you could always explicitly name the import in the generated code, `import foo "lots.of.dashes"`.

tv42:
The code around this is a bit weird/neglected. This looks like it tries to massage the package name to strip common prefixes and suffixes, but then the carefully crafted `typename` is *not* used to set `o.GoPackage`. It seems to have confused type name vs package name...

I think it'd be a pretty easy refactor to discard all this logic and just explicitly name all non-stdlib imports.

https://github.com/kyleconroy/sqlc/blob/ba4c9bfc0bfdcb1616ca5ef9b1299f35b313c254/internal/config/config.go#L231-L244

kyleconroy:
Sorry that this PR has languished. I'm heads down working on MySQL support. I promise this is next, but it's still going to be a while.

Maxiride:
> Sorry that this PR has languished. I'm heads down working on MySQL support. I promise this is next, but it's still going to be a while.

It surely needs polishing or a complete rewrite using @tv42 suggested method.

For what's worth I've been using my fork with these modifications since April without issues so far.

JessBellon:
Any update on this? 
If its alright with y'all I can take a swing at a new PR written using @tv42's idea?

maxiride:
Haven't heard anything from kyleconroy but I honestly think the new proposed approach is waaaay smoother than my amateurish approach.

I'd be glad to see a polished and merged into master PR to address the issue. This PR "just works" but I do recognise is far from clean code. Please do go ahead.

maxiride:
I'm closing the PR in favour of #757 that use the tools `x/tools/packages` offers which, in turn, brings to sqlc a more elegant, concise and clean solution to the issue. I hope to see it merged soon in sqlc, I've already switched to use the binary built from #757 branch.

PR #474, opened 2020-05-02
Add LeavePlural config option

tgulacsi:
If set to true, the struct names are not singularized.

My table names are not English, stripping the "s" from the end
is quite annoying.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -114,7 +114,8 @@ type SQLGen struct {
 type SQLGo struct {
 	EmitInterface       bool              `json:"emit_interface" yaml:"emit_interface"`
 	EmitJSONTags        bool              `json:"emit_json_tags" yaml:"emit_json_tags"`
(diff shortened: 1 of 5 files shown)


internal/config/config.go
```diff
@@ -114,16 +114,18 @@ type SQLGen struct {
 type SQLGo struct {
 	EmitInterface       bool              `json:"emit_interface" yaml:"emit_interface"`
 	EmitJSONTags        bool              `json:"emit_json_tags" yaml:"emit_json_tags"`
-	EmitPreparedQueries bool              `json:"emit_prepared_queries" yaml:"emit_prepared_queries":`
+	EmitPreparedQueries bool              `json:"emit_prepared_queries" yaml:"emit_prepared_queries"`
```
kyleconroy, line 117: Thanks for catching this (it was also caught in #463)

PR #483, opened 2020-05-12
parser: Use same function to load SQL files

### LINKED ISSUE #478: #426 causes non-.sql schema files to be examined
Before #426, if the schema path contained a file not ending in .sql, that file would be skipped. Now it results in a syntax error (if you're lucky 😄).

kyleconroy:
Fixes #478


carlmjohnson:
Thanks!

PR #489, opened 2020-05-14
parser: Support columns from subselect statements

### LINKED ISSUE #467: Columns not found when joining subquery
If you define a new column in a subquery and do a join the new column can't be referenced by the containing query. A "does not exist" error is returned.

An example:

```sql
CREATE TABLE users (
    user_id INT PRIMARY KEY
);
CREATE TABLE names (
    user_id INT REFERENCES users,
    first_name TEXT,
    last_name TEXT
);

SELECT
    user_id,
    full_name
FROM users
JOIN (
    SELECT
        user_id,
        first_name || last_name AS full_name
    FROM names
) AS full_names USING (user_id);
```

Returns `column "full_name" does not exist`, though the query is valid.

maxhawkins:
Fixes #467


peteole:
Cool that it works, however there is still an issue when using joins in a subselect statement.
Take this example:

```sql
SELECT r.*
FROM (
        project_saves
        JOIN projects
        ON project_saves.project_id = projects.id
    ) r;
```
This fails with the message 
```bash
# package sqlc
sql-queries/projects.sql:159:1: empty edit contents
```
sqlc seems to believe that thesubselect `r` has no columns since when selecting a field of `r` the error message is "column "id" does not exist". Is there an easy way to fix this?

gwillem:
Also, subquery aliasing does not work when `sqlc.arg` is used

```sql
SELECT *
FROM (SELECT * FROM report LIMIT ...) AS r
JOIN ... 
WHERE r.created_at >= sqlc.arg('since')
```
```
table alias "r" does not exist
```
(I need to do pre-filtering, if I do a flat join, MySQL runs out of sort memory)

PR #492, opened 2020-05-14
internal/dinosql: Remove unneeded column suffix

kyleconroy:
When returned columns have no name set, a name is generated based on
index (e.g. column_1, column_2). When generating Go structs, sqlc wasn't
using the correct column name to track duplicate, resulting in nonsense
field names like `Column5_2`.

This has the added benefit for generating correct JSON tag values for
these columns.


kyleconroy:
No surprise here, but credit to @kevinburkemeter for the initial bug report.

PR #511, opened 2020-05-19
Support for dbmate migrations

PeachIceTea:
Adds support for the migration syntax of [dbmate](https://github.com/amacneil/dbmate).


kyleconroy:
Thanks for the perfect pull request :D

PR #523, opened 2020-05-24
compiler: Apply rename rules to enum constants

### LINKED ISSUE #522: case for structs
I have the following in SQL:

```sql
CREATE TYPE ip_protocol AS enum('tcp', 'ip', 'icmp');
```

Which generates the unglamorous:

```go
type IpProtocol string

const (
	IpProtocolTcp  IpProtocol = "tcp"
	IpProtocolIp   IpProtocol = "ip"
	IpProtocolIcmp IpProtocol = "icmp"
)
```

I'm wondering if there's a way to control the case of the generated constants.

kyleconroy:
I'm not 100% sure I want to make this change, as it could break existing generated code.

Fixes #522


kyleconroy:
@kevinburkemeter Can you try out this commit and tell me if it breaks any of your existing code?

kevinburkemeter:
guessing you mean "not 100%"

kyleconroy:
Yep, edited my original comment.

kevinburkemeter:
No diff using this commit, though I haven't tried adding the enum rename, let me try that now.

kevinburkemeter:
Works perfectly though I had to read the patch contents to figure out the underscore trick.

kyleconroy:
@kevinburkemeter what value for the rename struct would you have expected instead of `ip_protocol_tcp`?

kevinburkemeter:
I don't know, ip_protocol.tcp maybe?

Issue #526, opened 2020-05-27
go: json.Marshal of empty rows result encodes to `null` rather than `[]`


powersjcb:
`json.Marshal` encodes `null` for all empty results rather than `[]`.

# Example
When this function returns an empty list, the result is marshaled to `null`.

https://github.com/powersjcb/monitor/blob/be4992358e6f3589d5095fca88f6f179b4483842/go/src/server/db/metric.sql.go#L37

I was able to resolve this by changing the way the empty results are instantiated.

`var items = []ResultRow` -> `items := make([]ResultRow, 0)`

https://github.com/golang/go/issues/26866
https://apoorvam.github.io/blog/2017/golang-json-marshal-slice-as-empty-array-not-null/

powersjcb:
Should have a PR open for this in a few min.

kyleconroy:
@powersjcb Thanks for the bug report and the pull request.

I've decided to keep the existing behavior. Marshalling a nil slice is documented behavior in the `json/encoding` package. This change could break existing applications that rely on the current behavior. I also like the fact that memory isn't allocated if the query returns zero results.

I hope you understand and I'm sorry you spent time putting together a PR only to have it not merged.

kyleconroy:
Fixed by #551 and #528

PR #528, opened 2020-05-28
Enhancement: add `emit_empty_json_arrays` config so `[]` can be encoded for empty results

powersjcb:
It looks like the default behavior of `json.Marshal` leaks go language implementation details. Empty slices which are initialized with the syntax, `var items []Data` are backed by `nil`. The golang stdlib marshaler encodes `nil` slices as `null`.

Empty slices which are initialized with the syntax `items := make([]Data, 0)` are backed by a pointer to an empty slice. The marshaller encodes empty slices as `[]`.

There's unfortunately no way to add constructor checks or json tag annotations with the current go runtime. (It appears the only other way to resolve this issue would be to use a 3rd party marshaller.)

This change adds a configuration that allows for control of the way slices are constructed and causes `json.Marshal` to encode `[]` for empty results.

Resolves Issue #526


kyleconroy:
Thanks for fixing the bug your found, really appreciate it.

I'm still mulling over if I want to make this change. I understand the bug, but want to make sure it doesn't break existing code for others. It's always scary when sqlc generate creates a diff.

powersjcb:
Yeah, this change will break a client implementations in both of these cases:
- client explicitly handles the `null` case with something other than skipping iteration over a for loop
- client assumes that `items.length > 0` if `items !== null`

I feel like the second assumption is unlikely, but the first case is plausible.


Have any thoughts about making this a config change or a sem version bump?

examples/booktest/postgresql/jsonemptyarrays/db_test.go
```diff
@@ -0,0 +1,40 @@
+// +build examples
```
powersjcb, line 1: Any opinions about this being nested under `./postgresql/json` vs being it's own top level test dir?

powersjcb:
Updated the PR to make this an optional config.

kyleconroy:
@powersjcb ahh, I’m so sorry this didn’t get merged. I got this PR confused with a very similar one (https://github.com/kyleconroy/sqlc/pull/551) which I ended up merging.

If you’d like, you can update this PR to use that configuration option, but for now I think the best action is to close this PR.

Sorry again for the mistake, I don’t want you to feel like your contributions aren’t welcome.

PR #529, opened 2020-05-29
Support SELECT * in MySQL 

kyleconroy:
This uses the compiler package and generic AST


internal/compiler/go_type.go
```diff
@@ -33,6 +33,23 @@ func (r *Result) goInnerType(col *Column, settings config.CombinedSettings) stri
...
 	}
 
+	// TODO: Extend the engine interface to handle types
+	switch settings.Package.Engine {
+	case config.EngineMySQL, config.EngineXDolphin:
+		return r.mysqlType(col, settings)
+	case config.EnginePostgreSQL:
+		return r.postgresType(col, settings)
+	case config.EngineXLemon:
+		return r.postgresType(col, settings)
```
kevinburkemeter, line 43: Why wouldn't you have this in the same case statement as the line above?

kyleconroy: Should have left a comment; calling postgresType for SQLite is a temporary solution. SQLite offers a very small number of built in types and should not use the same mapping.

PR #530, opened 2020-05-30
compiler: temp fix for typecast function parameters

### LINKED ISSUE #520: Type cast in positional function args yields invalid params
```sql
CREATE FUNCTION plus(a integer, b integer) RETURNS integer AS $$
    BEGIN
        RETURN a + b;
    END;
$$ LANGUAGE plpgsql;

-- name: MakeIntervalPositional :one
SELECT plus($1, $2::INTEGER);
```

yields the params struct
```go
type MakeIntervalPositionalParams struct {
	A       int32
	B       int32
	Column3 int32
}
```

With an extra field `Column3`.

The extra field disappears if the INTEGER cast is removed.

maxhawkins:
In cases where a FuncCall has a TypeCast node as an
argument, resolveCatalogRefs would erroneously output
two parameters for that argument instead of one.

This fix eliminates the duplicate param and allows queries
with typecast function args to compile, but it loses track
of the argument's connection with the function called. This
means the parameters use the generic ColumnN names even when
the function has named parameters.

A future refactor of resolveCatalogRefs to better address
function calls will be necessary to address the name issue.

(this only edits the experimental package and  does not fix
the old dinosql compiler)

Closes #520


maxhawkins:
Does this look OK?

kyleconroy:
Yes! I hit the "Merge master into branch" button and then forgot to merge. Doing so now.

PR #531, opened 2020-06-01
MySQL: Prevent Panic when WHERE clause contains parenthesis. 

JessBellon:
The sqlc MySQL parser didn't have the logic to handle a where clause with parenthesis in it and would panic. This seems to fix it, but there might be some edge cases I didn't account for. (Sorry for the semi-contrived test case, its similar to what I needed but its also kinda silly)


kyleconroy:
Thanks!

PR #547, opened 2020-06-17
cmd/sqlc: Bump version to v1.4.0

kyleconroy:


internal/cmd/cmd.go
```diff
@@ -46,7 +46,7 @@ var versionCmd = &cobra.Command{
 		if version == "" {
 			// When no version is set, return the next bug fix version
 			// after the most recent tag
-			fmt.Printf("%s\n", "v1.3.1-devel")
+			fmt.Printf("%s\n", "v1.4.0")
```
kevinburkemeter, line 49: You could just use Println here... or `Print("v1.4.0\n")`

PR #551, opened 2020-06-19
Add option to return empty slices in generated :many query, to improve usability with JSON

philippta:
## Description

sqlc is very useful when creating REST APIs. However when querying for a collection of models using the `:many` query, while the corresponding database table is empty, the result in JSON is marshalled into `null`. To improve the usability with APIs returning JSON and preferabily `[]` instead of `null`, I propose to add an option to change the returned slice of generated `:many` queries. See examples below.

## Current behavior

Let's take a look at the example in `sqlc/examples/authors/`:

The following query in `query.sql`:
```sql
-- name: ListAuthors :many
SELECT * FROM authors
ORDER BY name;
```

Generates the following Go code:
```go
func (q *Queries) ListAuthors(ctx context.Context) ([]Author, error) {
	// snip ...
	var items []Author
	// snip ...
	return items, nil
}
```

When the database table `authors` is empty, the returned results of this function is `[]pkg.Authors(nil), <nil>`.
Therefore when marshalling the result as JSON, the output is therefore `null` aswell. 

Example code:
```go
// Create response struct
resp := struct {
	Authors []Author `json:"authors"`
}{}

// Fill in authors
resp.Authors, _ = db.ListAuthors(context.Background())

// Marshall to JSON
data, _ := json.MarshalIndent(resp, "", "\t")

// Print it
fmt.Println(string(data))
```
Output:
```json
{
	"authors": null
}
```

In usecases like JavaScript applications or other API consumers it is easier to deal with an empty array `[]` instead of `null` because `null` has to be treated specifically.

---

## Proposed behavior
With this pull request I've added an additional configuration option (opt in) which changes this behavior, so the marshalled JSON turnes into an empty array `[]` instead of `null`. The added configuration option is called `emit_empty_slices` (but could be changed if someone suggests a more precise keyword).

In the sqlc.json this option goes right where the other options go:
```javascript
{
  "version": "2",
  "sql": [
    {
      "schema": "schema.sql",
      "queries": "query.sql",
      "engine": "postgresql",
      "gen": {
        "go": {
          "package": "authors",
          "out": ".",
          "emit_json_tags": true,
          "emit_empty_slices": true // <----- new addition
        }
      }
    }
  ]
}
```

With this option enabled, the generated Go code for the same function will have just a very suble difference in the initialization of the returned slice. `var items []Author` becomes `items := []Author{}`:

```go
func (q *Queries) ListAuthors(ctx context.Context) ([]Author, error) {
	// snip ...
	items := []Author{}
	// snip ...
	return items, nil
}
```

Executing the Go code in the example above on an empty authors table will now result in this JSON output:
```json
{
	"authors": []
}
```

## Conclusion
This makes direct use of sqlc results in a JSON encoded REST API much more comfortable for e.g. JavaScript clients.

### DIFF
--- a/internal/codegen/golang/gen.go
+++ b/internal/codegen/golang/gen.go
@@ -652,7 +652,11 @@ func (q *Queries) {{.MethodName}}(ctx context.Context, {{.Arg.Pair}}) ([]{{.Ret.
 		return nil, err
 	}
 	defer rows.Close()
+	{{- if $.EmitEmptySlices}}
+	items := []{{.Ret.Type}}{}
+	{{else}}
 	var items []{{.Ret.Type}}
+	{{end -}}
 	for rows.Next() {
 		var {{.Ret.Name}} {{.Ret.Type}}
 		if err := rows.Scan({{.Ret.Scan}}); err != nil {
@@ -729,6 +733,7 @@ type tmplCtx struct {
 	EmitJSONTags        bool
 	EmitPreparedQueries bool
 	EmitInterface       bool
+	EmitEmptySlices     bool
 }
 
 func (t *tmplCtx) OutputQuery(sourceName string) bool {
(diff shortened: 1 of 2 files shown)


philippta:
Oh, I have just noticed there was another pull request proposing this change. Was wondering what you meant with „your other pull request“. 
But I am very happy was merged so quickly! Thank you!

powersjcb:
Glad we got one of these merged!

PR #566, opened 2020-06-23
dolphin: Add support for column aliasing

kyleconroy:


internal/engine/dolphin/convert.go
```diff
@@ -165,6 +180,7 @@ func convert(node pcast.Node) ast.Node {
 		return convertSelectStmt(n)
 
 	default:
+		fmt.Printf("%T\n", n)
```
kyleconroy, line 183: I'm keeping this for now, as it's helpful during tests

Issue #583, opened 2020-07-07
Case statements don't work with mysql


JessBellon:
Neither the new or old mysql parser can handle `case` statements. 

The new parser fails with the error message: 
`2020/07/07 15:40:49 unknown MySQL type: any`

kyleconroy:
This has been [fixed](https://play.sqlc.dev/p/014994fa8898ca1e139ed00a55bd8e3900846018a73f82f4494c9fc537f7339a) (somewhat). The query now compiles but the return type is `interface{}`.

PR #584, opened 2020-07-07
endtoend: Add (failing) mysql test for bool case stmt

JessBellon:
#583


kyleconroy:
Closed in favor of #642

PR #597, opened 2020-07-16
Fixed :execresult generation

DanielAtwood:
`:execresult` is generating with an extra `:=` in the return. This just removes those two characters.


kyleconroy:
Thanks for catching this! I obviously need to add a test case here

PR #601, opened 2020-07-20
Add support for 'cidr' and 'interval' Postgres types

coadler:


kevinburkemeter:
Thanks for this. Could you add a test case for each of the new types you're adding? You can look at the commit that adds inet support for an example.

Kevin

coadler:
Ah, cool. Will do!

coadler:
@kevinburkemeter I seem to have found https://github.com/kyleconroy/sqlc/pull/141, which from a brief scan doesn't add any tests. I assume the place to go is `internal/endtoend` and to add an `ipaddr` folder with the test cases? Thanks

PR #611, opened 2020-07-22
Add benchmarks for end-to-end tests and examples

kyleconroy:


internal/endtoend/endtoend_test.go
```diff
@@ -44,6 +44,30 @@ func TestExamples(t *testing.T) {
...
+		b.Fatal(err)
+	}
+	for _, replay := range files {
+		if !replay.IsDir() {
+			continue
+		}
+		tc := replay.Name()
+		b.Run(tc, func(b *testing.B) {
+			path := filepath.Join(examples, tc)
+			var stderr bytes.Buffer
```
kevinburkemeter, line 63: You probably want to reset stderr between runs otherwise it will grow without bound.

internal/endtoend/endtoend_test.go
```diff
@@ -139,3 +163,30 @@ func expectedStderr(t *testing.T, dir string) string {
...
+		return nil
+	})
+	if err != nil {
+		b.Fatal(err)
+	}
+	for _, replay := range dirs {
+		tc := replay
+		b.Run(tc, func(b *testing.B) {
+			path, _ := filepath.Abs(tc)
+			var stderr bytes.Buffer
```
kevinburkemeter, line 186: Ditto here.

PR #612, opened 2020-07-22
reset stderr between benchmark runs

kyleconroy:
otherwise it will grow without bound


kyleconroy:
cc @kevinburkemeter

kevinburkemeter:
:thumbsup:

PR #620, opened 2020-07-24
dolphin: Implement exapansion with reserved words

kyleconroy:
compiler: rename the Engine struct to Compiler
compiler: move methods onto the Compiler struct
compiler: Use backticks for reserved words in MySQL
codegen: Handle queries with backticks


internal/endtoend/testdata/star_expansion_reserved/mysql/go/query.sql.go
```diff
@@ -0,0 +1,36 @@
...
+
+package querytest
+
+import (
+	"context"
+	"database/sql"
+)
+
+const starExpansionReserved = `-- name: StarExpansionReserved :many
+SELECT ` + "`" + `group` + "`" + ` FROM foo
```
kyleconroy, line 12: It's gross, but it works 100% of the time and doesn't require a call to `strings.Replace`

PR #621, opened 2020-07-24
dolphin: Implement parameter refs

kyleconroy:
For some reason, the dolphin engine isn't generating two parameters for the query


internal/endtoend/testdata/params_two/mysql/go/query.sql.go
```diff
@@ -0,0 +1,37 @@
...
+	"context"
+	"database/sql"
+)
+
+const fooByAandB = `-- name: FooByAandB :many
+SELECT a, b FROM foo 
+WHERE a = ? and b = ?
+`
+
+func (q *Queries) FooByAandB(ctx context.Context, a sql.NullString) ([]Foo, error) {
```
kyleconroy, line 16: This should be `FooByAandB(ctx context.Context, a, b sql.NullString)`

Issue #594, opened 2020-07-16
Master fails to build due to github.com/pingcap/tidb


coadler:
See https://github.com/pingcap/tidb/issues/18584

```
$ go build ./...
go: github.com/pingcap/tidb@v1.1.0-beta.0.20200526100040-689a6b6439ae requires
	github.com/pingcap/br@v0.0.0-20200521085655-53201addd4ad requires
	github.com/pingcap/tidb@v1.1.0-beta.0.20200509133407-a9dc72cf2558 requires
	github.com/pingcap/pd/v4@v4.0.0-rc.1.0.20200422143320-428acd53eba2: invalid pseudo-version: revision 428acd53eba2 is not a descendent of preceding tag (v4.0.0-rc.1)
```

Reverting to `github.com/pingcap/tidb v1.1.0-beta.0.20200426094148-df4c9f05c49c` which was present in the `v1.4` release fixes it.

kyleconroy:
I can't reproduce this bug. Running `go build ./...` on my local machine works. CI on the master branch is also working. Any idea why my build works and your build doesn't? Maybe because I have some packages cached?

coadler:
Here's a rundown of what happens on my machine

```
$ go version
go version go1.14.3 linux/amd64

$ git --no-pager log -1
commit ba09fa0f23672853925e906a432af3ae93f0b71c (HEAD -> master, origin/master, origin/HEAD)
Author: Daniel Atwood <daniel@atwood.io>
Date:   Tue Jul 21 13:06:08 2020 -0400

    Fixed :execresult generation (#597)

$ go build ./...
go: github.com/pingcap/tidb@v1.1.0-beta.0.20200526100040-689a6b6439ae requires
	github.com/pingcap/br@v0.0.0-20200521085655-53201addd4ad requires
	github.com/pingcap/tidb@v1.1.0-beta.0.20200509133407-a9dc72cf2558 requires
	github.com/pingcap/pd/v4@v4.0.0-rc.1.0.20200422143320-428acd53eba2: invalid pseudo-version: revision 428acd53eba2 is not a descendent of preceding tag (v4.0.0-rc.1)

$ go get github.com/pingcap/tidb@master
go: github.com/pingcap/tidb@v1.1.0-beta.0.20200526100040-689a6b6439ae requires
	github.com/pingcap/br@v0.0.0-20200521085655-53201addd4ad requires
[...]

PR #629, opened 2020-07-28
dolphin: Use test driver instead of TiDB driver

kyleconroy:
Thus removing the dependency on pingcap/tidb. If we need a more advanced
driver, we can always write one ourselves


kyleconroy:
We also can avoid issues like #594 in the future

PR #631, opened 2020-07-28
Added ISNULL return type

DanielAtwood:
[Documentation](https://dev.mysql.com/doc/refman/8.0/en/comparison-operators.html#function_isnull)
> If expr is NULL, ISNULL() returns 1, otherwise it returns 0. 

I've added `isnull` to return `int` as that's what `mysql` does.

MYSQL does understand `true` and `false` so maybe it should return `bool` instead of `int`. I'm not very familiar with how `functionReturnType` is used so I don't know if that's an option.


DanielAtwood:
Would it be beneficial for me to go through all the functions and add their return types?

kyleconroy:
Thanks! 

> Would it be beneficial for me to go through all the functions and add their return types?

At this time, no. The mysql package is deprecated. All current and future MySQL work is happening in the engine/dolphin. For functions, we're going to automatically generate the type signatures. See the work done for PostgreSQL in https://github.com/kyleconroy/sqlc/pull/550

kyleconroy:
I wrote up a bit more about the deprecation of the mysql package https://github.com/kyleconroy/sqlc/issues/633

PR #635, opened 2020-07-29
compiler: Support calling functions with defaults

kyleconroy:
Fix a number of bugs when resolving function calls.
Switch to using functions generated from a default PostgreSQL instance.
Add a new test case ripped from the PostgreSQL docs.

I've noted the improvements in individual comments


internal/endtoend/testdata/func_args/go/query.sql.go
```diff
@@ -11,9 +11,9 @@ const makeIntervalDays = `-- name: MakeIntervalDays :one
 SELECT make_interval(days => $1::int)
 `
 
-func (q *Queries) MakeIntervalDays(ctx context.Context, dollar_1 int32) (interface{}, error) {
+func (q *Queries) MakeIntervalDays(ctx context.Context, dollar_1 int32) (int64, error) {
```
kyleconroy, line 14: `make_interval` now returns an actual type instead of `interface{}`. This happened because we can now resolve function calls with defaults and named parameters.

internal/endtoend/testdata/func_args/go/query.sql.go
```diff
@@ -33,9 +33,9 @@ const makeIntervalSecs = `-- name: MakeIntervalSecs :one
 SELECT make_interval(secs => $1)
 `
 
-func (q *Queries) MakeIntervalSecs(ctx context.Context, secs interface{}) (interface{}, error) {
+func (q *Queries) MakeIntervalSecs(ctx context.Context, secs float64) (int64, error) {
```
kyleconroy, line 36: The compiler is now smart enough to pull out type information for named parameters in function calls.

internal/endtoend/testdata/func_args/go/query.sql.go
```diff
@@ -45,12 +45,12 @@ SELECT plus(b => $2, a => $1)
...
 
 type PlusParams struct {
-	B int32
 	A int32
+	B int32
 }
 
 func (q *Queries) Plus(ctx context.Context, arg PlusParams) (int32, error) {
-	row := q.db.QueryRowContext(ctx, plus, arg.B, arg.A)
+	row := q.db.QueryRowContext(ctx, plus, arg.A, arg.B)
```
kyleconroy, line 53: !!!

The compiler generated incorrect code here. The order of the arguments wasn't correct.

internal/endtoend/testdata/generate_series/go/query.sql.go
```diff
@@ -17,15 +17,15 @@ type GenerateSeriesParams struct {
 	Column2 time.Time `json:"column_2"`
 }
 
-func (q *Queries) GenerateSeries(ctx context.Context, arg GenerateSeriesParams) ([]interface{}, error) {
+func (q *Queries) GenerateSeries(ctx context.Context, arg GenerateSeriesParams) ([]string, error) {
```
kyleconroy, line 20: Goodbye `interface{}`! Real return values for function calls

internal/endtoend/testdata/pg_advisory_xact_lock/go/query.sql.go
```diff
@@ -11,8 +11,8 @@ const advisoryLock = `-- name: AdvisoryLock :many
 SELECT pg_advisory_unlock($1)
 `
 
-func (q *Queries) AdvisoryLock(ctx context.Context, key int64) ([]bool, error) {
-	rows, err := q.db.QueryContext(ctx, advisoryLock, key)
+func (q *Queries) AdvisoryLock(ctx context.Context, pgAdvisoryUnlock int64) ([]bool, error) {
```
kyleconroy, line 14: While the PostgreSQL documentation claims the first parameter to `pg_advisory_unlock` is named `key`, that name isn't actually included in the definition of the function stored in the database.

internal/engine/postgresql/catalog.go
```diff
@@ -5,7 +5,7 @@ import "github.com/kyleconroy/sqlc/internal/sql/catalog"
 func NewCatalog() *catalog.Catalog {
 	c := catalog.New("public")
 	c.Schemas = append(c.Schemas, pgTemp())
-	c.Schemas = append(c.Schemas, pgCatalog())
+	c.Schemas = append(c.Schemas, genPGCatalog())
```
kyleconroy, line 8: This is a bit scary! I'm curious if we'll have to walk this specific change back.

PR #639, opened 2020-07-30
compiler: Return a single column from coalesce

### LINKED ISSUE #458: Multiple model fields generated for COALESCE
`sqlc version:  v1.1.1-devel`

DB: Postgres

Given the following statement in a select list:

`COALESCE(ua.firstnames, ti.firstnames) as tenantfirstnames`

This results it 2 fields in the generated Row model:

```
&i.Tenantfirstnames,
&i.Tenantfirstnames_2,
```

which is erroneous and causes execution of the query to fail

kyleconroy:
Fixes #458

### DIFF
--- a/internal/compiler/output_columns.go
+++ b/internal/compiler/output_columns.go
@@ -97,17 +97,25 @@ func outputColumns(qc *QueryCatalog, node ast.Node) ([]*Column, error) {
 			}
 
 		case *pg.CoalesceExpr:
+			var found bool
 			for _, arg := range n.Args.Items {
+				if found {
+					continue
+				}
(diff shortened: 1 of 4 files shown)


internal/compiler/output_columns.go
```diff
@@ -97,18 +97,26 @@ func outputColumns(qc *QueryCatalog, node ast.Node) ([]*Column, error) {
 			}
 
 		case *pg.CoalesceExpr:
```
kyleconroy, line 99: This behavior is improved, but still a bit strange. We don't really know what the return type will be until typechecking lands, which is far off.

PR #651, opened 2020-08-06
Dolphin: Implement Rename

alecbz:
This change is [Reviewable](https://reviewable.io/reviews/kyleconroy/sqlc/651)


kyleconroy:
Fantastic. I'm planning on moving [the PostgreSQL catalog tests](https://github.com/kyleconroy/sqlc/blob/master/internal/engine/postgresql/catalog_test.go) to the end to end tests, which would give you a better place to add these tests in the future.

Issue #364, opened 2020-02-27
Support dynamic queries


mehdijoafshani:
Does it make sense to support dynamic queries ? *(Dynamic `where` and `orderBy` clauses)*.

Due to some requirements we have within our team, we need select statements with dynamic filters (a set of unknown numbers of where conditions, or lets say a set of optional where conditions which could be added to the query based on generated function's arguments) and also dynamic columns to order.

I know `sqlc` only supports static queries, however we still prefer to use `sqlc` with some extra work to provide dynamic stuff.
I am wondering if you have any plan to support dynamic queries or you are open to have this feature in your repo (I'd like to contribute and can create the PR in next few days I guess). BTW I think if we provide the feature, lots of more people can leverage `sqlc`.

mehdijoafshani-tb:
I would use named parameters and `CASE WHEN` instead to make the queries dynamic.

dharmjit:
Hi @mehdijoe, I requested for the same in slack but it's not supported as of now. I was thinking even if this got supported there is another mess of switch case statements for right sqlc method to call based on the dynamics parameters passed.

kyleconroy:
@mehdijoe Can you provide a concrete example of your needs? I'm surprised that the dynamic filters are unknown. Having a concrete example would also make it easier to discuss the best solution. @dharmjit, it sounds like you have the same issue? Could you provide your use case?

mehdijoe:
Thanks for ur response @kyleconroy
> I'm surprised that the dynamic filters are unknown. 

Actually unknown **numbers** of conditions ..., OK, here is my use case:

```sql
create table t1
(
    field1    varchar(100),
    field2    varchar(100),
);
```
I need to select on `t1`, which the filters are determined from UI, so I need any combination of filters, it means I should support following 9 queries on my back-end:
```sql
select * from t1 
select * from t1 where field1 = $1
select * from t1 where field1 like $1
select * from t1 where field2 = $1
select * from t1 where field2 like $1
select * from t1 where field1 = $1 and field2 = $2
select * from t1 where field1 like $1 and field2 = $2
select * from t1 where field1 = $1 and field2 like $2
select * from t1 where field1 like $1 and field2 like $2
```
Users can also sort on any column(s) they wish (ASC/DESC), it means for each query listed above I need to append following `Order By` clauses:
```sql
order by field1 ASC
order by field1 DESC
order by field2 ASC
order by field2 DESC
order by field1, field2 ASC
[...]

PR #656, opened 2020-08-06
Add an emit_db_tags config option to support interoperability with gorp

alecbz:
This option is very similar to emit_json_tags, but it uses "db" as the tag name.

This allows us to use sqlc-generated models with [gorp](https://github.com/go-gorp/gorp), which is especially useful as a workaround to https://github.com/kyleconroy/sqlc/issues/364.

---
This change is [Reviewable](https://reviewable.io/reviews/kyleconroy/sqlc/656)


internal/codegen/golang/result.go
```diff
@@ -73,10 +73,17 @@ func buildStructs(r *compiler.Result, settings config.CombinedSettings) []Struct
 				Comment: table.Comment,
 			}
 			for _, column := range table.Columns {
+				tags := map[string]string{}
```
alecbz, line 76: @kyleconroy This feels a little clunky -- happy to try and do something fancier/less repetitive.

Though given its only in three places (one of which is deprecated I think?), maybe not too bad in current form?

internal/config/config.go
```diff
@@ -113,6 +113,7 @@ type SQLGen struct {
 type SQLGo struct {
 	EmitInterface       bool              `json:"emit_interface" yaml:"emit_interface"`
 	EmitJSONTags        bool              `json:"emit_json_tags" yaml:"emit_json_tags"`
+	EmitDBTags          bool              `json:"emit_db_tags" yaml:"emit_db_tags"`
```
kyleconroy, line 116: [sqlx](https://github.com/jmoiron/sqlx#usage) also uses the `db` tag for annotating structs. Can we change this value to `emit_gorp_tags`? I want to make sure we have the flexibility to support other tools. This means we'd need to change the name of the test folder as well.

internal/endtoend/testdata/emit_db_tags/go/models.go
```diff
@@ -0,0 +1,14 @@
+// Code generated by sqlc. DO NOT EDIT.
+
+package querytest
+
+import (
+	"database/sql"
+)
+
+type User struct {
+	ID        int32          `db:"id"`
```
kyleconroy, line 10: According to the gorp docs, we could add a bit more information here

```go
type User struct {
    ID         int64     `db:"id, primarykey, autoincrement"`
    ...
}
```

internal/config/config.go
alecbz: sqlx uses the `db` tag in the same way though, no? (I guess this depends on the resolution to that that other comment thread).

internal/endtoend/testdata/emit_db_tags/go/models.go
alecbz: 1. Can you point me to this in the gorp docs?

2. Does this provide any value? AFAIK other attributes like that would only be used for creating tables based on gorp structs, but if we're using sqlc, the tables must have already been created.

internal/codegen/golang/result.go
kyleconroy: I'm all for clunky and repetitive. The generation code is already a bit of a mess and I don't think this makes it any worse.

internal/config/config.go
kyleconroy: I don't think sqlx has support for the autoincrement and primary key annotations? We'd need to confirm with both library authors that the annotations are compatible with each other

internal/endtoend/testdata/emit_db_tags/go/models.go
kyleconroy: There doesn't appear to be any section in the docs. You can find a few different use cases in the [quickstart](https://github.com/go-gorp/gorp#quickstart).

```go
type Post struct {
    // db tag lets you specify the column name if it differs from the struct field
    Id      int64  `db:"post_id"`
    Created int64
    Title   string `db:",size:50"`               // Column size set to 50
    Body    string `db:"article_body,size:1024"` // Set both column name and size
}


// Example of using tags to alias fields to column names
// The 'db' value is the column name
//
// A hyphen will cause gorp to skip this field, similar to the
// Go json package.
//
// This is equivalent to using the ColMap methods:
//
//   table := dbmap.AddTableWithName(Product{}, "product")
//   table.ColMap("Id").Rename("product_id")
//   table.ColMap("Price").Rename("unit_price")
//   table.ColMap("IgnoreMe").SetTransient(true)
//
// You can optionally declare the field to be a primary key and/or autoincrement
//
type Product struct {
    Id         int64     `db:"product_id, primarykey, autoincrement"`
    Price      int64     `db:"unit_price"`
    IgnoreMe   string    `db:"-"`
}
```

Also, there are apparently two versions of grop? v1 and v2?

> Does this provide any value?
I think it does. If a field is marked as primary key, `Insert` can return the ID of the created record.

alecbz: > I think it does. If a field is marked as primary key, Insert can return the ID of the created record.

Ah yeah, that's true. Ok, I can add that then and rename the field to be gorp-specific.

kyleconroy: The other option, before you get started on the rename, is to have the `emit_db_tags` only emit the column names. Then have a separate `emit_gorp_tags` that does all the magic. My preference is for tool-specific configuration, because I doubt there are too many people using a combination of grop / sqlx / gorm.

alecbz: My preference would probably be to just leave it as the column names and not worry about the other gorp annotation stuff, since that's consistent across gorp/sqlx/gorm, and also gets us most of the value (all of the value for my specific use-case -- we use `.SetKeys()` to explicitly specify the primary keys instead of the `primarykey` tag).

But if we're going to include the extra gorp features now, yeah, I think having a single "gorp" option makes sense.

kyleconroy: > My preference would probably be to just leave it as the column names and not worry about the other gorp annotation stuff

Great, let's go ahead and keep it as is, naming included. I'll open an issue tracking a future use case for orm-specific tags

PR #663, opened 2020-08-14
Support MySQL `CREATE TABLE x LIKE y` statements

alecbz:
This change is [Reviewable](https://reviewable.io/reviews/kyleconroy/sqlc/663)


internal/sql/catalog/table.go
```diff
@@ -135,14 +135,29 @@ func (c *Catalog) createTable(stmt *ast.CreateTableStmt) error {
...
-	for _, col := range stmt.Cols {
-		tbl.Columns = append(tbl.Columns, &Column{
-			Name:      col.Colname,
-			Type:      *col.TypeName,
-			IsNotNull: col.IsNotNull,
-			IsArray:   col.IsArray,
-		})
+
+	if stmt.ReferTable != nil && len(stmt.Cols) != 0 {
+		return errors.New("create table node cannot have both a ReferTable and Cols")
```
alecbz, line 142: @kyleconroy Not sure what to do here. Could raise this while parsing instead of here? Should we create a dedicated error type for this?

kyleconroy: It's better to catch errors like this here instead of parsing. It's fine to just return an one-off error, especially considering error would come from a mistake inside sqlc as opposed to an error by a user.

alecbz:
> Excellent! PostgreSQL also supports `CREATE TABLE x LIKE y` statements. Can you please move the test you created into a `mysql` directory inside the `create_table_like` directory? I'll go ahead and implement it for PostgreSQL.

Done.

internal/sql/catalog/table.go
```diff
@@ -135,14 +135,31 @@ func (c *Catalog) createTable(stmt *ast.CreateTableStmt) error {
...
+	}
+
+	if stmt.ReferTable != nil {
+		_, original, err := c.getTable(stmt.ReferTable)
+		if err != nil {
+			return err
+		}
+		for _, col := range original.Columns {
```
alecbz, line 150: FYI: Modified this to makes copies of the columns themselves too, in case any renames etc. change the columns in the original table.

PR #676, opened 2020-08-25
dolphin: Add ENUM() support

### LINKED ISSUE #502: MySQL  Enum redeclarations
I am evaluating sqlc for one of our future projects. Currently, I am using sqlc to write a small migration tool for converting a legacy MySQL database to PostgreSQL.

At the very first step I ran into a problem with the generated code for this schema fragment:

```
CREATE TABLE `users` (
  `id` int(14) unsigned NOT NULL AUTO_INCREMENT,
  `status` enum('active','inactive','blocked') NOT NULL DEFAULT 'active',
  `testuser` enum('N','Y') NOT NULL DEFAULT 'N',
  `readonly` enum('N',Y') NOT NULL DEFAULT 'N',
...
```

The problem is that those enums cause redeclaration errors as it creates constants with the raw Enum name.

```
type TestuserType string

const (
	N TestuserType = "N"
	Y TestuserType = "Y"
)

type ReadonlyType string

const (
	N ReadonlyType = "N"
	Y ReadonlyType = "Y"
)
```

The real `users` schema includes 12 enums of the type `enum('N','Y')` and a lot of the other tables share the same enum names too.

I wonder that it seems that I can not use enums with the same names in different columns or even different tables.

Is this because MySQL support is being experimental or am I doing something wrong here and there is a way to solve this problem?

kyleconroy:
I'm still deciding on how to pick the type name for MySQL enums. See the discussion in #502


kyleconroy:
I'm going to merge this now, but will revisit the naming scheme before v1.6.0 is release.

internal/sql/catalog/table.go
```diff
@@ -153,13 +153,24 @@ func (c *Catalog) createTable(stmt *ast.CreateTableStmt) error {
...
 				Name:      col.Colname,
 				Type:      *col.TypeName,
 				IsNotNull: col.IsNotNull,
 				IsArray:   col.IsArray,
 				Comment:   col.Comment,
-			})
+			}
+			if col.Vals != nil {
+				typeName := ast.TypeName{
+					Name: col.Colname,
```
alecbz, line 165: @kyleconroy I'm not 100% sure when this if triggers (for columns that use an implicit enum for the type, instead of a named one?) but this breaks with "type already exists" if it triggers twice for the same column name.

Thoughts on how to address? Prepending the table's name might be an easy-ish fix?

alecbz: https://github.com/kyleconroy/sqlc/pull/716

Issue #477, opened 2020-05-06
mysql: unable to query with reserved words


gpontesss:
MySQL column names can be named with reserved words when escaped with backticks. An example table with a MySQL reserved word `KEY`:

```sql
-- file: schema.sql
CREATE TABLE `some_table` (
   -- ...
  `key` varchar NOT NULL
);
```

`sqlc` can parse the table and create models for create statements such as the above. But if we want to query `some_table` with an expression that contains the `key` column it seems to ignore de backticks. An exemple of what I mean:

```sql
-- file: queries.sql
/* name: GetAllSomeTable :many */
SELECT
  *
FROM some_table
WHERE
  `some_table`.`key` = ?;
```

With these files, we can have a `sqlc.yaml` file as bellow:

```yaml
# file: sqlc.yaml
version: "2"

sql:
  - engine: "mysql"
    schema: "./schema.sql"
    queries: "./queries.sql"
    gen:
      go:
        package: "db"
        out: "./db"
```

`sqlc generate`, for that example, throws the following output:

```
// Code generated by sqlc. DO NOT EDIT.
// source: queries.sql

package db

import (

        "context"




)




const getAllSomeTable = `-- name: GetAllSomeTable :many
select id, `key` from some_table where some_table.`key` = ?
`








func (q *Queries) GetAllSomeTable(ctx context.Context, key string) ([]SomeTable, error) {
        rows, err := q.db.QueryContext(ctx, getAllSomeTable, key)
        if err != nil {
                return nil, err
        }
        defer rows.Close()
        var items []SomeTable
        for rows.Next() {
                var i SomeTable
                if err := rows.Scan(&i.ID,&i.Key); err != nil {
                        return nil, err
                }
                items = append(items, i)
        }
        if err := rows.Close(); err != nil {
                return nil, err
        }
        if err := rows.Err(); err != nil {
                return nil, err
        }
        return items, nil
}









# package db
error generating code: source error: 19:13: expected ';', found key (and 3 more errors)
```

kyleconroy:
The `mysql:beta` engine correctly escapes MySQL keywords

```go
// Code generated by sqlc. DO NOT EDIT.
// source: query.sql

package querytest

import (
	"context"
	"database/sql"
)

const getAllSomeTable = `-- name: GetAllSomeTable :many
SELECT
  ` + "`" + `key` + "`" + `
FROM some_table
WHERE
[...]

PR #677, opened 2020-08-25
compiler: Use engine-specific reserved keywords

kyleconroy:
A few months I added this comment to the compiler:

> Add a method to the parser / engine for [reserved keyword] instead

Today is that day!


kyleconroy:
This should also fix #477

PR #680, opened 2020-08-25
examples: Port ondeck to MySQL

kyleconroy:


kyleconroy:
I need to figure out how to create different MySQL databases in sqltest

PR #716, opened 2020-09-21
Prepend table names to the generated enum names

alecbz:
Otherwise we'll get a collision if a column name is shared for an enum column across two tables

---
This change is [Reviewable](https://reviewable.io/reviews/kyleconroy/sqlc/716)


internal/sql/catalog/table.go
```diff
@@ -162,7 +163,7 @@ func (c *Catalog) createTable(stmt *ast.CreateTableStmt) error {
 			}
 			if col.Vals != nil {
 				typeName := ast.TypeName{
-					Name: col.Colname,
+					Name: fmt.Sprintf("%s_%s", stmt.Name.Name, col.Colname),
```
alecbz, line 166: One minor annoyance with this is that it doesn't "singularize" the table's name.

It might also be frustrating to always opt-in to the longer enum names, when sometimes the terser names might suffice. But that seems trickier to fix

kyleconroy:
I was going to say that this change is backwards incompatible, but this is the new MySQL implementation, so I'm okay making a breaking change. Not sure why you're test is failing, but I think this is overall a good change.

alecbz:
> I was going to say that this change is backwards incompatible, but this is the new MySQL implementation, so I'm okay making a breaking change.

Oh, hm. Well I'm all for avoiding backwards incompatible changes, but I think in this case it's more or less unavoidable? Right now the generated code will cause name collisions on the generated "enum" Go types; not sure  how else to resolve that.

That said, it might be worth spending more time deciding how the names should best be formed? E.g., if we want to get the de-pluralization to work, might be better to do it now than introduce another incompatible change in the future?

alecbz:
@kyleconroy Any thoughts on ^ ?

If not, can fix the tests/go forward with this change.

kyleconroy:
I think this is fine as-is! Will merge once the tests have been updated and are pasing.

alecbz:
Sounds good -- tests green now

kyleconroy:
Thank you!

meutraa:
Thanks for this one. Drove me crazy with databases where many tables had `type` being an enum.
I was just overriding the type to a string in the config file and it felt wrong.

PR #722, opened 2020-10-03
Fix unused imports for Go code generator

### LINKED ISSUE #721: Generated Go code has build errors due to unused imports
Generated Go files for queries and the Querier interface import the return type of `:exec` queries, which can result in unused imports--a build error in Go. For example, `database/sql` is imported for void or nullable return types; and `github.com/lib/pq` is imported for arrays.

Examples below are generated using sqlc v1.5.0, with `engine: postgresql` and `emit_interface: true`.

### Example 1
This a practical example that reproduces this issue. In this example, `database/sql` is imported because sqlc considers `sql.NullBool` as the Go type for `void`.

SQL schema/query:
```sql
create function do_something() returns void as
$$
begin
  return;
end;
$$
language plpgsql;

-- name: DoSomething :exec
select do_something();
```
Go query (unused `database/sql` import):
```go
// Code generated by sqlc. DO NOT EDIT.
// source: do_something.sql

package db

import (
	"context"
	"database/sql"
)

const doSomething = `-- name: DoSomething :exec
select do_something()
`

func (q *Queries) DoSomething(ctx context.Context) error {
	_, err := q.db.ExecContext(ctx, doSomething)
	return err
}
```
Go Querier interface (unused `database/sql` import):
```go
// Code generated by sqlc. DO NOT EDIT.

package db

import (
	"context"
	"database/sql"
)

type Querier interface {
	DoSomething(ctx context.Context) error
}

var _ Querier = (*Queries)(nil)
```

### Example 2
The example below also reproduces this issue, even though it's not very practical.

SQL schema/queries:
```sql
create table foo (
  bar  int,
  bars int[] not null
);

-- name: Bar :exec
select bar from foo;

-- name: Bars :exec
select bars from foo;
```
Go queries (unused `database/sql` and `github.com/lib/pq` imports):
```go
// Code generated by sqlc. DO NOT EDIT.
// source: queries.sql

package db

import (
	"context"
	"database/sql"

	"github.com/lib/pq"
)

const bar = `-- name: Bar :exec
select bar from foo
`

func (q *Queries) Bar(ctx context.Context) error {
	_, err := q.db.ExecContext(ctx, bar)
	return err
}

const bars = `-- name: Bars :exec
select bars from foo
`

func (q *Queries) Bars(ctx context.Context) error {
	_, err := q.db.ExecContext(ctx, bars)
	return err
}
```
Go Querier interface (unused `database/sql` import):
```go
// Code generated by sqlc. DO NOT EDIT.

package db

import (
	"context"
	"database/sql"
)

type Querier interface {
	Bar(ctx context.Context) error
	Bars(ctx context.Context) error
}

var _ Querier = (*Queries)(nil)
```

soroushj:
Fixes #721.

Essentially, I've changed the condition for importing return types (and imports required for their scanning) from `!q.Ret.isEmpty()` to `q.Cmd != ":exec" && !q.Ret.isEmpty()`.

I've also added a new end-to-end test that fails without this patch.

P.S. I didn't test the Kotlin code generator, but probably it has the same issue (see [here](https://github.com/kyleconroy/sqlc/blob/ae0a1f8f545e5114775682015fe2035a86a5c2b9/internal/codegen/kotlin/imports.go#L43) and [here](https://github.com/kyleconroy/sqlc/blob/ae0a1f8f545e5114775682015fe2035a86a5c2b9/internal/codegen/kotlin/imports.go#L139)). But since unused imports are not a compile error in Kotlin, it's not a big issue anyway.


kyleconroy:
👏 👏 👏 

Thank you for: 

a) an amazing bug report
b) a pull request to fix it
c) tests
d) all of the above

PR #727, opened 2020-10-07
Add option to use pointers instead of sql.Null* types

mvrhov:
This adds support to use pointers instead of sql.Null* types. I'd like to get general comments before having a go at adding a few tests.

### DIFF
--- a/internal/codegen/golang/mysql_type.go
+++ b/internal/codegen/golang/mysql_type.go
@@ -12,24 +12,28 @@ import (
 func mysqlType(r *compiler.Result, col *compiler.Column, settings config.CombinedSettings) string {
 	columnType := col.DataType
 	notNull := col.NotNull || col.IsArray
+	pointer := ""
+	if !notNull && settings.Go.UsePointers {
+		pointer = "*"
+	}
 
 	switch columnType {
 
 	case "varchar", "text", "char", "tinytext", "mediumtext", "longtext":
-		if notNull {
-			return "string"
+		if notNull || pointer != "" {
+			return pointer + "string"
 		}
 		return "sql.NullString"
 
 	case "int", "integer", "smallint", "mediumint", "year":
-		if notNull {
-			return "int32"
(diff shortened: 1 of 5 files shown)


alecbz:
There's already a way to configure the use of different types than the standard ones for fields, right? Could we expand that to support pointers vs. adding a dedicated config option just for pointers?

heikkilamarko:
Hi! What is the status of this issue?
Is there any way to generate pointers instead of sql.Null* types yet?

mvrhov:
Different version of this was implemented it's a bit annoying as you have to change the config and add the types you want to have pointers to...
This is my PostgreSQL list for now.

```yaml
overrides:
  - db_type: "pg_catalog.timestamp"
    go_type:
      import: "time"
      type: "Time"
      pointer: true
    nullable: true
  - db_type: "pg_catalog.timestamptz"
    go_type:
      import: "time"
      type: "Time"
      pointer: true
    nullable: true
  - db_type: "text"
    go_type:
      type: "string"
      pointer: true
    nullable: true
  - db_type: "ltree"
    go_type:
      type: "string"
      pointer: true
    nullable: true
  - db_type: "jsonb"
    go_type:
      import: "encoding/json"
      type: "RawMessage"
      pointer: true
    nullable: true
  - db_type: "pg_catalog.int2"
    go_type:
      type: "int16"
      pointer: true
    nullable: true
  - db_type: "pg_catalog.int4"
    go_type:
      type: "int"
      pointer: true
    nullable: true
  - db_type: "pg_catalog.int8"
    go_type:
      type: "int64"
      pointer: true
    nullable: true
  - db_type: "pg_catalog.bool"
    go_type:
      type: "bool"
      pointer: true
    nullable: true
  - db_type: "pg_catalog.float4"
    go_type:
      type: "float32"
      pointer: true
    nullable: true
  - db_type: "pg_catalog.float8"
    go_type:
      type: "float64"
      pointer: true
    nullable: true
```

PR #733, opened 2020-10-16
Detect nullable fields when having LEFT_JOIN, OUTER_JOIN or RIGHT_JOIN

ryad-eldajani:
Detects nullable fields when executing nullable joins LEFT_JOIN, OUTER_JOIN or RIGHT_JOIN.

I'm not sure though, if the implementation is consistent nor on the right layer in the application. I think, this is a lower layer issue that has to be detected and/or fixed in pg_query parser but I might be wrong.


gwenaskell:
Is there any progress to expect on this feature in a near future? Supporting LEFT JOINs is a very demanded feature on our side, I would be happy to help if desired

fr3fou:
Any updates on this? Or at least some kind of workaround would be apprciated

jwilner:
@gwenaskell, @fr3fou I've updated the good work here and have handled a few more cases in [this PR](https://github.com/kyleconroy/sqlc/pull/983)

kyleconroy:
Closing in favor of #983

PR #742, opened 2020-10-23
mysql update docs with env vars, update test with json field, fill cl…

kristijorgji:
- drop the test db after test finishes. I runned the tests several times and had many random named databases created, this fixes it
- update docs to specify the mysql env vars used in order to make it easier for contributing
- add one json field type in the example update models and test for that too


README.md
```diff
@@ -480,7 +480,9 @@ go test --tags=exp ./...
 
 To run the tests in the examples folder, a running PostgreSQL instance is
 required. The tests use the following environment variables to connect to the
-database:
+database
+
+####For PostgreSQL
```
kyleconroy, line 485: Can you add a space between `####` and the title? Same for the other one below

kyleconroy (changes requested):
The ondeck example is shared across MySQL and PostgreSQL, so I'd like to keep it the same. Please remove the added JSON column. A MySQL [JSON example](https://github.com/kyleconroy/sqlc/blob/master/internal/endtoend/testdata/json/mysql/query.sql) already exists in the form of an end-to-end test.

Really appreciate the README updates!

kristijorgji:
> The ondeck example is shared across MySQL and PostgreSQL, so I'd like to keep it the same. Please remove the added JSON column. A MySQL [JSON example](https://github.com/kyleconroy/sqlc/blob/master/internal/endtoend/testdata/json/mysql/query.sql) already exists in the form of an end-to-end test.
> 
> Really appreciate the README updates!

Done, reverted that change please re-review

kyleconroy (approved):
Thanks! I've rebased your changes onto master and will merge once the build passes.

PR #749, opened 2020-10-26
internal/cmd: print correct config file on parse failure

kevinburkemeter:


internal/cmd/generate.go
```diff
@@ -63,18 +63,19 @@ func Generate(e Env, dir string, stderr io.Writer) (map[string]string, error) {
 	}
 
 	if !yamlMissing && !jsonMissing {
-		fmt.Fprintln(stderr, "error parsing sqlc.json: both files present")
+		fmt.Fprintln(stderr, "error: both sqlc.json and sqlc.yaml files present")
```
kyleconroy, line 66: should be `fmt.Fprintf`

kyleconroy (changes requested):

internal/cmd/generate.go
kevinburkemeter: This one doesn't have anything interpolated... ? 

The one below this one should be Fprintf, I made a mistake there, but not this one

kevinburkemeter:
Done

kyleconroy (approved):

PR #756, opened 2020-11-01
Adding support to rename json fields in generated structs

al3jo:
I'm looking for a way to manipulate the generated JSON fields on structs.
These changes allow to do that without breaking existing configurations.

I'm wondering if you think this is a good idea or a good way to implement it.


kyleconroy:
I'm so sorry I never left a review for this PR. I know getting a message eight months later isn't any good, but I think it's better than getting no message ever.

As for the PR, applying renames to JSON would be too big of a breaking change for right now. I understand that sqlc is missing functionality around managing struct generation. It's something that I'm looking forward to fixing in the future.

PR #757, opened 2020-11-01
Improve overriding go_types with more complex import paths

JessBellon:
Follow up to #463, see the discussion there.
Also added support for overriding to a pointer (which I think solves the issue in #727 )


maxiride:
Much cleaner, x/tools really is the way to go!

mvrhov:
This means that there is a lot of work to be done to use pointers e.g. Don't forget to catch a type an add override.. copy overrides from project to project and don't miss any. The #727 solves this with one switch.

JessBellon:
@maxiride I was not aware of that error, thank you so much for letting me know! My first instinct is to have it auto `go get` the missing packages, but I worry that might be out of the scope of sqlc? I would love to get @kyleconroy's thoughts on that decision.

I'll make some changes over the weekend and see what I get working. Thanks again!

JessBellon:
@mvrhov I'm not sure I understand what you mean? Could you provide an example?

maxiride:
> 
> 
> @maxiride I was not aware of that error, thank you so much for letting me know! My first instinct is to have it auto `go get` the missing packages, but I worry that might be out of the scope of sqlc? I would love to get @kyleconroy's thoughts on that decision.
> 
> I'll make some changes over the weekend and see what I get working. Thanks again!

You're welcome! It's behind my understanding however, how config_test.go passed the tests... how can x/tools/packages fetch a package infos when the test is run in the IDE (I use Goland) while it can't after being built? If by any chance you find out, I'll be glad to learn about it.

mvrhov:
> 
> 
> @mvrhov I'm not sure I understand what you mean? Could you provide an example?

The test/example shows this code....
```json
        {
          "go_type": "*string",
          "db_type": "text",
          "nullable": true
        }
```

And overriding it like that means that I have to repeat this from config to config and catch any possible type that can be overridden like that. What I was purposing in #727 was a single switch in config file. Vs. 20 (guessing) native types * 3 lines

internal/config/config.go
```diff
@@ -229,26 +224,23 @@ func (o *Override) Parse() error {
...
-			typename = typename[len("go-"):]
-		}
-		if strings.HasSuffix(typename, "-go") {
-			typename = typename[:len(typename)-len("-go")]
+
+		importpath := o.GoType[:lastDot]
+		typename := o.GoType[lastDot+1:]
+
+		// Load given import path to ensure validity
+		p, err := packages.Load(&packages.Config{Mode: packages.NeedName}, importpath)
```
kyleconroy, line 232: I believe `packages.Load` required the Go tooling to be installed on the host, a dependency I'm not willing to take on right now. sqlc should work without any additional software. I'm investigating if there's parts of the code base that we can use to parse the import path without having to verify that it's valid.

kyleconroy (commented):
Left one comment about using the `packages` package. Right now I don't think it's a fit because it requires the Go toolchain.

kyleconroy:
Wrote up my plans here https://github.com/kyleconroy/sqlc/issues/785

PR #759, opened 2020-11-02
Use docker compose in CI

kyleconroy:


shaxbee:
@kyleconroy Could you consider using dockertest instead? It will allow tests to spin up docker containers themselves without requirement on docker-compose and should be faster as well.

PR #760, opened 2020-11-02
feat: use dockertest for examples test

shaxbee:
Use [dockertest](https://github.com/ory/dockertest) to spin up per-test postgres/mysql containers for examples tests.
Removes dependency on docker-compose and provides better test isolation, removing need to wait on database readiness.
Cleanup return value for `sqltest.MySQL` and `sqltest.PostgreSQL` was removed and is instead handled by `t.Cleanup` callback that is executed correctly when `os.Exit` is called.
Updated github workflow, removing now unnecessary services.
Removed docker-compose.yml.

Environment variables:
`DOCKERTEST_DISABLED` - do not spin up containers, will retain existing behaviour using `MYSQL*` and `PG*` environment variables.
`DOCKERTEST_RETAIN` - retain containers for failed tests, containers are named based on database driver and test name with random suffix to prevent collisions between tests with same short name and different runs.

MySQL tests are slower due to slow database startup.
MySQL Ping() will output annoying diagnostics that do not have impact on tests running correctly:
```
[mysql] 2020/11/03 15:16:19 packets.go:36: unexpected EOF
```

This could be possibly expanded to run mysql tests against dolphin engine easily and possibly postgres tests against cockroachdb.


shaxbee:
Ready for review.

kyleconroy:
@shaxbee Thanks for the changes. I appreciate all the work you've done on porting the tests. Despite the work, I'm inclined to keep the current setup since the only tests that need the databases are the examples test. I also like the current Compose setup because I need an instance of PostgreSQL or MySQL to test one-off queries.

I'd like to keep the changes you made to the README if you're up for backing out the other changes.

shaxbee:
@kyleconroy Sure, I'll close this PR and make separate one with reformatted README.md.

I use [prettier](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode) vscode extension, you can also use it standalone. Hope this PR might be useful at some time in future.

Issue #761, opened 2020-11-02
Support for json_build_object is broken in 1.5


ovadbar:
Because json_build_object's arguments are variable the argument check will not pass unless it is called with no arguments.

PR #771, opened 2020-11-08
Allow json_build_object to work 

ovadbar:
For issue https://github.com/kyleconroy/sqlc/issues/761. While this is kind of a hack, it gets the job done in the meantime.


kyleconroy:
It is indeed a hack! Thanks for the PR, but I'm going to go ahead and work on adding true support for variadic functions.

ovadbar:
Did you look at the latest changes? I am no longer using the hack and just ignoring the argument length test if they include a variadic argument. 
If you are looking to accomplish more than what I have laid out I am willing to work more on this.
I suppose if there is 1 variadic argument then no additional positional arguments should be able to follow, and I suppose that we can still check whether the minimum number of arguments is met, but I figure there is a superficial argument check currently.

PR #798, opened 2020-11-19
sql/catalog: Add support for variadic functions

### LINKED ISSUE #761: Support for json_build_object is broken in 1.5

kyleconroy:
Fixes #761


internal/sql/catalog/catalog.go
```diff
@@ -251,6 +251,7 @@ type Argument struct {
 	Name       string
 	Type       *ast.TypeName
 	HasDefault bool
+	Variadic   bool
```
ovadbar, line 254: Why did you add this struct key when there is a mode for Variadic that you can use to check?

internal/engine/postgresql/pg_catalog.go
```diff
@@ -11646,13 +11646,23 @@ func genPGCatalog() *catalog.Schema {
...
+			},
 			ReturnType: &ast.TypeName{Name: "json"},
 		},
 		{
-			Name:       "json_build_object",
-			Args:       []*catalog.Argument{},
+			Name: "json_build_object",
+			Args: []*catalog.Argument{
+				{
+					Type:     &ast.TypeName{Name: "any"},
```
ovadbar, line 11662: Shouldn't you mark Variadic arguments as having defaults because there can b 0 of them otherwise the following would fail.

```
postgres=# select json_build_object();
 json_build_object
-------------------
 {}
(1 row)
```

internal/endtoend/testdata/json_build/go/query.sql.go
```diff
@@ -0,0 +1,63 @@
...
+		&i.JsonBuildArray_2,
+		&i.JsonBuildArray_3,
+		&i.JsonBuildArray_4,
+	)
+	return i, err
+}
+
+const selectJSONBuildObject = `-- name: SelectJSONBuildObject :one
+SELECT
+  json_build_object('foo'),
```
ovadbar, line 40: I would add the following edge case to the tests this will likely fail.
```
json_build_object(),
```

Also I would remove the following tests because they actually fail in postgresql with the error
"ERROR:  argument list must have even number of elements"
```
  json_build_object('foo'),
  json_build_object('foo', 1, 2),
```
But catching those cases would be a little more difficult to catch without custom coding.

Issue #605, opened 2020-07-21
Postgres @named parameters don't always work


tv42:
Based on https://github.com/kyleconroy/sqlc/blob/master/docs/named_parameters.md
`@foo` and `sqlc.arg(foo)` are identical. However:

```
-- schema says create table foo ( id bigint );
-- name: demo1 :exec
DELETE FROM foo WHERE id=$1;
-- name: demo2 :exec
DELETE FROM foo WHERE id=@id;
-- name: demo3 :exec
DELETE FROM foo WHERE id=sqlc.arg(id);
```

results in

```
$ grep ^func demo.sql.go
func (q *Queries) demo1(ctx context.Context, id sql.NullInt64) error {
func (q *Queries) demo2(ctx context.Context) error {
func (q *Queries) demo3(ctx context.Context, id sql.NullInt64) error {
```

The `@foo` form doesn't get recognized in this scenario.

kyleconroy:
I'm not sure why, but the `@` form works when there's a space between the `=` and `@`. Take a look here: https://play.sqlc.dev/p/a2beb1f05746a8cac18a0d161645866c83f95896311b1e1b8ad5943199d7c5dd

Long term, I think I'm going to regret adding the `@` form of named parameters because of issues like this.

maxhawkins:
Sorry if this is digging up something that got settled in #71, but I wish the syntax was `:param` instead of `@param`. That way you could use psql to debug your queries against a running database:

```sh
psql --variable "param=1" -f queries.sql database
```

Unsure if using that syntax would help with the parsing problem here.

kyleconroy:
@maxhawkins The `:param` syntax doesn't work with the MySQL parser.

tv42:
If `@foo` is not even supposed to work right, https://docs.sqlc.dev/en/latest/howto/named_parameters.html needs to be updated to not recommend it as a shortcut.

> If the `sqlc.arg()` syntax is too verbose for your taste, you can use the `@` operator as a shortcut.

PR #816, opened 2020-11-27
Allow for mixing parameter styles

ovadbar:
This should allow for mixed styles of parameters.


internal/endtoend/testdata/mix/test.sql
```diff
@@ -0,0 +1,17 @@
...
+	id serial not null,
+	name text not null,
+	phone text not null
+);
+
+-- name: CountOne :one
+SELECT count(1) FROM bar WHERE id = $2 AND phone < @phone_param and name <> $1;
+
+-- name: CountTwo :one
+SELECT count(1) FROM bar WHERE id = sqlc.arg(id_param) AND phone < @phone_param and name <> $1;
```
kyleconroy, line 11: The @param style will be deprecated soon (see #605) so let's remove them from the test.

internal/endtoend/testdata/mix/sqlc.json
```diff
@@ -0,0 +1,11 @@
+{
+  "version": "1",
+  "packages": [
+    {
+      "path": "go",
+      "name": "querytest",
+      "schema": "test.sql",
```
kyleconroy, line 7: Please add the "engine" key and set it `postgresql`.

kyleconroy (changes requested):
Thanks for the patch! I agree that it doesn't make much sense to require a single parameter style. A few request:

* please pick a more descriptive name for the tests, such as `sqlc_arg_mixed` or `params_mixed`.
* Move the current tests into a `postgresql` subdirectory and add the same test for MySQL

ovadbar:
Thanks for the feedback.

(One thing I did notice is if you have a query as follows, the numbers always precede the sqlc.args convetion)

```
SELECT count(1) FROM table_name WHERE id > $1 AND id < sqlc.arg(id);
```

dant4301:
This would be great fix and would be awesome if merged.

There is an outstanding issue w.r.t using named_parameter for LIMIT for MySQL. We can use this fix to sort of work around it in the short term, but for the below:

I applied the patch and tried it for the query:
```
SELECT * FROM books WHERE book_id >= sqlc.arg(min_book_id) ORDER BY title ASC LIMIT ?;
```
The generated code looks like this:
```
type GetBooksByNameParams struct {
        Limit                 int32 `json:"limit"`
        MinBookID int64 `json:"min_book_id"`
}

func (q *Queries) GetBooksByName(ctx context.Context, arg GetBooksByNameParams) ([]Category, error) {
        rows, err := q.query(ctx, q.getBooksByNameStmt, getCategoriesByName, arg.Limit, arg.MinBookID)
...
```
Aren't the orders of parameters arg.Limit and arg.MinBookID switched?

ovadbar:
Yeah I kind of realized that was an issue but, with postgresql you can use dollar numbers format so it is not an issue and I did not think of how it would work with mysql.

I can look into fixing it but I really do not have the time.

kyleconroy:
> I can look into fixing it but I really do not have the time.

Understandable. I also don't have the time to look into fixing this. @dant4301 If you'd like this feature, you're going to have to get your hands dirty and take a shot at fixing this PR. I'm going to close this for now, but feel free to fork this branch and open up another PR if you can get it to work.

ovadbar:
@dant4301 can you look at https://github.com/kyleconroy/sqlc/pull/1071 and see if the new pull request solves your problems

dant4301:
@ovadbar,  I do not use sqlc now since I did a limited amount of codegen just enough for my needs. Thanks again though.

PR #818, opened 2020-11-28
catalog: Improve alter column type

aitva:
This commit adds support for altering a column type into an array and a new test to ensure that the generated code is valid.

This was working prior to v1.4.0 and may have been introduced by the refactor of the compiler.


kyleconroy (approved):

kyleconroy:
Thank you for the perfect PR: short, easy to review, and tested!

PR #824, opened 2020-12-05
Add a flag to allow the config file path to be specified

dmayle:
This diff allows us to pass the location of the config file on the command line, allowing us to call sqlc from another directory than the one where the config file is located.

This is useful in build systems where there are restrictions on the execution environment.


kyleconroy:
Replaced by https://github.com/kyleconroy/sqlc/pull/863

Issue #751, opened 2020-10-28
doc: in README  , supported types in gen.go links is 404


uta-mori:
### Type Overrides

The default mapping of PostgreSQL types to Go types only uses packages outside
the standard library when it must.

For example, the `uuid` PostgreSQL type is mapped to `github.com/google/uuid`.
If a different Go package for UUIDs is required, specify the package in the
`overrides` array. In this case, I'm going to use the `github.com/gofrs/uuid`
instead.

```yaml
version: "1"
packages: [...]
overrides:
  - go_type: "github.com/gofrs/uuid.UUID"
    db_type: "uuid"
```

Each override document has the following keys:
- `db_type`:
  - The PostgreSQL type to override. Find the full list of supported types in [gen.go](https://github.com/kyleconroy/sqlc/blob/master/internal/dinosql/gen.go#L438).
- `go_type`:
  - A fully qualified name to a Go type to use in the generated code.
- `nullable`:
  - If true, use this type when a column is nullable. Defaults to `false`.

maxiride:
I don't know how the project evolved and why the types aren't there anymore.
Anyway you can find them up to the tag 1.4.0 

See here: https://github.com/kyleconroy/sqlc/blob/v1.4.0/internal/dinosql/gen.go

uta-mori:
@maxiride 
thanks
I found it
https://github.com/kyleconroy/sqlc/blob/master/internal/codegen/golang/postgresql_type.go

kyleconroy:
Fixed

PR #827, opened 2020-12-07
📝  fix readme supported types file link

uta-mori:
`internal/dinosql/gen.go` was already missing and the link led to a 404 page, so I fixed it. to `internal/codegen/golang/postgresql_type.go`

#751


kyleconroy:
Thanks!

PR #842, opened 2020-12-27
mysql tinyint(1) as bool

hayama-junpei:
The current sqlc generate mysql `bool` column type as go `int32`.
pingcap/parser seems to interpret mysql `bool` column type as a `tinyint` type.
Probably because `bool` is an alias for  `tinyint(1)` in mysql.
This PR makes the mysql `bool` column to be output to the go `bool` and the mysql `tinyint(1)` to be output to the go `bool` as well.


kyleconroy:
This was implemented in #873. I took a slightly more verbose approach

PR #845, opened 2020-12-30
Added "emit_group_by_file" option

hayama-junpei:
If "emit_group_by_file" is true, output the struct separately for each input query file, instead of being output collectively in one struct `Queries`.
The name of the struct is determined from the name of the input query file. For example, if the input file is `authors.sql`, the` type Authors struct` will be output.

Depending on the value of "emit_group_by_file", there are the following syntax differences when using it.

- "emit_group_by_file": false
  ```go
  q := query.New(db)
  result, err := q.CreateAuthor(ctx, "Unknown")
  result, err = q.CreateBook(ctx, "my book title") 
  ```

- "emit_group_by_file": true
  ```go
  authors := query.NewAuthors(db)
  result, err := authors.Create(ctx, "Unknown")
  books := query.NewBooks(db)
  result, err = books.Create(ctx, "my book title") 
  ```

When "emit_group_by_file" is true, the following two options are also affected.

- "emit_prepared_queries"

  - If true, outputs prepared statement support with a lower level syntax.

- "emit_interface": 

  - if true, output a interface for each input query file, instead of being output a `Querier` interface.


kyleconroy:
Thanks for putting together this PR, but I'm unlikely to merge this change. The reason is that it's possible to get a very similar output by using two separate packages. I know it's inconvenient, but I'd rather not add more options that complicate the code generation.

PR #846, opened 2021-01-01
Support table returns for postgres go

warent:
I'm not sure if this is the correct/best way of doing this, but it seems to get the job done! It enables table types for postgres go, which is particularly useful in functions which can return a table type so we get the benefit of our existing structs rather than a string.

### DIFF
--- a/internal/codegen/golang/postgresql_type.go
+++ b/internal/codegen/golang/postgresql_type.go
@@ -6,6 +6,7 @@ import (
 	"github.com/kyleconroy/sqlc/internal/compiler"
 	"github.com/kyleconroy/sqlc/internal/config"
 	"github.com/kyleconroy/sqlc/internal/debug"
+	"github.com/kyleconroy/sqlc/internal/inflection"
 	"github.com/kyleconroy/sqlc/internal/sql/catalog"
 )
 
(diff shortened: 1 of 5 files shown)


warent:
I've added another part to this so that you can do `select * function()` and it will select on that function return value.

At the moment it only works for functions that return tables. Additional work is needed for primitives, composite types, etc. However, it should be pretty straightforward to utilize the existing `postgresType()` function https://github.com/kyleconroy/sqlc/blob/5c5629192ec6bd5e71f991b83c4a00fae13aef6e/internal/codegen/golang/postgresql_type.go

kylecarbs:
@kyleconroy any feedback for merging this?

It may not be perfect, but enables a quite complex return function like:
```sql
-- Function for recursively iterating relationships.
CREATE FUNCTION recursive_relationships(aid VARCHAR(20)) RETURNS SETOF relationships AS $func$ BEGIN RETURN QUERY WITH RECURSIVE rels AS (
    SELECT r1.*
    FROM relationships r1
    WHERE r1.actor = aid
    UNION
    SELECT r2.*
    FROM relationships r2
        INNER JOIN rels ON r2.actor_type = 'team'
        OR r2.actor_type = 'account'
) TABLE rels;

END $func$ LANGUAGE plpgsql;
```

I've tested with merging latest master.

kylecarbs:
Superseded by https://github.com/kyleconroy/sqlc/pull/964

kyleconroy:
Closed in favor of #964

PR #849, opened 2021-01-11
docs: Update README.md for emit_db_tags

uta-mori:
`emit_db_tags` option has been implemented, but it was not mentioned anywhere in the documentation, so I added it.


kyleconroy:
Thanks!

Issue #837, opened 2020-12-18
add invalid query comment error message


utamori:
It seems that the almost error handling is done.
but the following two typo cases error handling are missing in this.

- without a space after `--`
    - `--name: {queryname} :{cmd}`    
- without `:` after `name`
    - `-- name {queryname} :{cmd}`

https://github.com/kyleconroy/sqlc/blob/master/internal/metadata/meta.go

_Originally posted by @uta-mori in https://github.com/kyleconroy/sqlc/issues/836#issuecomment-747948485_

PR #859, opened 2021-01-24
Better query metadata errors (#837)

henrybaxter:


henrybaxter:
@kyleconroy this pull request also includes tests, let me know if there is a problem, happy to fix, thanks!

Issue #851, opened 2021-01-12
PRIMARY KEY at end of table definition results in nullable value


abagshaw:
When specifying `PRIMARY KEY` inline:
```
CREATE TABLE authors (
  id   BIGSERIAL PRIMARY KEY,
  name text      NOT NULL,
  bio  text
);

-- name: GetAuthor :one
SELECT * FROM authors WHERE id = $1 LIMIT 1;
```

The `ID` field is assigned `int64` as expected.
```
type Author struct {
	ID   int64          `json:"id"`
	Name string         `json:"name"`
	Bio  sql.NullString `json:"bio"`
}
```

---

But...when specifying `PRIMARY KEY` at the end of the table (not inline):
```
CREATE TABLE authors (
  id   BIGSERIAL,
  name text      NOT NULL,
  bio  text,

  PRIMARY KEY (id)
);

-- name: GetAuthor :one
SELECT * FROM authors WHERE id = $1 LIMIT 1;
```

The `ID` field is now assigned a nullable type `sql.NullInt64` even though it isn't nullable.
```
type Author struct {
	ID   sql.NullInt64  `json:"id"`
	Name string         `json:"name"`
	Bio  sql.NullString `json:"bio"`
}
```

kyleconroy:
Fixed!

PR #860, opened 2021-01-24
Primary key constraint forces not null (#851)

henrybaxter:
Feedback appreciated, also thanks for the fantastic library!


kyleconroy:
Fixes #851

kyleconroy:
Thanks for the first-time code! Really appreciate you adding a test as well

PR #862, opened 2021-01-26
Override globs

zelch:
So, I'm going to start with the use case.

Our MariaDB database has a lot of KSUIDs, enough that I would be better off writing a tool to automate building the override statements than doing it by hand.

It would be nice if we could just override on type, replace 'binary(20)' with 'github.com/segmentio/ksuid.KSUID' and call it a day.

Sadly, at least the mysql parser turns 'binary(20)' into 'char' well before the override code gets it.  In fact, it does this well before any of the sqlc code gets it.

That leaves doing it by column name, but as mentioned above...  There are quite a lot of them.

On the other hand, our naming convention is at least vaguely straight forward.  The columns all end with _ksuid, so we can automatically detect them.

But why write rules to do this, and have a huge override table, when we can just add globbing support to sqlc?

And so I did that instead. :)

This passes all of the tests, and is reasonably straight forward.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,7 @@ require (
 	github.com/antlr/antlr4 v0.0.0-20200209180723-1177c0b58d07
 	github.com/davecgh/go-spew v1.1.1
 	github.com/go-sql-driver/mysql v1.6.0
+	github.com/gobwas/glob v0.2.3
 	github.com/google/go-cmp v0.5.5
 	github.com/jackc/pgx/v4 v4.11.0
 	github.com/jinzhu/inflection v1.0.0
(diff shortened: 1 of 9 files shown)


kyleconroy:
Can you attach your sqlc.yaml or sqlc.json file? I'd like to see how this works in practice.

zelch:
Looks like I should update to version 2, but here you go.

[sqlc.yaml.txt](https://github.com/kyleconroy/sqlc/files/5871499/sqlc.yaml.txt)

zelch:
Ping on this @kyleconroy

zelch:
These changes have been rebased on the current head, specifically to add support for the python code generation.

kyleconroy (changes requested):
I like this change in theory. Before we can get it merged, there will need to be more tests and documentation. If you'd like to write those tests, great. If not, it may be a few more months before this gets merged.

kyleconroy:
@zelch Going to close this one out, as it has conflicts and no tests. I'd be happy to support this change, but it needs to have test coverage.

PR #864, opened 2021-01-26
Set default config filepath

JordanKnott:
Fixes an issue where the default config path is not used (`sqlc.yaml`)


kyleconroy:
The code should already handle the case where the flag value isn't specified. Was this failing for you?

kyleconroy:
Ignore my previous comment, I reproduced the issue

Issue #755, opened 2020-10-29
Support for custom models


mlukasik-dev:
Example:

`db/queries/roles.sql`
```sql
-- name: ListRolesWithPermissions :many
-- model: Roles
-- path: internal/domain/roles
SELECT
  r.role_id,
  r.display_name,
  r.description,
  array_agg(permissions.method)::text[] AS permissions
  FROM roles r
    LEFT JOIN permissions USING (role_id)
GROUP BY r.role_id;

-- name ListRoles :many
-- model: Roles
-- path: internal/domain/roles
SELECT  role_id, display_name, description FROM roles; /* returned colums exists in the model so we can reuse it */
```

It isn't generated file!
`internal/domain/roles`
```go
type Role struct {
	RoleID      uuid.UUID // here we can add some custom struct tags and methods
	DisplayName string
	Description string
	Permissions []sql.NullString // if array can is empty using []string throws error
}
```

Generated method
```go
func (q *Queries) ListRoles(ctx context.Context) ([]roles.Role, error)
```

P.S. This features can be useful because with it we can both use sqlc and follow **Clean Architecture**. And reuse structs for tables for complex queries with joins.

So my question is if I'd correctly implement this functionality would you approve PR (I'm asking because this feature can be not in the spirit of the project)

mlukasik-dev:
This feature could help as solve issues like: #185 , #534 , #658

jonleopard:
Would also love this! I use sqlx and frequently use tags like   `db:"user_id" json:"id"` or `json:"password_confirm" validate:"eqfield=Password"`

Jille:
I agree this would be useful. Especially when you have a few queries with different WHERE clauses that return the same columns.

This proposal has one big question: How would we verify that the given struct matches the data types returned from the query?
* Do we assert they match exactly with the types sqlc would've chosen? That'd be safe, but I think a bunch of people here are expecting to use this to override exactly that? Also, if the types match 1:1, Go would allow you to cast it anyway.
* If we allow uncheck structs, wouldn't that defeat one of sqlc's USPs: That you don't have runtime failures for type mismatches.

Would love to hear your thoughts on this.

jamesleeht:
Also very interested in this feature. Here's another related issue: #444

PR #868, opened 2021-01-28
when we use snowflake generate id

Yunfei1982:
when we use snowflake generate id
add 'int64_to_string_tags: true' to config
will generate json tag like:
type User struct {
    ID int64 `json:"id,string"`
}

### DIFF
--- a/internal/codegen/golang/gen.go
+++ b/internal/codegen/golang/gen.go
@@ -362,6 +362,7 @@ type tmplCtx struct {
 	SourceName string
 
 	EmitJSONTags        bool
+	Int64ToStringTags   bool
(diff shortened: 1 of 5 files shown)


Yunfei1982:
when we use snowflake generate id
add 'int64_to_string_tags: true' to config
will generate json tag like:
type User struct {
ID int64 json:"id,string"
}

kyleconroy:
@Yunfei1982 Thanks for the pull request.

Sadly, this solution isn't general enough to get merged. There are many different types and columns that may require different JSON tags. Instead, I think the correct solution here is to support custom models #755.

PR #905, opened 2021-02-22
feat(json-tags): Add case style config option

### LINKED ISSUE #878: camelCase
Would there a possibility to get camel casing behind a flag? 
This would be great feature in few cases. For example a web-app which uses multiple APIs and some of them return camel case. Would be great if everything would use the same casing.

chutified:
This PR allows users to choose which case style will be used to generate JSON tags.
The style can be selected by adding the `json_tags_case_style` setting key.

- camelCase - "camel"
- PascalCase - "pascal"
- snake_case - "snake" (default)

#### Changes

- Add `JSONTagsCaseStyle` field to package settings and set its default value to `snake`.
  - `config/config.go`
  - `config/v_one.go`

- Implement `SetCaseStyle` function which converts the tag name to the specified case style.
  - `codegen/golang/field.go`

- Wrap JSON tags name with the `SetCaseStyle` function.

- Add test for each case style.
  - `endtoend/testdata/camel_case/...`
  - `endtoend/testdata/pascal_case/...`
  - `endtoend/testdata/snake_case/...`

- Add `json_tags_case_style` key to the documentation.

Resolve #878


kyleconroy (changes requested):
First off, the code looks great. Thank you for updating the documentation as well as writing tests. 

However, this change includes two breaking chages.

First, sqlc currently doesn't format the JSON tags in any particular case. The default is to just use the name of the column in the database. This behavior needs to stay the default.

Second, renames do not currently apply to JSON tags. I don't think we should start doing that now.

If you could fix up these two things, I can merge this.

chutified:
> The default is to just use the name of the column in the database. This behavior needs to stay the default.

I removed the renames, a sure thing, I totally agree.

>  renames do not currently apply to JSON tags. I don't think we should start doing that now.

I added a wrapper which applies the case style formatter only if the `json_tags_case_style` is set. If its value is `none` or has no value (the key is omitted) the name of the column in the database is used (the default).

Changes:

- Removed the renames
- Added `JSONTagName` which decides if `SetCaseStyle` is applied:
```go
func JSONTagName(name string, settings config.CombinedSettings) string {
	style := settings.Go.JSONTagsCaseStyle
	if style == "" || style == "none" {
		return name
	} else {
		return SetCaseStyle(name, style)
	}
}
```
- Changed the function pattern of the `SetCaseStyle`:
```diff
- func SetCaseStyle(name string, settings config.CombinedSettings) string {
+ func SetCaseStyle(name string, style string) string {
```
- Updated the documentation

kyleconroy (approved):

PR #924, opened 2021-03-04
workflows: Only run tests once

kyleconroy:


kevinburkemeter:
I would simply not run the tests multiple times

Issue #933, opened 2021-03-09
Creates bad code when enums are similar with special characters.


ovadbar:
When enums are slightly different except for a value for example

```
CREATE TYPE "digit" AS ENUM (
  '0',
  '1',
  '2',
  '3',
  '4',
  '5',
  '6',
  '7',
  '8',
  '9',
  '#',
  '*'
);
```

Will generate
```
type Digit string

const (
    Digit0  Digit = "0"
    Digit1  Digit = "1"
    Digit2  Digit = "2"
    Digit3  Digit = "3"
    Digit4  Digit = "4"
    Digit5  Digit = "5"
    Digit6  Digit = "6"
    Digit7  Digit = "7"
    Digit8  Digit = "8"
    Digit9  Digit = "9"
    Digit Digit = "#"
    Digit Digit = "*"
)
```

This is will not work for 2 reasons.

PR #934, opened 2021-03-09
Fix for #933 

ovadbar:
This should append digits to the end of the enum so it will generate valid code. 
Tests TBD


kyleconroy:
Fixes #933

Issue #812, opened 2020-11-25
Option for Context-free and Pointer returning struct


kokizzu:
Currently the generated values are:
```
func (q *Queries) BuffById(ctx context.Context, id int64) (Buff, error) {
```

it would be nice if there's option to set these values:
```
   emit_context: false
   emit_pointer: true
```

so it would generate something like this:
```
func (q *Queries) BuffById(id int64) (*Buff, error) {
```

maratori:
Can you add use case of `emit_pointer: true`, please?

kokizzu:
when the struct is large (>word size of the compiler/os/processor), it's inefficient to copy everything by value (not using pointer), copy by pointer/reference is [faster](https://stackoverflow.com/questions/23542989/pointers-vs-values-in-parameters-and-return-values).

andrewmbenton:
Closing this as it's quite old and we already have `emit_result_struct_pointers`. To remove context from generated output I'd suggest forking and modifying [sqlc-gen-go](https://github.com/sqlc-dev/sqlc-gen-go), which is just sqlc's internal codegen package extracted into an external plugin.

PR #952, opened 2021-03-21
Add emit_pointer option for Go generator

ipsusila:
Partial fix for issue #812


kyleconroy (changes requested):
Thanks for starting on a fix for this. For a feature that touches many aspects of the generated Go code, you'll need to add a large set of end-to-end tests before we can merge this in.

My guess is that the current implementation in `gen.go` doesn't cover many of the edge cases with overrides.

kyleconroy:
Closed in favor of #979

PR #954, opened 2021-03-22
Fix init command fails to create config file with empty name

lukahartwig:
When no filename was provided through the `file` flag, the flag falls back to an empty string, not `nil`. The init command tries to create a config file with the empty file flag and fails.

As a fix, I set the default value in the flag. I looked through the code to see if anything depends on the empty string behavior but couldn't find anything.


kyleconroy:
Thanks for the PR. I don't actually want to set a default for the file parameter, so I fixed it a different way in #959

PR #964, opened 2021-04-05
sql/ast: Resolve return values from functions

### LINKED ISSUE #199: SELECT from FUNCTION return
```sql
CREATE FUNCTION foo() RETURNS bool AS $$ SELECT true $$ LANGUAGE sql;

-- name: FromFoo :one
SELECT * FROM foo();
```

Fails with the error `empty edit contents`

kylecarbs:
Closes #199.

Thanks to @warent for the prior art in #846!

Feedback is very welcomed. Not certain I'm doing this the best way...


cmoog:
cc @coadler

kyleconroy (changes requested):
The changes look great. Can you add a end-to-end test for this feature? There are a bunch of examples in the `internal/endtoend/testdata` folder. If you could add it for both MySQL and PostgreSQL that would be even better.

kylecarbs:
@kyleconroy MySQL returns single-types from functions, so this isn't really applicable. However, I did add a PostgreSQL test. 🥳

kylecarbs:
@kyleconroy anything else needed for this to merge?

kyleconroy (approved):

kyleconroy:
Nope!

PR #976, opened 2021-04-12
Handle convertPatternInExpr.

zelch:
This can be handled just by calling convert on the subquery, and it is
necessary for some kinds of queries.

Like the following, without this we don't pickup the parameter, and thus
can't handle the query at all:

select * from foo where field in (select substring(?, 1, seq) from seq_1_to_19)

### DIFF
--- a/internal/engine/dolphin/convert.go
+++ b/internal/engine/dolphin/convert.go
@@ -873,7 +873,7 @@ func (c *cc) convertPartitionByClause(n *pcast.PartitionByClause) ast.Node {
 }
 
 func (c *cc) convertPatternInExpr(n *pcast.PatternInExpr) ast.Node {
-	return todo(n)
+	return c.convert(n.Sel)
 }
 
 func (c *cc) convertPatternLikeExpr(n *pcast.PatternLikeExpr) ast.Node {


kyleconroy:
I'm sorry I didn't get around to review this PR until now.

I can't merge this PR in its current form. The `PatternInExpr` struct has four fields: `Expr`, `List`, `Not` and `Sel`. We need to take into account all four of these fields. Just returning `c.convert(Sel)` won't generate correct code.

zelch:
Completely understood, and after staring at it for a short while, that all makes sense to me.

With that said, I'm running into a problem trying to fix it for the Expr and List cases, specifically, sqlc does not know what type the parameters should be, and using the mysql engine I can't see any way to give sqlc that information.

For the postgresql engine, ?::int does the job, but mysql has no similar syntax.

Any suggestions on how I might go about addressing this?

asterikx:
@zelch have you made any progress on this? I'm very interested in supporting the `IN` clause for MySQL as well.

I also found this implementation in a fork of sqlc:
https://github.com/xiazemin/sqlc/blob/837db7ef88e628b3bd7926d7c02166bfd3dc8130/internal/engine/dolphin/convert.go#L906-L945

zelch:
@asterikx I have not yet had the time to do a deep dive to see if I can convince it to get the types right for the non-subexpression cases, my initial try simply failed later on when the code generator couldn't find what type the parameters were supposed to be.

I definitely need to take a deeper look at that fork when I get the spare moments though.

PR #979, opened 2021-04-21
Add emit_result_struct_pointers and emit_params_struct_pointers options

dchertkov:
Add emit_result_struct_pointers and emit_params_struct_pointers options for Go generator.

This is second attempt at improvement as noted here #812


jwilner:
Would love to see this landed soon! It'd make the generated types much more idiomatic: https://github.com/golang/go/wiki/CodeReviewComments#receiver-type

jwilner:
@kyleconroy anything we can do to help get this merged? Using pointer types would be a lot more idiomatic imo.

kyleconroy:
@dchertkov I'm so sorry reviewing this pull request fell through the cracks. If you're willing to fix the merge conflicts and update to the latest code, I can promise to review it as soon as it's ready. Let me know.

dchertkov:
@kyleconroy All is ready.

internal/codegen/golang/query.go
```diff
@@ -30,6 +36,9 @@ func (v QueryValue) Pair() string {
 	if v.isEmpty() {
 		return ""
 	}
+	if v.EmitPointer && v.Struct != nil {
```
kyleconroy, line 39: Use `v.IsPointer()` instead.

internal/codegen/golang/templates/pgx/interfaceCode.tmpl
```diff
@@ -2,10 +2,10 @@
     type Querier interface {
     {{- range .GoQueries}}
         {{- if eq .Cmd ":one"}}
-            {{.MethodName}}(ctx context.Context, {{.Arg.Pair}}) ({{.Ret.Type}}, error)
+            {{.MethodName}}(ctx context.Context, {{.Arg.Pair}}) ({{if .Ret.IsPointer }}*{{end}}{{.Ret.Type}}, error)
```
kyleconroy, line 5: Instead of updating every reference to `.Ret.Type` and `.Ret.Name`, we should add a new method that takes into account `IsPointer` and returns the correct value.

internal/endtoend/testdata/emit_result_and_params_struct_pointers/query.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE foo (a integer, b integer);
+
+/* name: GetOne :one */
+SELECT * FROM foo WHERE a = ? LIMIT 1;
+
+/* name: GetAll :many */
+SELECT * FROM foo;
+
+/* name: GetAllAByB :many */
+SELECT a FROM foo WHERE b = ?;
```
kyleconroy, line 10: This does not test the case where the params struct should be a pointer to a struct.

internal/endtoend/testdata/emit_result_and_params_struct_pointers/go/query.sql.go
```diff
@@ -0,0 +1,74 @@
...
+import (
+	"context"
+	"database/sql"
+)
+
+const getAll = `-- name: GetAll :many
+SELECT a, b FROM foo
+`
+
+func (q *Queries) GetAll(ctx context.Context) ([]*Foo, error) {
```
kyleconroy, line 15: The slice of structs seems weird to me, but I guess that's what you'd want in this case?

kyleconroy (changes requested):
Overall this approach looks good. A few things to clean up before we can get this merged.

internal/endtoend/testdata/emit_result_and_params_struct_pointers/go/query.sql.go
dchertkov: There has been some discussion about performance https://stackoverflow.com/questions/27622083/slices-of-structs-vs-slices-of-pointers-to-structs

kyleconroy (approved):
We'll need to update the docs with the new configuration options, but that can be done in a separate PR.

PR #983, opened 2021-04-26
Support for nullable fields in joins

jwilner:
- Builds on top of https://github.com/kyleconroy/sqlc/pull/733
- Properly detects left, right, full joins
- Uses proper enum ordering

PR #984, opened 2021-04-26
feat: Add embedded structs based on result match

kylecarbs:
Identifies every column with a returned struct for queries.

This is a breaking change. I'm very open to comments, ideas, or reasons why this shouldn't even happen. To prevent breakage we could move this under a configuration option, and eventually migrate to default in a major version.

```sql
-- schema.sql
CREATE TABLE users (id text, email text);
CREATE TABLE usernames (user_id text, name text);

-- query.sql
-- name: GetUsers :many
SELECT * FROM users;

-- name: GetUsersWithUsername :many
SELECT users.*, usernames.name FROM users JOIN usernames ON users.id = usernames.user_id;
```

Before:

```go
type GetUsersWithUsernameRow struct {
	ID    sql.NullString `db:"id"`
	Email sql.NullString `db:"email"`
	Name  sql.NullString `db:"name"`
}
```

After:

```go
type GetUsersWithUsernameRow struct {
	User
	Name sql.NullString `db:"name"`
}
```


cmoog:
Run `make regen` at the root to regenerate the expected output for e2e tests.

cmoog:
The following case is potentially a problem
```sql
CREATE TABLE foo (a text, b text);
CREATE TABLE bar (c text, d text);

/* name: LeftJoin :many */
SELECT foo.*, bar.* FROM foo LEFT JOIN bar on foo.a = bar.c;
```

Using embedded structs means we can't use the `sql.Null*` types for columns joined via a left join.

## Output from this PR
```go
const leftJoin = `-- name: LeftJoin :many
SELECT foo.a, foo.b, bar.c, bar.d FROM foo LEFT JOIN bar on foo.a = bar.c
`

type LeftJoinRow struct {
	Foo
	Bar
}

func (q *Queries) LeftJoin(ctx context.Context) ([]LeftJoinRow, error) {
	rows, err := q.db.QueryContext(ctx, leftJoin)
	if err != nil {
		return nil, err
	}
	defer rows.Close()
	var items []LeftJoinRow
	for rows.Next() {
		var i LeftJoinRow
		if err := rows.Scan(
			&i.Foo.A,
			&i.Foo.B,
			&i.Bar.C,
			&i.Bar.D,
		); err != nil {
			return nil, err
		}
		items = append(items, i)
	}
		return nil, err
	}
		return nil, err
	}
	return items, nil
}

```

## Output on `master`
```golang
const leftJoin = `-- name: LeftJoin :many
SELECT foo.a, foo.b, bar.c, bar.d FROM foo LEFT JOIN bar on foo.a = bar.c
`

type LeftJoinRow struct {
	A sql.NullString
	B sql.NullString
	C sql.NullString
	D sql.NullString
}

func (q *Queries) LeftJoin(ctx context.Context) ([]LeftJoinRow, error) {
	rows, err := q.db.QueryContext(ctx, leftJoin)
	if err != nil {
		return nil, err
	}
	defer rows.Close()
	var items []LeftJoinRow
	for rows.Next() {
		var i LeftJoinRow
		if err := rows.Scan(
			&i.A,
			&i.B,
			&i.C,
			&i.D,
		); err != nil {
			return nil, err
		}
		items = append(items, i)
	}
		return nil, err
	}
		return nil, err
	}
	return items, nil
}
```

cmoog:
Didn't realize the join handling hadn't reached master, but this will be a problem when https://github.com/kyleconroy/sqlc/pull/983 lands.

kylecarbs:
The embedded struct shouldn't be nullable though... right? It doesn't represent any data itself, simply a container for data.

cmoog:
Correct. But the problem comes when `Foo.A` is not null in the data model, but can be null after a left join. When we marshal a possibly nullable field to `Foo.A` directly this nullability type information is lost. Instead we'd want to use `sql.NullString` for `Foo.A` in that particular query. More context here: https://github.com/kyleconroy/sqlc/pull/983

kylecarbs:
Ahh understood.

cmoog:
bump

kyleconroy (changes requested):
This PR is now out of date, there are merge conflicts.

Also, I'm not sure how you're planning on solving the LEFT JOIN issue. You can't use the existing structs because a non-nullable column can now be null. I think you'll need to generate new structs, which may defeat the entire purpose. Maybe we just don't allow embedding to work when using left joins?

Lastly, the syntax for this needs to be a special sqlc function. I'd suggest using `sqlc.embed()`. The reason is that overloading the `table.*` method isn't backwards compatible.

kyleconroy:
@kylecarbs I haven't heard from you in a while, so I'm going to close this out. If you end up working on this, please open a new pull request or push to this one and mention me in a comment.

ludusrusso:
@kylecarbs I'm interesting in this feature. If you are not working on this anymore I can take it!

kylecarbs:
Feel free! I'd love to tackle it but don't quite have bandwidth at the moment.

PR #985, opened 2021-04-28
Better errors

ovadbar:
While this is kind of a hack it should produce better errors when there is a syntax error instead of just returning line 1 column 1 .

Ideally we would break the files up before running `c.parser.Parse` on the entire file, but I did not want to mess with the entire functionality with this.


kyleconroy:
@ovadbar Can you provide more examples of the bad errors and how this changes fixes them? It's not obvious from the code. We also need to make sure this works with both MySQL and PostgreSQL.

kyleconroy:
@ovadbar I never heard back from you. Without tests or examples, I can't accept this PR in its current state. Closing for now.

PR #998, opened 2021-05-03
Customizable output file names

adnsio:
This PR adds the possibility to customize the output file names, by default everything remains as before so it's transparent for all users.

Example configuration:
```
version: "1"
packages:
  - name: db
    path: internal/db
    queries: internal/db
    schema: migrations
    engine: postgresql
    emit_prepared_queries: true
    emit_interface: false
    output_db_file_name: db_gen.go
    output_models_file_name: model_gen.go
    output_querier_file_name: querier_gen.go
    output_files_suffix: _gen.go
```


kyleconroy (changes requested):
Thanks for the change! Can you please add a test in the `testdata` directory that uses your new configuration options?

kyleconroy (approved):
Thanks for adding tests

adnsio:
If I find time in the next week I would also like to add a few lines in the docs about the new options 😄

PR #1016, opened 2021-05-14
Add support for :iter queries

pcriv:
Just took the work from @eivindr here https://github.com/eivindr/sqlc/commit/64043b38bde83656934fb6354eed0711eb3b07a7 and made an MR.

### DIFF
--- a/internal/codegen/golang/gen.go
+++ b/internal/codegen/golang/gen.go
@@ -163,6 +163,9 @@ type Querier interface {
 	{{- if eq .Cmd ":execresult"}}
 	{{.MethodName}}(ctx context.Context, {{.Arg.Pair}}) (sql.Result, error)
 	{{- end}}
+	{{- if eq .Cmd ":iter"}}
+	{{.MethodName}}(ctx context.Context, iter func({{.Ret.Name}} {{.Ret.Type}}) error, {{.Arg.Pair}}) error
+	{{- end}}
 	{{- end}}
 }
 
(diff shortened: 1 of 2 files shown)


kyleconroy:
@pcriv Thanks for pulling the code into a PR. Instead of supporting iterators, I encourage folks to write SQL queries that utilize paging. I don't think enough people need iterators to add it as a supported query type.

eivindr:
@pcriv Going forward I'll try to keep https://github.com/eivindr/sqlc/tree/streaming in sync with the latest release (atm its in sync with main). The branch now also contains :iter support for Kotlin

pcriv:
@eivindr cool! A bit disappointed this didn't get merged.
@kyleconroy any chance you would accept the feature if it were behind a specific flag? Like, only generate the iterators if specifically requested via the command line? I get it that it's recommend to use pagination but in this case our only option is to use streams. In any case I respect your decision :)

eivindr:
Kotlin plugin with :iter support is available at https://github.com/eivindr/sqlc-gen-kotlin

You need to manually build it, use "file://" to reference the local wasm file in the plugin "url" config. The sha256 hash outputted by the make build also need to be used in the config.

tyro-jason:
Hi folks - I had a simiar PR for streaming - the use case is having to stream large datasets to clients which... you could page but is way more efficient in terms of chattiness to stream right from the DB to the API consumer stream without having to allocate massive slices / lists

kyleconroy:
We're continuing the conversation about streaming here: https://github.com/sqlc-dev/sqlc/issues/720#issuecomment-1731742748. The current plan is to land `:iter` support once the range-over-func experiment has been accepted and enabled in Go.

PR #1043, opened 2021-06-10
Better handling for parameters in simple json queries

### LINKED ISSUE #743: Paramaters matched to JSON fields are the wrong type
These parameters should end up as `interface{}`, not `json.RawMessage`.

```sql
CREATE TABLE "user" (
    "id" INT PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
    "metadata" JSONB
)

-- name: FindByAddress :one
SELECT * FROM "user" WHERE "metadata"->>'address1' = $1 LIMIT 1
```

```go
type FindByAddressParams struct {
	metadata json.RawMessage
}
```

- [Initial report](https://github.com/kyleconroy/sqlc/discussions/738?converting=1)

https://play.sqlc.dev/p/cd6227106fb8fb969d1fc9699dd9c03046e26578c589268e30d23573f67037bb

robholt:
Fixes #743 (for common cases)

The original issue says the types should always be `any` (which is mapped to `interface{}` in go), but I thought it would be better to use the return type for each operator (for example, `col_name ->> 'field_name'` returns the `text` type).

I also added support for type conversions as a way to "force" the correct type for situations that still aren't handled.


kyleconroy:
@robholt I think we need to come at this from a different approach. The problem is that the `->>` operators are PostgreSQL-specific. We need a way for the database engine to declare these operators and their return types. Adding hard-coded support will cause issues for people using MySQL, as their queries might compile.

kyleconroy:
I'm going to close this out as I'm pretty sure this isn't the approach we want to take. Instead, we should mimic how the function resolution logic works by looking up operators in the catalog to determine the return value.

PR #1044, opened 2021-06-10
Add support for Views

### LINKED ISSUE #193: View support

robholt:
Fixes #193 

This relies on the existing parsing of `SELECT` queries, so any `SELECT` that sqlc can normally determine the output columns for will work as a view.


kyleconroy (approved):

mrsufgi:
@robholt I want to implement this for materialized views, after reviewing the changes here I wonder, what is the key difference? Why can't we use the exact same behavior for materialized views?

PR #1057, opened 2021-06-23
Add skip_db_file_name option

yevgenypats:
This will skip the default generation of db.go.
This can be useful when you want to define your own db.go


kyleconroy:
@yevgenypats This feature can be implemented by wrapping the command that runs sqlc.

```
sqlc generate
cp ./path/to/other/db.go ./path/to/generated/db.go
```

Since the work around is simple, I won't be merging this feature in.

Issue #546, opened 2020-06-12
Data validation


tobiges:
Is there any way to implement data validation with sqlc?

kyleconroy:
The suggested way to implement data validation when using sqlc is to take advantage of database constraints. The [PostgreSQL](https://www.postgresql.org/docs/current/ddl-constraints.html) documentation covers the topic extensively.

Issue #534, opened 2020-06-05
Add support for custom struct tags


philippta:
When working with structs, struct tags are used to add additional information to a field. This is for example very common when using validation packages. But currently there is no way to add additional struct tags to take advantage of this. 

I think it would be a great idea to add a comment syntax just like the `-- name:NameForQuery` for table columns. 

A possible option to express these custom struct tags would be to write create table statements like this:
```sql
CREATE TABLE user (
  id   BIGSERIAL PRIMARY KEY,

  -- tags:`validate:"min=10,max=0"`
  username text NOT NULL,

  -- tags:`validate:"gt=18"`
  age  integer NOT NULL,

  -- tags:`validate:"url"`
  website text NOT NULL
);
```

or alternatively:
```sql
CREATE TABLE user (
  id   BIGSERIAL PRIMARY KEY,
  username text NOT NULL,     -- tags:`validate:"min=10,max=0"`
  age integer NOT NULL,       -- tags:`validate:"gt=18"`
  website text NOT NULL       -- tags:`validate:"url"`
);
```

huangc28:
Has anyone pickup this enhancement yet?

fr3fou:
Is there a way to achieve this now?
I want to ignore the password field on my user struct, like so
```go
type User struct {
    Password string `json:"-"`
}
```

eliaschampi:
any solution for this issue please?

tamoore:
@kyleconroy Would you be open to a solution that utilised the config overrides?

Something like the following:

```yaml
overrides:
  - column: "chat.id"
    custom_tag_key: "gorm"
    go_type: 
      type: "int"
```

I think this might also need something like `custom_tag_value` for instances like `validate` where consumers want to set a custom value. Otherwise, the default would be whatever the json emit type is set to. 

Further to the above, this wouldn't override json tags, it would be in addition to emitting json tags.

preslavrachev:
@kyleconroy can we hope to see a feature like this in the foreseeable future? In my project, we started using Gorm way back in time, and it has been something we have been considering moving away from (towards sqlc). However, we rely on some 3rd party code that uses Gorm under the hood, so completely running away from it isn't feasible at the moment.

If there were a way to let sqlc generate our Gorm-compatible models, this would be fantastic. It's the one thing that's stopping us from getting all in on sqlc.

The only other alternative is still keep two copies of each model, which isn't that bad, but possibly error-prone.

austincollinpena:
[...]

PR #1064, opened 2021-06-27
feat: add custom tags

ymzuiku:
sqlc is very perfect database model generation tool, we use sqlc and gorm(use dynamic query).
This pull request is extends use sql comment custom go struct tags it's style:

Schema example code in comment:

```
@<tagName>:<tagValue>
```

## Examples:

Add comment in SQL(Postgres):

```sql
COMMENT ON COLUMN account.id is '@gorm:primaryKey @validate:required,min=3,max=32';
```

Generate go struct code:

```go
type Account struct {
  ID  int32  `gorm:"primaryKey" json:"id" validate:"required,min=3,max=32"`
}
```

### DIFF
--- a/internal/codegen/golang/custom_tags.go
+++ b/internal/codegen/golang/custom_tags.go
@@ -1,0 +1,55 @@
+package golang
+
+import (
+	"fmt"
+	"regexp"
+	"strings"
+
+	"github.com/kyleconroy/sqlc/internal/sql/catalog"
+)
+
+const tagKeyValueSplit = "__@CUSTOM_TAG_VALUE__"
+
+var tagReg = regexp.MustCompile(`^@.*\:`)
+
+// Comment parse go struct tags
+//
+// Example:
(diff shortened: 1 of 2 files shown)


kyleconroy:
@ymzuiku Thank you very much for taking the time to code up this feature. Sadly, I won't be accepting this PR. If we add support for custom fields on a per-column basis, I'd like to make the mechanism more general so that it can be applied to any database engine or language backend.

imraan-go:
>  I'd like to make the mechanism more general so that it can be applied to any database engine or language backend.

Closing a PR just because he could not provide solution for all language and database engine is a bad idea. This feature was requested by many including me for Golang + Postgres. I think this PR should be merged so other developers are encouraged to provide support for other languages and database engines. These PR will solves the following issues:

https://github.com/kyleconroy/sqlc/issues/546
https://github.com/kyleconroy/sqlc/issues/534
https://github.com/jschaf/pggen/issues/24

imraan-go:
Also I'm not sure if other language has tag support like Golang?

PR #1071, opened 2021-06-30
Mix parameters2

ovadbar:
I did not realize the previous merge requests tests were not passing. 
I made some additional changes and it seems like the tests pass.

If there is anything that needs to be done. I can take a look with my limited time.


internal/endtoend/testdata/mix_param_types/mysql/go/test.sql.go
```diff
@@ -0,0 +1,57 @@
...
+SELECT count(1) FROM bar WHERE id = ? AND name <> ?
+`
+
+type CountOneParams struct {
+	Name string
+	ID   int64
+}
+
+func (q *Queries) CountOne(ctx context.Context, arg CountOneParams) (int64, error) {
+	row := q.db.QueryRowContext(ctx, countOne, arg.Name, arg.ID)
```
kyleconroy, line 20: This generated code is incorrect. The order of the parameters in the query should be `arg.ID, arg.Name`.

kyleconroy (changes requested):

internal/endtoend/testdata/mix_param_types/mysql/go/test.sql.go
ovadbar: I updated the test and the code to pass the test.

kyleconroy:
Closed in favor of #1072

PR #1072, opened 2021-07-01
Allow for mixed parameters types ($1 or ?) and sqlc.arg()

ovadbar:
This builds on pull request #1071. I realized there were issues when postgresql was using ? format so I added support for that format. I figure it will be useful for porting sqlc code from mysql to postgres. What I prohibit is mixed use of $1 and ? formats.


internal/sql/validate/param_style.go
```diff
@@ -12,21 +12,40 @@ import (
...
+		fc, ok := f.(*ast.FuncCall)
+		if ok {
+			/*
+				if len(fc.Args.Items) != 1 {
+					return &sqlerr.Error{
+						Code:    "", // TODO: Pick a new error code
+						Message: fmt.Sprintf("expected 1 parameter to sqlc.arg; got %d", len(fc.Args.Items)),
+					}
+				}
+			*/
```
kyleconroy, line 26: We shouldn't leave uncommented blocks of code around.

internal/sql/validate/param_style.go
```diff
@@ -12,21 +12,40 @@ import (
...
+			switch fc.Args.Items[0].(type) {
```
kyleconroy, line 27: You can use a type switch here to improve the code so you don't have to type assert three times.

kyleconroy (commented):
I left a few comments, but please let me know if you're free to work on this. If not, I'll merge it as-is and clean it up.

ovadbar:
I think I made the changes last night.

PR #1079, opened 2021-07-10
Separate Prepare and Queries.Prepare methods

carzil:
Hello!

This patch introduces a new method Prepare for Queries structure in backward-compatible manner.

Motivation: our dependency injection framework requires objects to be created first as plain structures. Then, in a separate step all network initialisations are done using context.Context with appropriate timeout/deadline is passed. So, we need to create Queries first and then Prepare it.


carzil:
Gentle bump :)
@kyleconroy

kyleconroy (changes requested):
Apologies for the extremely slow review on this. I haven't had much time lately to work on features, triage bugs, and review code.

This change is sadly not safe to make. The reason is that user-defined methods get added onto the `Queries` struct. If someone has an existing `Prepare` query, this PR will break their code. It's arguable that adding the `Close` and `WithTx` methods were a mistake as well.

Can you explain a bit more on why you need the `Prepare` method in the first place? I'm wondering if there's something we can do to solve your problem.

PR #1088, opened 2021-07-16
Handle convertPatternInExpr.

zelch:
This is a second attempt at #976 , but now with support for Expr, List, and Not in addition to Sel.

This is used for queries like the following:

SELECT * from foo where field in (select substring(?, 1, seq) from seq_1_to_19);

SELECT a, b from foo where foo.a in (select a from bar where bar.b = ?);

SELECT a, b from foo where foo.a in (?, ?);

SELECT a, b from foo where foo.a not in (?, ?);

SELECT a, b from foo where ? in (foo.a, foo.b);

The support is heavily based on that found in the fork found at:
https://github.com/xiazemin/sqlc.git

However it should be noted that this is not just a straight copy and
paste job, there are several pieces from that branch for features that
don't exist in the main sqlc repository, the sqlc repository has changed
a bit since the fork, and the fork version does not handle the latter
case at all.


zelch:
@asterikx pinging you on this one given the conversation on the previous merge request.

zelch:
Well, oops.  Please hold off on reviewing this, the code it generates won't compile, as noted in the pipeline checks.

I'll take a crack at resolving that sometime in the next week, I hope.

zelch:
Alright, so a bug in the golang code generator has been fixed, I'm not sure how easy it would have been to trigger this previously, but it is far too easy to trigger it with this change set.

And after merging the code with master, a logical conflict (lack of support for DefaultSchema in the pre-merge code) has been fixed.

At this point I _think_ this should be good to go.

kyleconroy:
@zelch This code is now out of date with `main`. Also, there are still no tests for PostgreSQL. Do you still have time to work on getting this across the line? If not, I'll close it out and work on `IN` support in a different PR.

zelch:
I can definitely get it back up to date with main.

As far as Postgres tests, I'm afraid that I only implemented support for dolphin.  Adding support for the postgresql engine would definitely be good to do, but I'm not sure that I'm going to have time in the immediate future to tackle it.

kyleconroy:
Great, let's get it updated and then we can merge in support as-is.

zelch:
Done, and it is passing all tests.

Issue #1007, opened 2021-05-06
UTF-8 with BOM encoding broken: syntax error at or near ""


Git-me-outta-here:
All of my migrations files are UTF-8 except for one, which is UTF-8 with BOM.

That's where I get `syntax error at or near ""`... Could you allow sqlc to read this encoding?

kyleconroy:
My plan is to create an internal `utfbom` package that has one function, DiscardBOM. It will take a byte slice and return a byte slice, which should work for all the cases we read in SQL files.

bodison:
yeah would be fine as well. or a package that wraps the file reader?

As I mentioned in the closed PR, the test coverage of these four files would be a separate issue.

PR #1101, opened 2021-07-31
Update parse.go

bodison:
fixes https://github.com/kyleconroy/sqlc/issues/1007


bodison:
Thanks for reviewing this. I'll add a test tomorrow :)

bodison:
I added tests and I added the BOM removal to a few more locations.

But I'm not happy with the test coverage.

bodison:
I'm worried that the three tests I added seem to only test one of these places:
internal/compiler/compile.go:108

It does not test:
internal/compiler/compile.go:71
internal/engine/dolphin/parse.go:60
internal/engine/postgresql/parse.go:162
internal/engine/sqlite/parse.go:51

Am I mistaken?

kyleconroy:
@bodison You're right that we need to solve this in a more general way. I'm going to close this PR and update the linked issue with better description of the issue and a few possible implementations.

bodison:
@kyleconroy I wouldn't say we have to solve it in a more general way. IMO removing the BOM is a very niche issue and can be done whenever you open a file to read.

The only thing I do not like/understand is the test coverage of the four files I edited. But I'm afraid I do not have a contribution to improve this, since I barely know the code base.

The original issue would be solved with this PR, the insufficient test coverage should be a different issue IMO.

<3

PR #1113, opened 2021-08-11
compiler: Handle subqueries in SELECT statements

### LINKED ISSUE #1042: Query with COUNT() subselect generates bool field instead of int
Field gets invalid type which in my case causes runtime error when scanning into struct:
`sql: Scan error on column index 6, name "orders_count": sql/driver: couldn't convert 14 into type bool`

I also managed to reproduce it in the sqlc playground:

query.sql
```sql
CREATE TABLE authors (
  id   BIGSERIAL PRIMARY KEY,
  name text      NOT NULL,
  bio  text
);

CREATE TABLE books (
  id BIGSERIAL PRIMARY KEY,
  author_id BIGSERIAL NOT NULL
  	REFERENCES authors(id),
  title text NOT NULL
);

-- name: GetAuthorsWithBooksCount :many
SELECT *, (
  SELECT COUNT(id) FROM books
  WHERE books.author_id = id
) AS books_count
FROM authors;
```

Code generated:
image

Noticed interesting thing: if i remove AS statement (column alias) it will give this field name `Exists`
```go
type GetAuthorsWithBooksCountRow struct {
	ID     int64
	Name   string
	Bio    sql.NullString
	Exists bool
}
```
I think it may help fix this bug.

kyleconroy:
Fixes #1042


kevinburkemeter:
Hooray!

PR #1119, opened 2021-08-12
Add hack to fix output type for coalesce

### LINKED ISSUE #1096: COALESCE returns a NullInt instead of an int in JOIN
If I have `COALESCE(connected_clients, 0)`, sqlc generates a NullInt64 for this instead of an int64. This is a regression from 1.7.0. (At least I can fix it by casting to the right value).

kyleconroy:
Add flag to skip table required check for joins when using COALESCE.

Obviously this is pretty gross hack. We really need to refactor how the type checking in sqlc works.

Fixes #1096


kevinburkemeter:
Thanks, Kyle!

PR #1204, opened 2021-09-21
Exportable Run function

### LINKED ISSUE #416: Exportable libraries for embedding sqlc functionality in other cli apps
Sometimes larger companies have their own tool chains including custom CLI apps for generating new projects, etc. I would love to add sqlc functionality to our custom CLI app, but most of the tool chain is behind the `internal` folder.  A good example of a CLI tool but that can also be easily embeddable in other projects is the Goose db migrations tool: https://github.com/pressly/goose 

Are you opposed to me just forking the project and messing around with the structure a bit to see if I could get it to fit in our custom tool chain? 

An example of something we would want to explore doing at Weave: We start with a proto definition that defines the data model, so I want to generate the sqlc functions to use the proto generated model instead of generating its own (I actually have this working as a POC but it is a hack).

oliverpool:
Closes #416

This PR adds an exported `Run` function to the root package.

This allows this library to be simply embedded into other cli apps.

Usage is just like https://github.com/kyleconroy/sqlc/blob/main/cmd/sqlc/main.go :
```go
package main

import (
	"os"

	"github.com/kyleconroy/sqlc"
)

func main() {
	os.Exit(sqlc.Run(os.Args[1:], os.Stdin, os.Stdout, os.Stderr))
}
```

This does not export any internal types, but simply exposes the same interface as the CLI to any go program.

---

In my case, I have a `tool` folder in my go project with this file, so that I can run `go run ./tool/sqlc.go` and be sure to have the right version from the project's `go.mod`

### DIFF
--- a/placeholder.go
+++ b/placeholder.go
@@ -1,6 +1,17 @@
 package sqlc
 
+import (
+	"io"
+
+	"github.com/kyleconroy/sqlc/internal/cmd"
+)
+
 // This is a dummy file that allows SQLC to be "installed" as a module and locked using
 // go.mod and then run using "go run github.com/kyleconroy/sqlc"
 
 type Placeholder struct{}
+
+// Run the sqlc command
+func Run(args []string, stdin io.Reader, stdout io.Writer, stderr io.Writer) int {
+	return cmd.Do(args, stdin, stdout, stderr)
+}


kyleconroy (changes requested):
Can we create mirror what eabuild does and create a new pkg/cli package?

https://pkg.go.dev/github.com/evanw/esbuild/pkg/cli

oliverpool:
@kyleconroy sure, I amened my commit.

kyleconroy (changes requested):
Can we remove the stdin, stdout, and stderr arguments? I'd like the exposed function to only take a string slice.

oliverpool:
Like this?

kyleconroy (approved):

PR #1251, opened 2021-10-28
golang templates: reduce dbtx interface when not using prepared queries

robertoamici:
Hi there,

i havent seen any PR requirements or existing tests on the repo so this is just the bare bone change.

Basically I am conditionally making the DBTX interface in the golang sql.DB implementation a little smaller by removing the `PrepareContext` when prepared queries are turned off, since all usages of `PrepareContext` are in the part of the template behind the same if statement. In general, if a method is not used, why add it to the interface? 

For some context, we ran across this because we are trying to pass to sqlc a wrapper of sql.DB (from https://pkg.go.dev/github.com/honeycombio/beeline-go/wrappers/hnysql) which has a slightly different PrepareContext return type. We don't use prepared queries so that difference shouldn't impact us.


kyleconroy (changes requested):
The expected test output needs to be updated using `make regenerate`.

PR #1252, opened 2021-10-28
fix(compiler): Set RV for TargetList items on updates

andreasgan:
Hi, thank you for this library!
I stumbled upon a bug with updates not working on `update` queries with multiple tables. It gave an error [`column does not exist`](https://github.com/kyleconroy/sqlc/blob/6ee39cb625260ef2beda50ba9378c96957e35674/internal/compiler/resolve.go#L354) which was because [ref.rv was nil](https://github.com/kyleconroy/sqlc/blob/6ee39cb625260ef2beda50ba9378c96957e35674/internal/compiler/resolve.go#L322).
This PR fixes this by looping through the TargetList and setting the ref.rv from the `Relation`, similarly to what inserts already do. Here's the relevant AST:

```go
   Relation: (*ast.RangeVar)(0xc000ac2ab0)({
    Catalogname: (*string)(<nil>),
    Schemaname: (*string)(<nil>),
    Relname: (*string)(0xc000aa3b40)((len=5) "media"),
    Inh: (bool) true,
    Relpersistence: (uint8) 112,
    Alias: (*ast.Alias)(0xc000aa3b60)({
     Aliasname: (*string)(0xc000aa3b50)((len=2) "m1"),
     Colnames: (*ast.List)(0xc000aa75f0)({
      Items: ([]ast.Node) <nil>
     })
    }),
    Location: (int) 1469
   }),
   TargetList: (*ast.List)(0xc000aa7608)({
    Items: ([]ast.Node) (len=1 cap=1) {
     (*ast.ResTarget)(0xc000ac2ae0)({
      Name: (*string)(0xc000aa3b70)((len=9) "agerating"),
      Indirection: (*ast.List)(0xc000aa7620)({
       Items: ([]ast.Node) <nil>
      }),
      Val: (*ast.ParamRef)(0xc000527f98)({
       Number: (int) 5,
       Location: (int) 1696,
       Dollar: (bool) true
      }),
      Location: (int) 1684
     })
    }
   }),
```


kyleconroy (changes requested):
Can you please add at least one test case? There are many examples in the `internal/endtoend/testdata` directory.

kyleconroy:
@andreasgan Sorry for the second message! I realized that I hadn't posted the comment as a review. No rush, seriously.

andreasgan:
> @andreasgan Sorry for the second message! I realized that I hadn't posted the comment as a review. No rush, seriously.

No worries. Tests added now.

kyleconroy (approved):

Issue #1077, opened 2021-07-04
"BETWEEN ? AND ?" not working in mysql


wirekang:
[Playground](https://play.sqlc.dev/p/191c72c32e4c4c35ec8c7f197c20ee52e38c54c978484bc7ae9dd7f296b9b0b2)

PR #1265, opened 2021-11-02
feat: Add MySQL support for BETWEEN arguments

timstudd:


timstudd:
Fixes https://github.com/kyleconroy/sqlc/issues/1077

kyleconroy (approved):
Thank you so much for the great PR

Cure4Life:
Hi! Unless I'm missing something here, the BETWEEN argument is not working. The playground link for #1077 shows no parameters able to be passed for the 2 BETWEEN arguments in the query methods. Is this supposed to be working? I too cannot get BETWEEN to work correctly, as I get the same results... a query method that ignores the arguments for BETWEEN.

Playground: https://play.sqlc.dev/p/191c72c32e4c4c35ec8c7f197c20ee52e38c54c978484bc7ae9dd7f296b9b0b2

Thank you

kyleconroy:
The playground is still using version 1.10. I'm hoping to release 1.11 today

Cure4Life:
Ahh that explains it. My bad.... for whatever reason I had it in my head it was supposed to have been in the latest release.... thanks!  Easy to work-around for now, at least for my uses, but there are times when BETWEEN just makes large queries more readable.

Issue #1247, opened 2021-10-22
MySQL: LEFT OUTER JOINed columns not considered NULLable


Funzinator:
### Version

1.10.0

### What happened?

When I join another table using `LEFT OUTER JOIN`, the referenced colums are not treated as potential `NULL`.

For example, `int32` is used instead of `sql.NullInt32`

### Database schema

```sql
CREATE TABLE authors
(
    id INT(10) NOT NULL,
    name VARCHAR(255) NOT NULL,
    parent_id INT(10),
    PRIMARY KEY (id)
);

CREATE TABLE parents
(
    id INT(10) NOT NULL,
    name VARCHAR(255) NOT NULL,
    PRIMARY KEY (id)
);
```

### SQL queries

```sql
-- name: AllAuthors :many
SELECT authors.id,
       authors.name,
       parents.id as alias_id,
       parents.name as alias_name
FROM authors LEFT OUTER JOIN parents ON (authors.parent_id = parents.id);
```

### Configuration

```yaml
{
  "version": "1",
  "packages": [
    {
      "path": "db",
      "engine": "mysql",
      "schema": "query.sql",
      "queries": "query.sql"
    }
  ]
}
```

### Playground URL

https://play.sqlc.dev/p/ce7fec524f816ea7f485c8d9b83c60a0e36c7de065db7d23150d7271ce59bd90

### What operating system are you using?

macOS

### What database engines are you using?

MySQL

### What type of code are you generating?

Go

Funzinator:
@timstudd, thank you. This fix already enabled me to get rid of a bunch of `COALESCE` statements.
Unfortunately, it still doesn't work with `LEFT OUTER JOIN`s with the same table. It might be related to #1237.

I've prepared an [example](https://play.sqlc.dev/p/cffeef39709f61e77413bc75ae15cfed67965d99a22b1470365cb4fb62bd29fa).

Should I open another ticket for that or is it OK to track it here?

Funzinator:
Thank you so much!

PR #1270, opened 2021-11-04
fix(compiler): Support nullable fields in joins on same table

timstudd:


timstudd:
Fixes second issue reported in https://github.com/kyleconroy/sqlc/issues/1247

Issue #383, opened 2020-03-06
WithTx is not part of the method set of interface generated


dharmjit:
Hi, I have interface embedding where my db interface is embedding sqlc generated interface. As db interface is then injected to service layer, I am not able to use this. Is it possible to include WithTX merhod in emitted interface also.

kyleconroy:
> Is it possible to include WithTX method in emitted interface also.

I actually don't think it's possible to include the `WithTx` method. The reason is that `WithTx` currently returns a concrete type, `*Queries`. I think we'd need to change the signature of `WithTx` to return a `Querier` and then always emit the interface. I'm not sure it's worth a breaking change.

There may be a better way to do this. I'll need to think about it a bit more.

https://github.com/golang/go/issues/7512

dharmjit:
I guess most project which will adoptsqlc, will try to keep flexibility for some custom methods  which will not come under code genration. Then interface embedding will be required.

dharmjit:
Hi @kyleconroy , what do you suggest meanwhile to handle transactions.

carlmjohnson:
I have a similar problem. I wanted to add a BeginTx call to my app, but I was using the interface everywhere, so I would have to tear all that out and replace it with a concrete type to do that.

I think it would make sense to add `BeginTx(ctx context.Context, opts *db.TxOptions) (Querier, error)` to the interface to simplify usage. If you're worried about backwards compatibility, it could be a config option that defaults to false.

ScottyFillups:
Can't you just create a wrapper interface and struct?

```go
package wrapper

type Querier interface {
	db.Querier
	WithTx(tx *sqlx.Tx) Querier
}

type Queries struct {
	*db.Queries
}

func (q *Queries) WithTx(tx *sqlx.Tx) Querier {
	q.Queries.WithTx(tx.Tx)
	return q
}

func NewQuerier(q *db.Queries) Querier {
	return &Queries{
		Queries: q,
	}
}
```

I haven't tested this out yet; I'll edit this post if there's any problems with this approach.

abagshaw:
> I have a similar problem. I wanted to add a BeginTx call to my app, but I was using the interface everywhere, so I would have to tear all that out and replace it with a concrete type to do that.
> 
> I think it would make sense to add `BeginTx(ctx context.Context, opts *db.TxOptions) (Querier, error)` to the interface to simplify usage. If you're worried about backwards compatibility, it could be a config option that defaults to false.

[...]

PR #1279, opened 2021-11-10
Add DBTX Param config option

danielmmetz:
### Problem
I'd like to be able to define a type that wraps a small interface along the lines of
```go
type WidgetInserter interface {
    InsertWidget(ctx context.Context, db DBTX, name string) error
}
```

Using `sqlc` to generate the underlying code today would give a `*Queries` that has these two relevant methods:
```go
InsertWidget(ctx context.Context, name string) error
WithTx(tx *sql.Tx) *Queries
```

Because `*Queries` operates against an internal `DBTX`, to use `InsertWidget` against a pre-existing transaction then requires creating a new `*Queries` by use of its `WithTx` method. But because `WithTx` returns `*Queries`, it's non-trivial to keep this as just an implementation detail—to have an interface that doesn't expose the underlying `*Queries` requires either: 
* manually adding variant implementations of `WithTx` that return the desired interface (which also requires now that the interface be accessible to the generated package, instead of being defined potentially only at the consumer site), or
* requiring that the interfaces being used compose the entire generated query set (i.e. the generated `Querier` instead of just one of its methods)

Some of the why for both of these is covered in https://github.com/kyleconroy/sqlc/issues/383 and https://github.com/golang/go/issues/7512.

### Proposal
Support a new configuration parameter `dbtx_param` that modifies the generated method sets to provide methods like shown in `WidgetInserter`. If set to `true`, the generated `*Queries` omits storing a DBTX as a struct field and requires it be passed in to all method calls. In doing so, it allows callers to easily provide the connection for standalone use or for use as part of a broader transaction and makes it easy for the surrounding code to use a narrowly defined interface.

### Notes
This PR's current implementation treats `emit_prepared_queries` and `dbtx_param` as mutually exclusive config options. The thought being that since prepared queries are prepared against a particular database connection, and because executing the queries would now use a caller-provided connection, these two connections may end up distinct and void the value of preparing the queries. The implementation can be changed to support two the config options together, it'd just be more involved to do so.


kyleconroy (changes requested):
Instead of adding an test case in the examples, can you create one in the `internal/endtoend/testdata` directory?

I also think we can improve the configuration option name. Instead of `dbtx_param`, I'd pick `emit_methods_with_db_argument`. More wordy, but more clear. Thoughts?

danielmmetz:
> Instead of adding an test case in the examples, can you create one in the `internal/endtoend/testdata` directory?
> 
> I also think we can improve the configuration option name. Instead of `dbtx_param`, I'd pick `emit_methods_with_db_argument`. More wordy, but more clear. Thoughts?

Appreciate the feedback, thanks. I agree that name is more clear (and more consistent). I'll push up another commit accordingly.

kyleconroy (approved):

PR #1280, opened 2021-11-10
docs: add go.17+ install instructions

lopezator:
Add go 1.17+ install intructions, `go get` has been deprecated
from go 1.17 and onwards as an install method.


lopezator:
See: https://golang.org/doc/go-get-install-deprecation for more info.

Issue #437, opened 2020-04-06
Aliases are not validated


dharmjit:
If these is mismatch in aliases as used below, no errors are thrown by `sqlc generate`:

```sql
-- Example queries for sqlc
CREATE TABLE authors (
  id   BIGSERIAL PRIMARY KEY,
  name text      NOT NULL,
  bio  text
);

-- name: GetAuthor :one
SELECT * FROM authors a
WHERE p.id = $1 LIMIT 1;
``` 

[Playground](https://play.sqlc.dev/p/c6f562b6de51d792b868b7cdc53c637883535552103f001ca6254000f146ca47)

kyleconroy:
Thanks for the report. As you can see, there's still a bunch of work to be done. I appreciate all the issues you've opened.

PR #1283, opened 2021-11-13
fix(compiler): Validate table alias references

timstudd:


timstudd:
Fixes https://github.com/kyleconroy/sqlc/issues/437

Issue #1284, opened 2021-11-13
GROUP BY clause references not validated


timstudd:
### Version

1.10.0

### What happened?

An error should be returned when a GROUP BY clause references a column that does not exist

### Database schema

```sql
CREATE TABLE authors (
  id   BIGSERIAL PRIMARY KEY,
  name text      NOT NULL,
  bio  text
);
```

### SQL queries

```sql
-- name: ListAuthors :many
SELECT   *
FROM     authors
GROUP BY invalid_reference;
```

### Configuration

```json
{
	"version": "1",
	"packages": [
		{
			"path": "go",
			"engine": "postgresql",
			"name": "querytest",
			"schema": "query.sql",
			"queries": "query.sql"
		}
	]
}
```

### Playground URL

https://play.sqlc.dev/p/6e1c775ff5f5b3709903bc77dac8aa51f321a648fd00f4990e1c3eceead271ae

### What operating system are you using?

Linux

### What database engines are you using?

PostgreSQL

### What type of code are you generating?

Go

PR #1285, opened 2021-11-13
fix(compiler): Add validation for GROUP BY clause column references

timstudd:


timstudd:
Fixes https://github.com/kyleconroy/sqlc/issues/1284

kyleconroy (changes requested):
Can you also add a test for MySQL?

kyleconroy (approved):

PR #1300, opened 2021-11-21
Add fallback for unhandled parameter references

wabain:
Currently there are some constructs which will make the loop in internal/compiler.resolveCatalogRefs silently skip parameters which were identified in the find_params pass. The only concrete example I know how to trigger is passing a parameter inside a case expression as an argument to a function call, although it looks like there are some other spots in the code which could have similar results. The outcome is that currently `sqlc generate` completes without errors, but the generated queries will probably fail at runtime due to missing arguments.

This PR adds fallbacks to use a parameter with a default name and type `any` in those cases. I added a check at the end of the loop to panic if the length of the parameter list doesn't match the length of the input paramRef list.


kyleconroy:
> Currently there are some constructs which will make the loop in internal/compiler.resolveCatalogRefs silently skip parameters which were identified in the find_params pass.

Instead of adding a fallback, can we identify the cases this happens and fix those instead? I'm a bit skeptical of adding the fallback, mainly because defaulting to the "any" type isn't very useful.

wabain:
Yeah, adding a fallback is definitely just a quick fix. For the one concrete instance I know of (case statements as function call arguments) it should definitely be possible to do something. I can update this pull request to do something nicer, but it may be a while before I can get to it.

One complication is that it would be good to still use the function argument name for the parameter name where it makes sense, if there's nothing better, and currently that would keeping the function call node as the parameter parent and handling the child expression inline. The comment [here](https://github.com/kyleconroy/sqlc/blob/b68c669cd7ccc2679716a116cb7b6af228cac6be/internal/compiler/output_columns.go#L123-L124) about handling recursive types seems like it could apply. I'm guessing the same is true for the `in` and `between` expressions. I don't know quite what MySQL allows there but there are probably a number of different productions.

kyleconroy:
@wabain I'm sorry that we never go to a resolution here. It feels bad closing a PR after it's been open for more than six months, but I think it's the best option right now. If you'd still like to see this get fixed, let's start with an issue first.

PR #1301, opened 2021-11-21
Consistent whitespace handling in Go query template

wabain:
This PR reverts an extra line of whitespace in generated output which snuck in with #1279. This only applied to :execrows using prepared queries, where in one conditional branch there were two spots where whitespace wasn't removed, versus a single spot for the other branches.

```golang
…
{{end -}} // leading whitespace here
    {{- if $.EmitPreparedQueries}} // trailing whitespace here
    result, err := q.exec(ctx, q.{{.FieldName}}, {{.ConstantName}}, {{.Arg.Params}})
    {{- else if $.EmitMethodsWithDBArgument -}} // no trailing whitespace here
…
```

Rather than try to add tests for all of the possible configurations in this pull request, I've made sure all the branches in if/else constructs in the template use consistent whitespace escapes so that if there's extra whitespace it will show up in tests regardless of what branch is taken.


kyleconroy (changes requested):
Can you add a test for the execrows and prepared queries case?

wabain:
Added a test which generates the different kinds of queries using `emit_prepared_queries`.

kyleconroy (approved):

Issue #695, opened 2020-08-30
Support passing slices to MySQL queries


kyleconroy:
While MySQL supports the ANY operator, it does not support arrays. This means the [currently documented way](https://github.com/kyleconroy/sqlc/blob/master/docs/any.md) to pass slices to a query will not work.

kyleconroy:
Here's some related discussion from the go-sql-driver repository:
* https://github.com/go-sql-driver/mysql/issues/107
* https://github.com/go-sql-driver/mysql/issues/176

kyleconroy:
The [MEMBER OF](https://dev.mysql.com/doc/refman/8.0/en/json-search-functions.html#operator_member-of) syntax may offer a solution. The problem is that you need to encode the slice as JSON before passing it into the query.

```sql
CREATE TABLE pkg (
  id text NOT NULL,
  vic VARCHAR(50) NOT NULL,
  kim text NOT NULL,
  order INTEGER NOT NULL,
  PRIMARY KEY (id)
);

/* name: GetPkgID :many */
SELECT id, vic, kim FROM pkg
WHERE vic MEMBER OF (?) AND shop_id = ?;
```

kyleconroy:
Okay, here's my current idea. We add a second type parameter to `sqlc.arg`. That parameter, if passed, will compile to a cast statement. We can support the ARRAY type syntax in this type parameter and generate the correct code.

```sql
CREATE TABLE users (id SERIAL);

-- name: FilterUsers :many
SELECT * FROM users
WHERE id MEMBER OF(sqlc.arg('ids', 'SERIAL ARRAY'));
```

```go
const filterUsers = `-- name: FilterUsers :many
SELECT id FROM users
WHERE id MEMBER OF(CAST(? AS JSON));
`

func (q *Queries) FilterUsers(ctx context.Context, ids []int64) ([]User, error) {
	// ...
}
```

A few issues I can think of:
* The array type must convert to a scalar type that can be represented inside a JSON array
* We'd need to generate a helper method that converts slices to JSON
* We'd probably also need to add a method that unmarshals JSON as well

warent:
I'm not sure if it's bad practice, but maybe the generated code could produce a dynamic query, generating N placeholders for each array value. So, something like:

```go
func filterUsers(vals int[])
{
return fmt.Sprintf(`-- name: FilterUsers :many
SELECT id FROM users
WHERE id MEMBER OF(?%v);
[...]

PR #1312, opened 2021-11-27
Add sqlc.slice() new function type (#695)

cameronpm:
This feature (currently MySQL-specific) allows passing in a slice to an
IN clause. Adding the new function sqlc.slice() as opposed to overloading
the parsing of "IN (?)" was chosen to guarantee backwards compatibility.

   SELECT * FROM tab WHERE col IN (sqlc.slice("go_param_name"))

The MySQL FLOAT datatype mapping has been added too.


cameronpm:
Conflicts with future slated changes (ast based codegen - #1333), withdrawing

PR #1326, opened 2021-12-04
Map postgres interval to go string rather than int64

andrewbeckman:
Currently, Postgres's interval type is mapped to Go's int64 type.

I got around this by using an override, but I figured I'd open this PR as I don't believe it's ever correct to map interval to int64. Namely, I added this block to my `sqlc.yaml` and it fixed my issues:
```
overrides:
  - db_type: "pg_catalog.interval"
    go_type: "string"
  - db_type: "pg_catalog.interval"
    go_type: "database/sql.NullString"
    nullable: true
```

To prevent needing to use an override, I've updated the default type mapping in this PR. Please let me know if I've missed something and this isn't correct.

### DIFF
--- a/internal/codegen/golang/postgresql_type.go
+++ b/internal/codegen/golang/postgresql_type.go
@@ -180,9 +180,9 @@ func postgresType(r *compiler.Result, col *compiler.Column, settings config.Comb
 
 	case "interval", "pg_catalog.interval":
 		if notNull {
-			return "int64"
+			return "string"
 		}
-		return "sql.NullInt64"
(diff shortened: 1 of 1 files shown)


andrewbeckman:
Taking a deeper look at the codebase, I see there are some tests that would need to be updated with this PR, but before investing the time, I figured it'd be good to check if I'm just missing something obvious and in fact interval should be mapped to int64.

andrewbeckman:
After doing some more research, I think I'm wrong and going to close this PR.

PR #1329, opened 2021-12-04
fix(compiler): Check delete statements for CTEs

wabain:
At the point where the compiler processes CTEs it was missing a case for delete statements. This was causing sqlc generate for delete statements using CTEs to fail with the error `relation "<cte_name>" does not exist`.


kyleconroy:
Thanks!

PR #1330, opened 2021-12-06
Proposal: Generate go file with table and columns name to help during dynamic queries

5anthosh:
It will be helpful to have table and column names during dynamic queries

Example
generated [meta.go](https://github.com/5anthosh/sqlc/blob/main/internal/endtoend/testdata/exec_imports/pgx/go/meta.go) from [query.sql](https://github.com/5anthosh/sqlc/blob/main/internal/endtoend/testdata/exec_imports/pgx/query.sql)


kyleconroy:
Thank you so much for taking the time to put together this PR. I'm not going to be merging this right now. Dynamic queries are something that I want to support, but I'd like for sqlc to offer a comprehensive solution to the problem. This PR only gets us a bit closer to a solution.

I am working on a plugin system so that you could augment the output of sqlc with this metadata without having to make changes to sqlc itself. It's a few months out, but should provide the necessary functionality.

Issue #403, opened 2020-03-12
Postgres: Support inherited table schema


ivanvanderbyl:
When using table inheritance in Postgres, `sqlc` can't reference any columns in the parent table.

**Example setup (from pg_dump):**

```sql
-- schema.sql
CREATE TABLE public.reservations (
    id character varying(255) NOT NULL,
);

CREATE TABLE public.flight_reservations (
    flight_id character varying(255)
)
INHERITS (public.reservations);
```

```sql
-- queries.sql
-- name: InsertFlightReservation :one
insert into flight_reservations (id, flight_id)
VALUES ($1,
        $2) RETURNING (id);
```

**Result:**

```bash
sqlc generate
# package data
queries.sql:111:38: column "id" does not exist
```

**Expected:**

The expected result should be to reference the inherited table columns.

Ironpark:
Models created through inherited tables must either embed the parent table model or have all the columns the parent has.
```sql
CREATE TABLE transfer
(
    id           bigserial primary key        not null,
    txid         text                         not null,
    currency_id  int references currency (id) not null,
    from_address text                         not null,
    to_address   text                         not null,
    volume       decimal                      not null
);

CREATE TABLE transfer_eth
(
    txid text references raw_transaction_eth (txid) not null,
    UNIQUE (currency_id, txid, to_address)
) INHERITS (transfer);
```
expected result:
``` go
type Transfer struct {
	ID          int64
	Txid        string
	CurrencyID  int32
	FromAddress string
	ToAddress   string
	Volume      string
}

type TransferEth struct {
	Transfer
}
// OR
type TransferEth struct {
	ID          int64
	Txid        string
	CurrencyID  int32
	FromAddress string
	ToAddress   string
	Volume      string
}
```
actually generated result:
```go

type Transfer struct {
	ID          int64
	Txid        string
	CurrencyID  int32
	FromAddress string
	ToAddress   string
	Volume      string
}

type TransferEth struct {
	Txid string
}
```

carpenter-ryan:
Is someone willing to look into this issue? We could really use a solution.

PR #1339, opened 2021-12-13
Postgres Inherits

timwmillard:
Support for Postgres Inheritance.
https://www.postgresql.org/docs/current/ddl-inherit.html

Fixes https://github.com/kyleconroy/sqlc/issues/403

Add the columns from the parent tables to the child table.

This change also adds support for columns from partition tables.


kyleconroy:
@timwmillard How is inheritance supposed to work with DDL statements that alter tables?

timwmillard:
> @timwmillard How does inheritance with DDL statements that alter tables? The PostgreSQL docs seem to imply that all child tables should also be updated:
> 
> > Commands that are used for data querying, data modification, or schema modification (e.g., SELECT, UPDATE, DELETE, most variants of ALTER TABLE, but not INSERT or ALTER TABLE ... RENAME) typically default to including child tables and support the ONLY notation to exclude them

@kyleconroy, I did look into this.  When altering a table to add inheritance the child table need to already have all the parent column on the table.

Example:
```
CREATE TABLE parent (
   name text
);

CREATE TABLE child (
   name text
   age text
);

ALTER TABLE child INHERIT parent;
```
In this example the child table already has the `name` column.  If it didn't the `ALTER` statement would fail.  Therefore the generated queries already have the correct columns.

timwmillard:
Same as if you `ALTER TABLE child NO INHERIT` the column will remain the same, but just don't have the internal inherit link.

jlisthood:
> @timwmillard How does inheritance with DDL statements that alter tables? The PostgreSQL docs seem to imply that all child tables should also be updated:
> 
> > Commands that are used for data querying, data modification, or schema modification (e.g., SELECT, UPDATE, DELETE, most variants of ALTER TABLE, but not INSERT or ALTER TABLE ... RENAME) typically default to including child tables and support the ONLY notation to exclude them

So are ALTER statements just not supported for inheritance then? I can alter base tables without seeing derived tables getting the new fields: https://play.sqlc.dev/p/0a50edacd24090cbb53d83ec6fe69b11ba109a0ab6a80ae2e33aec35a6b6bef3

PR #1361, opened 2021-12-27
feat(kotlin): Generate code using ASTs

kyleconroy:


kyleconroy:
While I'm excited about this approach, I'm going to focus my time more on defining the interface between sqlc and the codegen packages instead of the internal codegen implementations.

Issue #1373, opened 2022-01-11
Generated Types for single param queries


Threpio:
### Version

1.11.0

### What happened?

Upon SQLC generate with the following schema and queries I have a type generated for the params for the GetUserByEmailOrUsername query but not the GetUserByEmail.

I understand that this might be by design and that for a lot of people they would end up with redundant types but I would much prefer that it generated the types (even with just a single field) to allow for more con-formative and similar code. 

I would like for a query with a single param to generate a struct like this:
```go
type GetUserByEmailParams struct {
     Email string
}
```

And for that to be used within the generated function:
```go
func (q *Queries) GetUserByEmail(ctx context.Context, arg GetUserByEmailParams) (User, error) {}
```

I know that this might not be what everyone prefers in terms of redundant types being generated for others. Perhaps a flag or a configuration field could be added to turn this on or off. Happy to work on it if it is something you would be willing to accept. 

### Database schema

```sql
CREATE TABLE users (
  id SERIAL PRIMARY KEY,
  username VARCHAR(255) NOT NULL,
  email VARCHAR(255) NOT NULL
);
```

### SQL queries

```sql
-- name: GetUserByEmail :one
SELECT * FROM users WHERE email = $1;

-- name: GetUserByEmailOrUsername :one
SELECT * FROM users WHERE email = $1 OR username = $2;
```

### Configuration

```yaml
version: 1
packages:
  - path: "queries"
    name: "queries"
    engine: "postgresql"
    schema: "schema.sql"
    queries: "query.sql"
```

### What operating system are you using?

macOS

### What database engines are you using?

PostgreSQL

### What type of code are you generating?

Go

Threpio:
I apologise if this has been bought up before in a discussion or in a bug report. I had a quick look but could not find anything. If I have missed it and you can link me to that discussion then please feel free to close this!

kyleconroy:
Closing in favor of https://github.com/kyleconroy/sqlc/issues/1498

PR #1375, opened 2022-01-13
Feature/generate structs with single param

Threpio:
To deal with: https://github.com/kyleconroy/sqlc/issues/1373

I have not covered all of the Docs as of yet.

I need confirmation before this will merge that the default for a boolean config is false so that the logic in commit #4fdd8c3 is correct


Jille:
Bools in Go default to false indeed.

kyleconroy:
Closing in favor of https://github.com/kyleconroy/sqlc/issues/1498. Please comment on the proposed design :)

PR #1376, opened 2022-01-14
feat(postgres): Add support for sql.NullInt16

egtann:
Go added support for sql.NullInt16 in the standard library as of go1.17 (https://github.com/golang/go/issues/40082), so sqlc can now use the appropriate nullable type for int16.


kyleconroy (approved):
I agree! Go 1.17 has been out long enough that I'm comfortable updating the target to require it. Once Go 1.18 lands, we'll need to figure out a strategy to support targeting different Go versions.

PR #1395, opened 2022-01-26
fix(postgresql): Remove extra comment newline

### LINKED ISSUE #1393: SQL comments have one too many newlines for GoDocs
### Version

1.11.0

### What happened?

For example, if I have the table with the comment
```sql
-- name: IncrementPlayerTokens :exec
-- This function will increment the tokens and the rollover (will set roll0ver to 0 if it would be negative)
UPDATE players
SET tokens = tokens + $1,
rollover = (
  CASE
    WHEN rollover + $1 >= 0 THEN rollover + $1
    ELSE 0
  END
)
WHERE id = $2;
```

The following go code is generated:

```go
// This function will increment the tokens and the rollover (will set roll0ver to 0 if it would be negative)

func (q *Queries) IncrementPlayerTokens(ctx context.Context, arg IncrementPlayerTokensParams) error {
	_, err := q.db.Exec(ctx, incrementPlayerTokens, arg.Tokens, arg.ID)
	return err
}
```

You can see there is an extra line between the comment and the function, which does not allow the comments to appear as godocs.

It would be sublime if that newline was removed so the comment and function were touching, thus being able to write godocs in the SQL :)

### Relevant log output

_No response_

### Database schema

_No response_

### SQL queries

_No response_

### Configuration

```
version: 1
packages:
  - path: "query"
    name: "query"
    engine: "postgresql"
    schema: "sqlcSchema.sql"
    queries: "./queries/"
    sql_package: "pgx/v4"
```

### Playground URL

https://play.sqlc.dev/p/19da31cef1e9d4ce1385c4d620f2bcfad53be81567e1df81551f59e9fb8b39b8

macOS

PostgreSQL

Go

kyleconroy:
Fixes #1393


kevinburke:
whitespace, it'll cost ya

PR #1410, opened 2022-02-02
Start filling out SQLite support

PadraigK:
This adds a few tests for SQLite and implements the necessary features to have them pass.


internal/engine/sqlite/parser/SQLiteParser.g4
```diff
@@ -76,6 +76,7 @@ alter_table_stmt:
             | COLUMN_? old_column_name = column_name TO_ new_column_name = column_name
         )
         | ADD_ COLUMN_? column_def
+        | DROP_ COLUMN_? column_name
```
PadraigK, line 79: Added this support to the parser required a region, which explains a lot of the size of this PR.

PadraigK:
I would welcome any feedback on Go style etc. I've tried to follow the patterns I see in the existing code, but I'm new to Go and am probably missing some conventions.

internal/endtoend/testdata/count_star/sqlite/go/query.sql.go
```diff
@@ -0,0 +1,32 @@
...
+	var count int64
+	err := row.Scan(&count)
+	return count, err
+}
+
+const countStarUpper = `-- name: CountStarUpper :one
+r;
+
+SELECT COUNT(*) FROM bar
+`
```
kyleconroy, line 25: Not sure what's happening here, but this query is invalid.

kyleconroy (changes requested):
A bit torn about improving SQLite support using the ANTLR parser. I think the best long-term approach is to use [golemon](https://github.com/gopikchr/golemon) and hand-port the SQLite parser.

That said, no one is working on that and this parser currently works, so I'm happy to have you continue to work on it.

internal/endtoend/testdata/count_star/sqlite/go/query.sql.go
PadraigK: It's something to do with the `--` comment parsing. I'll look into it properly tomorrow!

PadraigK:
Thank you for all your work on this project, I'm really enjoying tinkering with it.

> A bit torn about improving SQLite support using the ANTLR parser. I think the best long-term approach is to use [golemon](https://github.com/gopikchr/golemon) and hand-port the SQLite parser.
> 
> That said, no one is working on that and this parser currently works, so I'm happy to have you continue to work on it.

Interesting! I'm not against changing tack on this. I don't have a particular attachment to ANTLR, I continued with it mainly because the pieces were already in place. 

Just to weigh up the pros and cons, ANTLR is a very mature project and the generated parser seems to work fine for everything I've needed so far, so I'm inclined to continue with it. 

I understand that lemon is what SQLite itself uses so maybe there'll be some nuances that match up better by using `golemon`. Plus it's always a little nicer in terms of maintenance to have the full stack in the same language. Are there other other reasons in mind to favour `golemon`? 

Anyway, worst case, this work I'm doing so far will build up a good library of SQLite test cases and a list of supported sql functions which will be useful with any parser.

kyleconroy:
> I understand that lemon is what SQLite itself uses so maybe there'll be some nuances that match up better by using golemon. Plus it's always a little nicer in terms of maintenance to have the full stack in the same language. Are there other other reasons in mind to favour golemon

Nope, the sole reason is that it would be closer to the real SQLite parser. Our PostgreSQL support is better than our MySQL support because our PostgreSQL parser **is** the PostgreSQL parser, where the MySQL parser is what PingCAP currently supports.

That said, if you think the current parser works, I'm all for it.

internal/endtoend/testdata/count_star/sqlite/go/query.sql.go
PadraigK: This issue is fixed now and the code generates properly.

PR #1426, opened 2022-02-10
fix(config): Fix type name for versioned type overrides

egtann:
This fixes a bug for go_type overrides which have versions. Previously for something like this:

```
go_type: a/b/v2.MyType
```

sqlc would generate:

```
import "a/b/v2"

type MyStruct struct {
    A v2.MyType
}
```

Which is invalid and doesn't compile. This leaves the import statement alone but changes the outputted type name to `b.MyType`.

### DIFF
--- a/cmd/sqlc/main.go
+++ b/cmd/sqlc/main.go
@@ -3,7 +3,7 @@ package main
 import (
 	"os"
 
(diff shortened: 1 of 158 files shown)


egtann:
Closing in favor of a dedicated branch. Need this functionality live to support an application, so I'm using a fork until if/when this is upstreamed.

PR #1428, opened 2022-02-10
fix(config): Fix type name for versioned type overrides

egtann:
This fixes a bug for go_type overrides which have versions. Previously for something like this:

```
go_type: a/b/v2.MyType
```

sqlc would generate:

```
import "a/b/v2"

type MyStruct struct {
    A v2.MyType
}
```

Which is invalid and doesn't compile. This leaves the import statement alone but changes the outputted type name to `b.MyType`.

### DIFF
--- a/internal/config/go_type.go
+++ b/internal/config/go_type.go
@@ -149,6 +149,21 @@ func (gt GoType) Parse() (*ParsedGoType, error) {
 			typename = typename[:len(typename)-len("-go")]
 		}
 		o.ImportPath = input[:lastDot]
+
+		// set versioned overrides to reference the underlying package
+		// name. for example, the type for a/b/v2.C is b.C, not v2.C.
+		parts := strings.Split(input, "/")
+		if len(parts) >= 2 {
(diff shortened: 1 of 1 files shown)


kyleconroy:
`go_type` also accepts an object where you can specify the exact import, package, type and pointer. Here's how to use it:

```
{
  "version": "1",
  "packages": [
    {
      "path": "db",
      "engine": "postgresql",
      "schema": "query.sql",
      "queries": "query.sql"
    }
  ],
  overrides: [{
    "column": "authors.id",
	"go_type": {
	  import: "github.com/a/b/v2",
	  package: "b",
	  type: "MyType",
	  pointer: false
	}
  }]
}
```
[playground](https://play.sqlc.dev/p/56d319fcd45113fc76ea179d143d906862c0e37da43add5ad6692798e6b138be)

egtann:
This is great. I had no idea. Thanks!

kyleconroy:
Yep, the docs were lacking.

PR #1435, opened 2022-02-12
Add config option to customize Go interface & struct naming

ShivamSarodia:
### Problem

I'd like to
- define a wrapper interface around the Querier object for my application, and
- enforce that callers outside my database package define types only of my wrapper interface rather than the underlying Querier interface or Queries objects
 
In my case, this is specifically to enforce best practices around mockability and testability of database code. 

### Proposal

Support two new configuration parameters, `interface_name` and `queries_struct_name`, which enable customization of the generated interface and struct names. This helps in two ways:

- The generated interface and struct can be made private, preventing callers outside my package from accessing those types directly.
- The desirable interface name Querier can be freed up for my wrapper interface.

Open to feedback on this, especially around naming of the configuration parameters.


kyleconroy (commented):
@ShivamSarodia I'm really sorry it's taken me a month to review this PR.

For right now, I'm not looking to make the struct and interface names configurable. The reason is that I'm looking to do a larger refactor of the Go codegen with an eye towards making these types of changes possible without changing sqlc itself.

ShivamSarodia:
No worries, thank you for the great tool! Looking forward to the refactor.

PR #1437, opened 2022-02-13
feat(pgx): Add support for batch operations

### LINKED ISSUE #1275: Feature request: Support for pgx batching
As mentioned in #472, `pgx` batching can be a very important tool for efficient inserts and queries. There doesn't seem to be any way to use the generated methods with the existing `pgx` batching interface. The `pgx` batching interface is defined by the [Batch](https://pkg.go.dev/github.com/jackc/pgx/v4#Batch) type and its [Queue](https://pkg.go.dev/github.com/jackc/pgx/v4#Batch.Queue) method. It would be neat if we could have something similar to the generated `New(db DBTX) *Queries` function that operates on a `pgx.Batch`.

We'd need to think a little about the design, since in an ideal world we'd still be able to get typed responses for each entry on the batch. Maybe each function on the type that operates on a batch should take a callback or something for how to process the returned data?

jrperritt:
This change offers one possible implementation of batch operations via the `pgx` driver. A full example exercising all of the new API surface can be found in `examples/batch/postgresql/db_test.go`.

This implementation adds 3 new commands: `batchexec`, `batchmany`, and `batchone`.
Each of them returns a batch object. The batch object has 2 methods:
- a method corresponding to one of the methods on the [`pgx.BatchResults` object](https://pkg.go.dev/github.com/jackc/pgx/v4@v4.14.1#BatchResults) (`Exec`, `Query`, `QueryRow`, respectively).
Each of the methods takes a callback as a parameter as follows:
```go
Exec(func(int, error))
Query(func(int, []T, error))
QueryRow(func(int, T, error))
```
The first parameter in the callback is the index of the result from the batch result. That is, if the batch result is completely consumed, `i` will range from 0 to `len(input_params_slice) - 1`.
- a `Close` method for closing the batch result early and returning the underlying connection.

closes #1275


jrperritt:
I was surprised to see the error in CI, but I see it was [recently added](https://github.com/jackc/pgx/blob/v4.15.0/batch.go#L70) in `pgx`.
I also noticed a place where I didn't update the interface generation code to use `batchmany` and `batchone`, only `batchexec`.
Other than that, only rebased on `main`.

kyleconroy (approved):
If you have time, a few follow-on asks after this is merged.

* Can you update query-annotations.md to make it obvious that this only works for the PostgreSQL / Go / pgx combo?
* Can you rename the `batch import` test folder to `batch_import`? Our naming convention doesn't include spaces.

kyleconroy (changes requested):
Actually, it looks like that space issues with `batch import` is causing the build to fail, so you'll need to fix that before I can merge it in.

jrperritt:
> Can you update query-annotations.md to make it obvious that this only works for the PostgreSQL / Go / pgx combo?

Yes, good idea.

> Can you rename the batch import test folder to batch_import? Our naming convention doesn't include spaces.

Mine neither! Fixed.

kyleconroy (approved):

danielbprice:
This is really nice.  Thanks for doing it.

PR #1443, opened 2022-02-17
SQLite: Adds a few more tests, fixes DROP TABLE  

PadraigK:
Continuing to fill out SQLite end-to-end data.


internal/endtoend/testdata/ddl_create_table/sqlite/schema.sql
```diff
@@ -1 +1,3 @@
 CREATE TABLE venues (name text);
+
+CREATE TABLE IF NOT EXISTS arenas (name text PRIMARY KEY, location text, size int NOT NULL) WITHOUT ROWID;
```
kyleconroy, line 3: Can we create a separate test for create table without rowID? I don't think it applies to any other engine than SQLite

Issue #1450, opened 2022-02-21
Correct aggregate function variant not selected from catalog


PadraigK:
### Version

1.12.0

### What happened?

I have a query which runs an aggregate function on a column (`sum(pages)`). There are multiple variants of functions with this name (`sum`) defined in the various catalogs, e.g. `sum(numeric) → numeric, sum(integer) → integer`. 

My expectation is that the function chosen by the engine would be the one matching the type of the `pages` column parameter. Instead, the return type of the first function in the catalog that matches the name and argument count is used.  

The `ResolveFuncCall` in `public.go` doesn't seem to attempt any type matching. In the play.sqlc.dev example below, the `sum` function used is the one returning `bigint`, which is used because it is the first `sum` function defined. However there is also a variant defined which takes and returns a `numeric` type; this is the one that should be used. 

### Database schema

```sql
CREATE TABLE books (
                       id   integer PRIMARY KEY,
                       title text      NOT NULL,
                       author text     NOT NULL,
                       pages numeric      NOT NULL
);
```

### SQL queries

```sql
-- name: AuthorPages :many
select author, count(title) as num_books, SUM(pages) as total_pages
from books
group by author;
```

### Configuration

```yaml
{
  "version": "1",
  "packages": [
    {
      "path": "go",
      "engine": "postgresql",
      "name": "querytest",
      "schema": "schema.sql",
      "queries": "query.sql"
    }
  ]
}
```

### Playground URL

https://play.sqlc.dev/p/ca51e1357b60b0ab6ed38fbb98f8e26ef30783a65cce012c90035a7abf93088c

### What operating system are you using?

macOS

### What database engines are you using?

PostgreSQL, MySQL

### What type of code are you generating?

Go

PadraigK:
I think this is also the root cause of https://github.com/kyleconroy/sqlc/issues/1418

ebo-bpi:
Do you know is there a workaround for this? I'm also hitting this issue.

PadraigK:
The only work-around I can think of would be to manually reorder the catalog functions so that the function variant you want appears higher up the list than the one you're getting but then it'll only be right for that case and wrong for others.

ebo-bpi:
would be nice if you could give sqlc some sort of type hint to overcome this issue. something you do with the column parameter naming.

gissleh:
> would be nice if you could give sqlc some sort of type hint to overcome this issue. something you do with the column parameter naming.

[...]

PR #1455, opened 2022-02-23
feat(sqlite) Adds support for column aliases and function arguments

PadraigK:
* Support for SQLite column aliases
* Support for SQLite function arguments
* Adds SQLite versions for various tests that passed without changes
* Adds a new func_match_types end to end test for postgresql and mysql to illustrate https://github.com/kyleconroy/sqlc/issues/1450


internal/endtoend/testdata/func_match_types/sqlite/go/query.sql.go
```diff
@@ -0,0 +1,43 @@
...
+const authorPages = `-- name: AuthorPages :many
+select author, count(title) as num_books, sum(pages) as total_pages
+from books
+group by author
+`
+
+type AuthorPagesRow struct {
+	Author     string
+	NumBooks   int64
+	TotalPages float64
```
PadraigK, line 19: This should be an int64 once https://github.com/kyleconroy/sqlc/issues/1450 is fixed

PR #1458, opened 2022-02-28
feat(codegen): Insert comments in interfaces

kyleconroy:
Replaces #1451


kyleconroy:
@sbres Thanks for original PR! I updated the tests so that they're passing and will merge once they're green

PR #1462, opened 2022-03-01
Fix nullable colums for subqueries in left and full joins

richardbizik:
There is an issue when performing join on subquery. Fields that should be nullable are not because node is processed as ordinary select node. This pull request fixes the issue.

### DIFF
--- a/internal/sql/ast/range_subselect.go
+++ b/internal/sql/ast/range_subselect.go
@@ -2,6 +2,7 @@ package ast
 
 type RangeSubselect struct {
 	Lateral  bool
+	JoinType JoinType
 	Subquery Node
 	Alias    *Alias
 }
--- a/internal/sql/ast/range_subselect.go
+++ b/internal/sql/ast/range_subselect.go
@@ -2,7 +2,6 @@ package ast
 
 type RangeSubselect struct {
 	Lateral  bool
-	JoinType JoinType
 	Subquery Node
 	Alias    *Alias
 }
--- a/internal/sql/astutils/walk.go
+++ b/internal/sql/astutils/walk.go
@@ -1320,6 +1320,10 @@ func Walk(f Visitor, node ast.Node) {
 			Walk(f, n.Larg)
 		}
 		if n.Rarg != nil {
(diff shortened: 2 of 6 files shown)


internal/sql/astutils/walk.go
```diff
@@ -1320,6 +1320,10 @@ func Walk(f Visitor, node ast.Node) {
 			Walk(f, n.Larg)
 		}
 		if n.Rarg != nil {
+			switch t := n.Rarg.(type) {
+			case *ast.RangeSubselect:
+				t.JoinType = n.Jointype
+			}
```
kyleconroy, line 1326: The astutils package does not directly mutate the AST, so this can't go here.

internal/sql/ast/range_subselect.go
```diff
@@ -2,6 +2,7 @@ package ast
 
 type RangeSubselect struct {
```
kyleconroy, line 3: This type is based on the type here (https://pkg.go.dev/github.com/pganalyze/pg_query_go/v2#RangeSubselect), which does not include join information.

kyleconroy (changes requested):
Sorry, the current approach taken in this PR isn't going to work. The sample query added to the left join test is helpful to identify the problem.

richardbizik:
I made some changes. Would this approach work better?

sbres:
I've tried to run the sql query but I seem unable to make it work as is.

Could you check it ?
https://www.db-fiddle.com/f/hXNdApf4mV3Qi1jEgu2egZ/3

kyleconroy (changes requested):
Is this trying to fix the same problem as https://github.com/kyleconroy/sqlc/pull/1486? If not, I'd like to merge that first before we continue working on this change.

You have some merge conflicts here. Can you rebase?

sbres:
I've closed #1486 as #1491 applies the same fix.
But they do not fix what this PR is trying to achieve

richardbizik:
As @sbres pointed out they are fixing a different issue. This PR is trying to solve an issue with nested select statements where an information is lost about the parent being join that should produce nullable results.

Issue #288, opened 2020-01-22
Extensibility / Plugin Support


mightyguava:
Add support for calling into plugins during code generation. One specific use case I have is to support automatic encryption/decryption of columns. To illustrate, here's very naive proposal:

```sql
CREATE TABLE authors (
          id   BIGSERIAL PRIMARY KEY,
          name text      NOT NULL, -- :encrypted
          bio  text -- :encrypted
);
```

Adding the `:encrypted` annotation and having the `encryption` plugin enabled would cause sqlc to augment the generated code to look like the below:

```go
// db.go
interface Cipher {
  Encrypt(s string) (string, error)
  Decrypt(s string) (string, error)
}

func New(db DBTX) *Queries {
	return &Queries{db: db, enc: &NoopCipher{}}
}

func NewWithEncryptor(db DBTX, ciph Cipher) *Queries {
	return &Queries{db: db, ciph: ciph}
}

type Queries struct {
	db DBTX
	ciph Cipher
}

// query.sql.go
const createAuthor = `-- name: CreateAuthor :one
INSERT INTO authors (
          name, bio
) VALUES (
  $1, $2
)
RETURNING id, name, bio
`

type CreateAuthorParams struct {
	Name string
	Bio  sql.NullString
}

func (q *Queries) CreateAuthor(ctx context.Context, arg CreateAuthorParams) (Author, error) {
	var err error
	var i Author

	arg.Name, err = q.ciph.Encrypt(arg.Bio)
	if err != nil {
		return i, err
	}
	arg.Bio, err = q.ciph.Encrypt(arg.Bio)
	if err != nil {
		return i, err
	}
	row := q.db.QueryRowContext(ctx, createAuthor, arg.Name, arg.Bio)
	err := row.Scan(&i.ID, &i.Name, &i.Bio)
	if err != nil {
		return i, err
	}
	i.Name, err = q.ciph.Decrypt(i.Name)
	if err != nil {
		return i, err
	}
	i.Bio, err = q.ciph.Decrypt(i.Bio)
	if err != nil {
		return i, err
	}

	return i, err
}
```

The `Cipher` is be an interface dictated by the plugin that the user would have to provide.

[...]

Issue #192, opened 2019-12-17
CRUD methods


rubenv:
I'm currently a `gorp` user, mainly to map queries to structs. Replacing that with `sqlc` would be a nobrainer.

However, for some use cases you don't need specific queries, just standard CRUD operations are fine.

Am I understanding it correctly that I'd need to write create/select/update/delete queries for all types manually? Or is there a way to generate these convenience methods for a table?

kyleconroy:
> Am I understanding it correctly that I'd need to write create/select/update/delete queries for all types manually?

Yes, that it correct. While it may change in the future, sqlc does not generate SQL queries for you. It compiles existing SQL queries into type-safe Go.

> Or is there a way to generate these convenience methods for a table?
Convenience methods are not currently generated for tables

rubenv:
Cool, thanks for the clarification.

Brilliant project by the way, really interesting approach. Keep it up :+1:

kyleconroy:
Thanks! I'll ping this issue once progress is made on the CRUD front.

fr3fou:
Any news?

neverlink:
I've been dealing with the same thing, it's been a year, has there been any progress since?

PR #1463, opened 2022-03-02
feat(sdk): Add the plugin SDK package

kyleconroy:
Collect a few commonly used functions and put them into one package.
Move the utils.go file from the codegen package here as well.

cc @stephen


stephen:
Nice!

I am curious what your plan for plugins overall looks like. Is this code going to get pulled outside of `internal/` later? Similarly, what other internals might get pulled out? It would be nice to let other plugins reuse parts of the go (and kotlin/python/etc) code generators.

Some other loose thoughts:

- My endgoal is thinking about how something like #192 could be implemented via plugin. A tricky thing with that case is if the plugin wants to know what the primary key is (currently not exposed by the compiler), or if it wants to annotate individual columns (like in the `:encryption` case mentioned in #288) which might be hard to support since those inline comments are not passed through in the compiler.
- From #1414, are you thinking plugins would all be run via wasm?

kyleconroy:
I have a whole plan but haven’t written it up in a design doc. Will do tonight and share it here.

kyleconroy:
braindump.doc https://docs.google.com/document/d/1DVXSINp-ZAJj9YmzKh0kTE6Cf6lFadVtH1zb9D3Dhjg/edit?usp=sharing

Sec32fun32 (approved):

kyleconroy:
It's not well documented, but the SDK package is here: https://github.com/sqlc-dev/plugin-sdk-go

PR #1460, opened 2022-03-01
Refactor go codegen to plugin types

stephen:
See #288, #923, #1416 .

(hi @kyleconroy!)

I got excited about the plugin api and wanted to try a few things out with the golang code generation...

There are a couple questionable things in here that I wasn't sure about - will add some comments about them. If you think this work is worth continuing with, I'd be happy to try to clean it up and put it in an upstreamable state.

PR #1468, opened 2022-03-02
Remove data_type from column proto 

stephen:
The information is duplicated from `Column.type`

See https://github.com/kyleconroy/sqlc/pull/1460#discussion_r817029197

DataType is never filled in so this comparison is buggy. This change mirrors
what the go gen.go does instead.

Deleting a field from proto can be not safe for compat, so we could use
https://developers.google.com/protocol-buffers/docs/proto3#reserved,
but I think these protos are transient so it should be ok.


kyleconroy (approved):
The proto hasn't been published, so we can make as many changes as we want for now.

PR #1470, opened 2022-03-03
feat: Codegen plugins, powered by WASM

kyleconroy:
Replaces #1414


internal/plugin/codegen.pb.go
```diff
@@ -1807,53 +1875,56 @@ var file_plugin_codegen_proto_goTypes = []interface{}{
...
+	25, // [25:25] is the sub-list for method output_type
+	25, // [25:25] is the sub-list for method input_type
+	25, // [25:25] is the sub-list for extension type_name
+	25, // [25:25] is the sub-list for extension extendee
+	0,  // [0:25] is the sub-list for field type_name
 }
 
 func init() { file_plugin_codegen_proto_init() }
 func file_plugin_codegen_proto_init() {
+	return //XXX
```
kyleconroy, line 1927: Tracking here: https://github.com/tinygo-org/tinygo/issues/2667

internal/python/ast/ast.pb.go
```diff
@@ -2606,6 +2606,7 @@ var file_python_ast_proto_depIdxs = []int32{
 
 func init() { file_python_ast_proto_init() }
 func file_python_ast_proto_init() {
+	return
```
kyleconroy, line 2609: Tracking here: https://github.com/tinygo-org/tinygo/issues/2667

protos/plugin/codegen.proto
```diff
@@ -99,6 +100,12 @@ message GoCode
   string output_files_suffix = 18;
 }
 
+message WASM
+{
+  string path = 1;
+  string out = 2;
+}
```
kyleconroy, line 107: This is a temporary addition to make it easy to test out plugin support. I haven't figured out what the plugin config should look like yet.

skabbes:
I know you're probably not wanting / expecting feedback on this. But in the spirit of the internet, here's unsolicited feedback haha...

I think wasm is really nice to support out of the box, but producing json output over stdout is probably a bit more approachable for most code-gen plugin authors (and it opens it up to languages not supported by wasm easily).  Though supporting wasm plugins would also be trivial to support with a generic wasm runner.

```bash
# illustrative
# sqlc generate --protobuf > sqlcoutput.pb 
sqlc generate -f-file sqlc.json --json > sqlcoutput.json 
cat sqlcoutput.json | gen-go-code
cat sqlcoutput.json | gen-ts-code
cat sqlcoutput.json | wasm-runner ./python-gen.wasm
```

That said if your goal is here to to scratch an engineering itch you have - don't let me stop you!

dstrelau:
@kyleconroy would you be open to a PR that adds a JSON output in the manner @skabbes suggests? It looks like outputting the `plugin.CodeGenRequest` as JSON would be pretty trivial and provide all the relevant info you'd need to write a standalone codegen executable.

skabbes:
@kyleconroy can you clarify (just for my curiosity) what TinyGo has to do with wasm plugins? Is it because you are trying to _write_ the Go and Python code generators in wasm (in Go, compiled to wasm)? And only TinyGo supports wasm??

Or are you envisioning a world where _all_ of sqlc can run in wasm (maybe for security or something like that?).

skabbes:
FYI I did a little digging in the tinygo repo, it appears like you can fix your `defer atomic.StoreUint32` by specifically building a little shim for the symbol.  I did a little proof of concept, and it seems like you can do this completely outside of tinygo (until they can fix the bug the right way), by adding this little bit of code anywhere in your project.  Although that said, I think the fix is going to be relatively easy for someone familiar with the tinygo codebase - there are some other examples of doing similar things - and maybe they just need a little pushing.

```go
//go:build tinygo

package main

import (
	"sync/atomic"
	_ "unsafe" // needed for go:linkname
)

//go:linkname bugfixSyncAtomicStoreUint32 sync/atomic.StoreUint32
func bugfixSyncAtomicStoreUint32(addr *uint32, val uint32) {
	// this will get rewritten by tinygo into llvm atomic store
	atomic.StoreUint32(addr, val)
}

```

Issue #1459, opened 2022-03-01
MySQL/SQLite: support returning LastInsertId


housingdreams:
`:exec` only checks the error
`:execresult` returns the full [sql.Result](https://pkg.go.dev/database/sql#Result)
`:execrows` returns the sql.Result.`RowsAffected() (int64, error)`

We just need one more return type to support the full interface:

`:execlastinsertid` would just be a matter of calling `.LastInsertId()` rather than `.RowsAffected()` on the [sql.Result](https://pkg.go.dev/database/sql#Result).

This will avoid `sql.Result` database information leaking across the boundary of the `pkg.Querier` interface. It's not good for the service code to be handling DB specific method calls. This is a MySQL/SQLite issue as Postgres uses `RETURNING` so you can use `:one` instead.

Solution: [duplicate the template code used for :execrows](https://github.com/kyleconroy/sqlc/blob/4bbe1caf3744e00b6f0c92a5c75b69b08989e20b/internal/codegen/golang/templates/stdlib/queryCode.tmpl#L105-L125)

lpintes:
Recent versions of MariaDB supports returning clause in insert statement. So perhaps fixing parser could be sufficient.

housingdreams:
Unfortunately, there are a lot of companies still running older MySQL installs. In addition, using sqlc with products like TiDB or https://vitess.io/ would also require this.

Jille:
(I requested this before in https://github.com/kyleconroy/sqlc/issues/1200. I agree it would be nicer to have a :execinsertid)

PR #1484, opened 2022-03-11
Adding LastInsertId for Go

Xeoncross:
#1459 adds support for [sql.Result.LastInsertId()](https://pkg.go.dev/database/sql#Result)

`:execlastinsertid` is calls `.LastInsertId()` rather than `.RowsAffected()` with `:execrows`. 

This will avoid the sql.Result database package leaking across the boundary of the pkg.Querier interface generated by sqlc when using MySQL & MariaDB


internal/codegen/golang/templates/pgx/queryCode.tmpl
```diff
@@ -103,6 +103,23 @@ func (q *Queries) {{.MethodName}}(ctx context.Context, {{.Arg.Pair}}) (int64, er
...
+func (q *Queries) {{.MethodName}}(ctx context.Context, db DBTX, {{.Arg.Pair}}) (int64, error) {
+	result, err := db.Exec(ctx, {{.ConstantName}}, {{.Arg.Params}})
+{{- else -}}
+func (q *Queries) {{.MethodName}}(ctx context.Context, {{.Arg.Pair}}) (int64, error) {
+	result, err := q.db.Exec(ctx, {{.ConstantName}}, {{.Arg.Params}})
+{{- end}}
+	if err != nil {
+		return 0, err
+	}
+	return result.LastInsertId(), nil
```
Jille, line 119: pgx doesn't support driver.Result.LastInsertId. People usually do "INSERT INTO ... RETURNING id" with :one when they want the sequence number from Postgres.

You should probably return an error when people try to use :execinsertid with Postgres.

kyleconroy (commented):
> This will avoid the sql.Result database package leaking across the boundary of the pkg.Querier interface generated by sqlc when using MySQL & MariaDB

I'd rather leak the sql.Result than add another command. The reason is that we already support `:execresult`. I think adding `:execrows` was a mistake, so I don't want to make the same mistake twice.

kyleconroy:
> Can you share more about why :execrows was a mistake?

In this case, you're right that it's not much more work to support. Also, there are only two methods on sql.Result, so we aren't going to have to add any more in the future.

internal/codegen/golang/templates/pgx/queryCode.tmpl
kyleconroy, line 119: You want `return result.LastInsertId()` here.

kyleconroy: pgx also doesn't support the same signature for `RowsAffected()`. We'll need to address this is a different PR.

kyleconroy (changes requested):
Since `RowsAffected()` maps to `:execrows`, let's map `LastInsertId()` to`:execlastid`.

Also, please add tests similar to [exec_rows](https://github.com/kyleconroy/sqlc/tree/main/internal/endtoend/testdata/exec_rows) and [exec_result](https://github.com/kyleconroy/sqlc/tree/main/internal/endtoend/testdata/exec_result).

Xeoncross:
I tried to add the python generation example, but when I changed into the `internal/endtoend/testdata/exec_lastid/python_postgresql` folder and run it against my local sqlc it complained: 

```
$ ../../../../../sqlc2 generate
> error parsing sqlc.json: unknown target langauge "python"
```

I was using the same code from the [exec_rows](https://github.com/Xeoncross/sqlc/blob/feature-lastinsertid/internal/endtoend/testdata/exec_rows/python_postgresql/sqlc.json#L10) example.

I also wasn't sure what the PGX (go_postgresql_pgx) should be since this isn't something for PostgreSQL - so I left it.

kyleconroy:
It's okay that you weren't able to add the Python support, can do that in a subsequent PR.

You're also right that we don't need to add support for pgx, as it's Result type does not have this method.

kyleconroy (approved):

PR #1488, opened 2022-03-15
fix(makefile): test-endtoend working as intended

sbres:
The cd was one level too deep and was not running the tests
Also added added logging of the error on generation to help debug


sbres:
I'm not sure what you are testing with test-endtoend then.
As there is not a single test inside the testdata directory, is it just there to be sure the files in testdata are valid go files ?

kyleconroy:
> As there is not a single test inside the testdata directory, is it just there to be sure the files in testdata are valid go files ?

That's a very good point.

```
$ find internal/endtoend/testdata/ | grep _test
```

sbres:
I have no idea why `make test` failed the first time I tried that sent me on this tangent.

make test is correctly running the endtoend tests and does not need the db 👍

PR #1489, opened 2022-03-16
fix: Solve issue of commented SQL code treated as docstring

sbres:
When line are commented on the sql query it's taken into account for the docstring

- [ ] Create example

### DIFF
--- a/internal/source/code.go
+++ b/internal/source/code.go
@@ -89,13 +89,19 @@ func StripComments(sql string) (string, []string, error) {
 			continue
 		}
 		if strings.HasPrefix(t, "--") {
-			comments = append(comments, strings.TrimPrefix(t, "--"))
+			if len(lines) == 0 {
+				// We only store the first comment that should be the docstring
+				comments = append(comments, strings.TrimPrefix(t, "--"))
+			}
 			continue
 		}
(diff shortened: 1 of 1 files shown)


sbres:
This is actually quite a big task, the comments can be in a lot of multiple ways.
For example this query has almost all the tricky comments possible:
```
-- name: OneFooInlineComment2 :one
-- This function returns one Foo
SELECT * /*
What if the comment does not even start at the begining
*/
-- Extra comment that is not docsctring
/* Extra comment that is not docsctring  */
/*
  Extra comment that is not docsctring 
*/FROM foo -- This is a comment
/*
Super sneaky comment
*/WHERE bar = '42'/* Very Sneaky comment*/ OR bar = 'cow' /* 🐍 */
;
```

I've tried to re-write the StripComments function but I'm still unable to replicate the standard output of sqlc

kyleconroy:
@sbres Please feel free to re-open this if you get it to pass the tests at some point.

sbres:
Sorry, I've had some other things I had to jump to fix. As This was a quality of life improvement I did not try to push further.
Just to leave a record, if I or someone else tries this.
The regex is almost working there are some issues with the last line of the query I had not time to figure out.

PR #1518, opened 2022-03-30
Glob support

zmitry:
Hi, in our app we have migrations in different structure like this 
12321_folder/{up.sql, down.sql}
and current approach doesn't really work for us. Current change is kind of breaking change but maybe it's ok as it is.

### DIFF
--- a/internal/migrations/migrations.go
+++ b/internal/migrations/migrations.go
@@ -34,5 +34,5 @@ func RemoveRollbackStatements(contents string) string {
 
 func IsDown(filename string) bool {
 	// Remove golang-migrate rollback files.
-	return strings.HasSuffix(filename, ".down.sql")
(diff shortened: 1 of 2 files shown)


benefacto:
Glob support would be great for us as well since we're using [tern](https://github.com/JackC/tern), which supports environment variables in migrations via [its templates](https://github.com/JackC/tern#template-tips), for our migrations. We'd like to just exclude the migrations that have environment variables as we don't need models to be generated for them but this is not currently possible as we just have to provide a directory to sqlc.

kyleconroy:
@benefacto Could you provide a full example on how you'd use glob support to exclude those migrations?

I also just created https://github.com/kyleconroy/sqlc/issues/1824 to track adding official tern support.

kyleconroy (changes requested):
This is a backwards incompatible change with no tests, so I won't be merging it in its current form.

PR #1529, opened 2022-04-04
Golang SQL: Move defer rows.Close() before error check

aastein:
To be safe, move the deferred call to rows.Close before the error check, to handle the case where a connection is left open even though errors were encountered by the sql driver


aastein:
not needed

Issue #1498, opened 2022-03-19
Configure positional parameter logic


kyleconroy:
### What do you want to change?

For Go, sqlc will use a positional argument for a query if there's a single parameter. If there's more than one, it will generate a parameters struct. While this works for me, a few other folks would like different behavior.

@brentd would like to use positional arguments for up to three parameters (https://github.com/kyleconroy/sqlc/issues/1496)
@Threpio would like to always use parameter structs (https://github.com/kyleconroy/sqlc/issues/1373)

### How should we do it?

We can solve this using a single configuration parameter called `query_parameter_limit`, which defaults to 1 (the current behavior). We can set the value to a pointer in the configuration YAML / JSON so that zero is a valid option. Anything less than zero should be an error.

### What programming language backends need to be changed?

Go

iAziz786:
I am very new to sqlc, given I put some research in it I would like to pick this.

skabbes:
Just to chime in - I really like the positional argument for only 1 parameter - I would even like positional arguments for more parameters (maybe up to 3?), but _only_ if their types are distinct.  That's what I would like out of a Go generator.

When there are more than 3 parameters, then a struct is more ergonomic anyway. But when there are <= 3 parameters, of all distinct types, the compiler will protect you from transposing them, and it should be self-documenting _enough_.

**Crazy Idea**
Another kind of bizzare pattern that could be applied would be to apply the [functional options](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis) pattern(But applied to arguments, not options).  The boilerplate to facilitate it is rather tedious, but it would be generated anyway.  Here's what I'd be shooting for (in user-written code)

```go
_, _ := queries.CreateAuthor(ctx,
  db.Name("Stephen King"),
  db.Bio(sql.NullString{Valid: true, String: "wrote books"}))
  
// or maybe by using some better naming for the package names
_, _ := queries.CreateAuthor(ctx,
  params.Name("Stephen King"),
  params.Bio(sql.NullString{Valid: true, String: "wrote books"}))
```
This has a few side effects:

1. You would get full compile-time safety.  Forgot a parameter? compile error. Reordered parameters? compile error (but easy to fix, or maybe make them alphabetical?)
2. It would be slower than it is now, but still significantly faster than any reflection-based one
3. It is self-documenting at the _call_ site.
4. You have to write the parameters in exactly the right order :(. But I am not sure how to get around that without sacrificing compile safety.

[...]

PR #1530, opened 2022-04-05
Python query param limit

danicc097:
Related to #1498 , but gives support for a `query_parameter_limit` in Python codegen, very much desirable for my use case with FastAPI. 

Initially I wanted to also include Go, but I've failed miserably since all Go code generation uses a single param and don't know about the best way to proceed that requires minimal changes


kyleconroy (commented):
I updated the issue with a bit of different behavior

> We can set the value to a pointer in the configuration YAML / JSON so that zero is a valid option. Anything less than zero should be an error.

kyleconroy:
@danicc097 Sorry for taking forever on this code review. I can promise a quick review if you're able to fix the merge conflicts. If not, I totally understand.

danicc097:
@kyleconroy I noticed Python currently has a default limit of 4, the change to 1 would be breaking. Want to keep it anyway for consistency?

kyleconroy:
We should keep the default the same, even if it isn't the same as Go's

PR #1531, opened 2022-04-05
Inflection Singular table names exclusion configuration

### LINKED ISSUE #1526: Model generation inflection naming
### Version

1.12.0

### What happened?

While testing python codegen, I found a rather curious output: any table that ends in "**ta**" gets a model that ends in "**tum**" due to the inflection library, so e.g. many-to-many tables are particularly affected since I already use the singular form. A manual fix like with "meta" #1217  wouldn't work for these cases

Is it possible to allow passing a list of table names to the config file to omit in `inflection`'s `Singular()` instead? This way autogenerated SQL schemas will also be able to use it

Linux

danicc097:
Fixes #1526 

Not a fan of `exclude_table_names` being so ambiguous, but `emit_exact_table_names` is already taken. Any ideas for that or is it good enough?


kyleconroy (changes requested):
The code looks good, but I think you're right that `exclude_table_names` isn't a good name. I'll think about some alternatives that we can use.

kyleconroy:
Let's go with inflection_exclude_table_names. It's still not great, but I want to unblock us.

kyleconroy (approved):

kyleconroy:
@danicc097 You'll need to rebase as the configuration / proto files have changed.

kyleconroy (approved):

Issue #1525, opened 2022-04-02
RFC: nullability (and other metadata) for input arguments / parameters to a query 


skabbes:
### What do you want to change?

**Kudos**
Before I start, I wanted to give a huge kudos for sqlc - I love it. As a backend developer who knows SQL very well, I fight every single ORM I have used trying to do any SQL more complicated than CRUD.  Thank you for creating this.

**Background**
I use sqlc alongside Postgres, as sort of a foreign function interface to SQL.  I manage a PostgreSQL database with about 100 different tables and working in Typescript and Go to support every stage of solar operations for [Bright](https://www.ycombinator.com/companies/bright).  We do not currently use sqlc at Bright - but I do use it personally, but I am evaluating it Bright as well.

The types of problems I work on are not always CRUD-y (though sqlc works really well if they are).  And the challenge I am facing is that SQL is a little less-expressive than the target languages in some cases, and therefore using sqlc as a FFI is a little clunky.  A quick glance through the [some](https://github.com/kyleconroy/sqlc/issues/1446) [open](https://github.com/kyleconroy/sqlc/pull/1491) [issues](https://github.com/kyleconroy/sqlc/pull/1462), shows that "nullability" is one of those deficiencies.  sqlc does its best to always "do the right thing" and that has gotten it very far indeed. But I think a simple change can give developers full control, _and_ allow for sqlc to keep getting smarter (better interred mappings / types).

Say for example, instead of finding a really good GeoJSON library in my target language, I wanted to rely on PostGIS.
```sql
- name: WKBToGeoJSON :one
select ST_AsGeoJSON(ST_GeomFromWKB(sqlc.arg('wkb_bytes'))::text as output;
```

I would love for this Go code to be generated:
```Go
func (q *Queries) WKBToGeoJSON(ctx context.Context, db DBTX, wkbBytes []byte) (string, error) {
    // ...
}
```

**Or perhaps a more common use case, partial updates**
```sql
-- name: EditThing :exec
update thing
set
  name = coalesce(sqlc.arg('name'), name),
  description = coalesce(sqlc.arg('description'), description),
  status = coalesce(sqlc.arg('status'), status)
where thing.id = sqlc.arg(id)

-- name, description and status are all nullable arguments, id is not.
```

**Proposal**
I think sqlc should expose a way (like overrides) _in the query itself_ to control aspects of the generated code in the target languages.  

[...]

PR #1536, opened 2022-04-09
UPDATED: sqlc.narg() - allow user-specified nullability

skabbes:
EDITED from original PR after discussion here:  https://github.com/kyleconroy/sqlc/issues/1525

**Exec Summary**
I implemented user-specified nullability with `sqlc.narg`

 ```sql
sqlc.narg('param') -- regardless of whether this parameter was inferred to be nullable or not, it will be nullable in the generated code.
 ```

**Details**
Most of the work here came from building types to represent set of named parameters, which tracks the name and nullability status of any parameter (inferred or user defined).

`named.Param`
This type can track its nullability from the sqlc engine (inferred nullability) separately from any user-specified nullability, and always prefers the user-specified version to the inferred one.

`named.ParamSet`
This type is essentially a collection of params, and helps build up parameters, and encapsulate their positional representations as well.  As you can see - this type significantly cleaned up `compiler/resolve.go`

**Next Steps**
This code works, is tested. I think a code review would be great, and maybe a pointer to how to update the docs (or maybe the doc repos are still private).


internal/endtoend/testdata/sqlc_arg/mysql/query.sql
```diff
@@ -1,7 +1,13 @@
...
 SELECT name FROM foo WHERE name = sqlc.arg(slug);
 
 /* name: FuncParamString :many */
 SELECT name FROM foo WHERE name = sqlc.arg('slug');
+
+/* name: FuncParamStringOptional :exec */
+UPDATE foo SET name = coalesce(sqlc.arg('slug?'), name);
+
+/* name: FuncParamStringRequired :exec */
+UPDATE foo SET description = sqlc.arg('slug!');
```
skabbes, line 13: Example usage of sqlc.arg with user-defineable nullability.

internal/sql/rewrite/parameters.go
```diff
@@ -41,55 +41,61 @@ func isNamedParamSignCast(node ast.Node) bool {
...
+	last := name[len(name)-1]
+	if last == '!' {
+		return named.NewUserDefinedParam(name[:len(name)-1], true)
+	}
+
+	if last == '?' {
+		return named.NewUserDefinedParam(name[:len(name)-1], false)
+	}
+
+	return named.NewUnspecifiedParam(name)
```
skabbes, line 60: This is the core logic for the nullability options, I'm happy to rip this out and replace it with anything else.

internal/sql/rewrite/parameters.go
```diff
@@ -41,55 +41,61 @@ func isNamedParamSignCast(node ast.Node) bool {
...
-				if _, found := args[param]; !found {
-					args[param] = []int{argn}
-				} else {
-					args[param] = append(args[param], argn)
-				}
-				cr.Replace(&ast.ParamRef{
-					Number:   argn,
-					Location: fun.Location,
-				})
-			}
```
skabbes, line 81: _aside_: encapsulating the positional tracking into `named.ParamSet` significantly cleaned up this code.

internal/source/code.go
```diff
@@ -54,25 +54,31 @@ func Mutate(raw string, a []Edit) (string, error) {
 	if len(a) == 0 {
```
skabbes, line 54: I was having a bit of trouble with this, so I tested the heck out of it and these were the resulting edits.

skabbes:
@kyleconroy I went ahead and implemented `sqlc.narg` like we discussed, could you take a look at you earliest convenience?

internal/source/code.go
kyleconroy: The edit code is super hacky so I'm not surprised you needed to add all these tests. Hopeful in the future we can get rid of all of this code and generate SQL using an AST.

kyleconroy (approved):

kyleconroy:
It's a lot to ask, but can you also add docs to https://docs.sqlc.dev/en/stable/howto/named_parameters.html in a separate PR? Want to make sure that people know this option exists.

skabbes:
ah, I didn't look hard enough for the docs. But yes I can do that.

skabbes:
Added here: https://github.com/kyleconroy/sqlc/pull/1579

skabbes:
Just kidding, one that actually passes CI here: https://github.com/kyleconroy/sqlc/pull/1580

PR #1538, opened 2022-04-10
Pydantic models config field

danicc097:
Since dataclasses are the default since #939, I added an optional `use_pydantic_models` field to use Pydantic for all models (discussed since #923) which requires much less boilerplate later at least for my use case


protos/plugin/codegen.proto
```diff
@@ -68,6 +68,7 @@ message PythonCode
   bool emit_async_querier = 3;
   string package = 4;
   string out = 5;
+  bool use_pydantic_models = 6;
```
kyleconroy, line 71: To match the other configuration options, rename this to `emit_pydantic_models`.

examples/python/requirements.txt
```diff
@@ -3,3 +3,4 @@ pytest-asyncio~=0.14.0
 psycopg2-binary~=2.8.6
 asyncpg~=0.21.0
 sqlalchemy==1.4.0
+pydantic~=1.9.0
```
kyleconroy, line 6: remove

internal/plugin/codegen.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.27.1
-// 	protoc        v3.19.4
+// 	protoc-gen-go v1.26.0
+// 	protoc        v3.6.1
```
kyleconroy, line 4: Please install the same versions of protoc and protoc-gen-go.

examples/python/sqlc.json
```diff
@@ -10,7 +10,8 @@
           "out": "src/authors",
           "package": "authors",
           "emit_sync_querier": true,
-          "emit_async_querier": true
+          "emit_async_querier": true,
+          "use_pydantic_models": true
```
kyleconroy, line 14: Instead of updating the example, please add a new end-to-end test for Pydantic models.

kyleconroy (changes requested):
Thanks for the first draft. Requested a few changes; let me know if you have any questions.

examples/python/requirements.txt
danicc097: isn't this required for ci to pass?

kyleconroy (approved):

PR #1554, opened 2022-04-16
Prepend an underscore to a StructName if it starts with a digit.

avalonbits:
This situation can occur if the column name starts with an underscore prefix and the following character is a digit:

```
create table thing (
 _1099Form text
);
```

currently the generator would balk because  the generated struct would be :

```
type Thing struct {
  1099Form string
}
```

and `1099Form` is not a valid name (it starts with a digit).

with this change we would instead get instead:

```
type Thing struct {
  _1099Form string
}
```

### DIFF
--- a/internal/codegen/golang/struct.go
+++ b/internal/codegen/golang/struct.go
@@ -2,6 +2,8 @@ package golang
 
 import (
 	"strings"
+	"unicode"
+	"unicode/utf8"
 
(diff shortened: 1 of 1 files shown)


kyleconroy (approved):
Thanks for the code! If the tests pass, I'll merge (and add a test too)

PR #1558, opened 2022-04-18
feat(Go):Add query_parameter_limit conf to codegen

go-mez:
Related to: #1498.

I'm not sure about the approach I took, so just let me know if im solving it wrongly


go-mez:
@kyleconroy sorry, now its passing all the tests, i think... I was not aware of the dev guide (https://github.com/kyleconroy/sqlc/blob/main/docs/guides/development.md)... i was just running `make` but it wasnt throwing any error

docs/reference/config.md
```diff
@@ -78,6 +79,8 @@ Each package document has the following keys:
   - Customize the name of the querier file. Defaults to `querier.go`.
 - `output_files_suffix`:
   - If specified the suffix will be added to the name of the generated files.
+- `query_parameter_limit`:
+  - Positional arguments that will be generated in go functions (>= `1` or `-1`). To always emit a parameter struct, you would need to set it to `-1`. `0` is invalid. Defaults to `1`.
```
kyleconroy, line 83: I changed my mind here. We should allow for zero to be a valid value. Anything less than zero should be an error.

internal/cmd/shim.go
```diff
@@ -75,6 +75,10 @@ func pluginPythonCode(s config.SQLPython) *plugin.PythonCode {
 }
 
 func pluginGoCode(s config.SQLGo) *plugin.GoCode {
+	if s.QueryParameterLimit == nil {
+		s.QueryParameterLimit = new(int32)
```
kyleconroy, line 79: Let's default this to 1

kyleconroy (commented):
This looks good. Left a few comments about the new behavior we should support.

kyleconroy:
@go-mez Sorry for taking so long to get you a review. I know you've resolved conflicts a few times. I can promise that you'll only need to do it one more time. Thanks again for understanding.

kyleconroy:
Not sure what's happened, but the tests haven't run for this PR. Maybe another rebase and push will fix things?

polderudo:
@go-mez any chance to get this rebased and pushed? seems like only the docs were waiting... Would be cool to have that feature

go-mez:
@polderudo i'll do it. dont worry!

go-mez:
@kyleconroy i think it is okay now. could you approve the workflows please ?

go-mez:
@kyleconroy could you please take a look at the PR and approve the workflows?

internal/codegen/golang/field.go
```diff
@@ -77,3 +77,12 @@ func toCamelInitCase(name string, initUpper bool) string {
 	}
 	return out
 }
+
+func toLowerCase(str string) string {
+	var b strings.Builder
```
josharian, line 82: Note that this panics on an empty string. And it is over complicated. 

I suggest:

```go
if str == "" {
  return ""
}
return strings.ToLower(string[:1]) + string[1:]
```

internal/endtoend/testdata/process_plugin_sqlc_gen_json/gen/codegen.json
```diff
@@ -15,6 +15,8 @@
       "plugin": "jsonb",
       "options": "eyJmaWxlbmFtZSI6ImNvZGVnZW4uanNvbiIsImluZGVudCI6IiAgIn0="
     },
+    "python": null,
+    "kotlin": null,
```
josharian, line 19: this change seems unrelated. revert?

protos/plugin/codegen.proto
```diff
@@ -192,4 +193,4 @@ message CodeGenRequest
 message CodeGenResponse
 {
   repeated File files = 1 [json_name="files"];
-}
+}
```
josharian, line 196: restore trailing newline please

josharian (commented):
I left a few comments scattered about.

I really like this (and I would set the parameter to 3 or 4 myself), with one caveat.

I would personally prefer a struct in any case (no matter how few params) in which all the parameter types aren't distinct. That is, these are all good function signatures:

```go
func SomeQuery(ctx context.Context, id int32)
func SomeQuery(ctx context.Context, id int32, s string)
func SomeQuery(ctx context.Context, id int32, s string, created time.Time, x SomeEnum, b bool)
```

But this one isn't:

```go
func SomeQuery(ctx context.Context, id1 int32, id2 int32)
```

Even though it has only two params, they have the same type, which makes it easy to accidentally transpose them.

(It would be even better if each model's ID automatically had its own type, like `type UserID int32`, but that's a different topic. That can be achieved manually as needed.)

internal/codegen/golang/result.go
```diff
@@ -162,14 +162,16 @@ func buildQueries(req *plugin.CodeGenRequest, structs []Struct) ([]Query, error)
...
+
+		if len(query.Params) == 1 && qpl != 0 {
 			p := query.Params[0]
 			gq.Arg = QueryValue{
 				Name:      paramName(p),
 				Typ:       goType(req, p.Column),
 				SQLDriver: sqlpkg,
 			}
-		} else if len(query.Params) > 1 {
+		} else if len(query.Params) >= 1 && (qpl >= 0) {
```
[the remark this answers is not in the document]
go-mez: This validation is done when the config is being parsed:

https://github.com/kyleconroy/sqlc/blob/ae7a7bd894c4dbff8f7f2e5100c7c8b0a6ec4e86/internal/config/v_one.go#L79

https://github.com/kyleconroy/sqlc/blob/ae7a7bd894c4dbff8f7f2e5100c7c8b0a6ec4e86/internal/config/v_two.go#L78

internal/endtoend/testdata/process_plugin_sqlc_gen_json/gen/codegen.json
```diff
@@ -36,8 +38,7 @@
       "output_files_suffix": "",
       "emit_enum_valid_method": false,
       "emit_all_enum_values": false,
-      "inflection_exclude_table_names": [],
-      "emit_pointers_for_null_types": false
```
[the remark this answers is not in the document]
go-mez: the file was regenerated again. i think it is correctly done now

go-mez:
@josharian thank you for the review and the suggestion. i will try to address it

internal/codegen/golang/result.go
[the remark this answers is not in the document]
josharian: Right. So if qpl can't be negative, we could remove the second check here.

josharian (commented):
lgtm, modulo on little comment and my larger design comment above.

guess it's back to @kyleconroy

josharian:
ping @kyleconroy

kyleconroy (approved):

polderudo:
Works nice. Thanks for the work, although there is a issue with pgx.
Using `sql_package: "pgx/v5" `
and building with
`sqlc generate --experimental -f sqlc.yaml`
will not add the import to
`"github.com/jackc/pgx/v5/pgtype"`
to the generated files.

polderudo:
Did you find any time to look into this go-mez?

go-mez:
@polderudo do you know if this error was happening before? ill try to check it anytime soon

polderudo:
It's working, if i remove the new config
"query_parameter_limit": 999

Here is a palyground:
https://play.sqlc.dev/p/29a65ea348dd6ce7915b9106fc8e2eecb6ea83b53705ae72d9096ac4ac7c56d4

polderudo:
ping. any update?

PR #1560, opened 2022-04-21
Refactor struct name

jakoguta:
This PR refactors `func StructName(name string, settings *plugin.Settings) string` to `func StructName(name, rename string) string` to make it easier to isolate during test. It only needs renamed string from the `settings *plugin.Settings`, thus thee is no need to provide the entire `settings *plugin.Settings`.

In addition `strings.Title(s string)` from the Go strings package has been deprecated in Go v1.18. Therefore, we should instead use `cases.Title(language.English).String("myString")` from the cases package.


jakoguta:
I am working on the errors.

jakoguta:
I have removed the unit tests.

jakoguta:
This pull request in now not relevant as the updates it proposed have already been handled by #1595.

Issue #296, opened 2020-01-24
Add a Typescript language backend


kyleconroy:
sqlc, like protoc, aims to generate code for multiple languages. The second language we're looking to add is TypeScript.

cmoog:
With typescript, there are a few additional considerations in terms of which driver to use. Here is my first pass at what the sqlc output could look like: 

https://gist.github.com/cmoog/c7b612f6aa829c426ec665ad7cfcb7cb

any thoughts?

g5becks:
@kyleconroy 

As someone who comes from using mostly Typescript for backend work, I'm not sure what sqlc can add to that landscape?

I'm sure you "could" do it and the implementation would be great, but there are already so many best of breed options that do the same thing for typescript.

[PgTyped](https://pgtyped.now.sh/)
[Zapatos](https://jawj.github.io/zapatos/), (my personal favorite)
[Prisma](https://www.prisma.io/), (the most popular) . 

For Kotlin, theres [Jooq](https://www.jooq.org/doc/3.1/manual/getting-started/jooq-and-kotlin/) and [Exposed](https://github.com/JetBrains/Exposed) that I know of, which are both VERY mature.

I'd love to see sqlc committed to providing the best Go experience there is since both Typescript and Kotlin have great options for providing the same functionality already.

debackerl:
I had a look at PgTyped and Zapatos, but that quite different than what sqlc proposes: they both seem to require an existing database where the schema already has been deployed, and then only, the CLI tool can pull the schema and generate code.

sqlc is parsing the DDL SQL files only, no database needed, so it depends how the developer works:
1) either the developer works by prototyping his schema live in a local DB (for example), generate his code and migration files from there.
2) or the developer maintains his migration files first, iterate quickly, and only deploy to a DB once it's stable enough. What makes sqlc unique here is its use of pg_query as a robust query parser.

I wouldn't push for one style or another, however, sqlc is clearly for case 2, and the other tools doesn't help there. I fall in case 2, using Sqitch to handle my migration files in a more language-agnostic fashion.

guzmonne:
For PostgreSQL I would suggest Slonik.

- It doesn't depend on a database.
- Wraps `node-pg` with security layers.

skabbes:
[...]

PR #1565, opened 2022-04-23
feat: Add JSON "codegen" output

kyleconroy:
Inspired by @skabbes here https://github.com/kyleconroy/sqlc/pull/1470#issuecomment-1086681778

Add a `json` option to the codegen output. You'll need to use version 2 of the configuration file. You can configure the output path using the `out` option.

```json
{
  "version": "2",
  "sql": [
    {
      "schema": "postgresql/schema.sql",
      "queries": "postgresql/query.sql",
      "engine": "postgresql",
      "gen": {
        "json": {
          "out": "gen",
          "indent": "  "
        }
      }
    }
  ]
}
```

```
sqlc generate --experimental
cat gen/codegen_request.json | gen-go-code
cat gen/codegen_request.json |  gen-ts-code
cat gen/codegen_request.json | wasm-runner ./python-gen.wasm
```

- [ ] Add documentation for the JSON plugin


skabbes:
This is incredible! It is a testament to the architecture how trivial this was to set up.

At $COMPANY, we rely heavily on code generation for our Typescript and Go base models (~100 tables).  Our architecture is basically this pipeline (collectively called "sync schema")

* A Go program (`readschema`) that:
  * runs information_schema (and pg_*) queries against a dev (or prod!?) database 
  * generates a json file
* A Typescript program (`writeSchema.ts`) that:
  * reads the JSON, and runs them though handlebars templates
  * generates TableName_gen.ts and TableName_gen.go files
  * applies auto-formatting (go fmt, prettier, etc)

This change could completely replace `readschema` for us, and would allow any improvements there to instead be contributed to directly sqlc.  Additionally, the code generators we have would be immediately be useful to others (sqlc -> Typescript) for example (the sqlc Go code generator is already light years ahead of ours though).

Anyway, that's a long winded -hell ya!  And I also think this change will inspire someone (maybe me?) to just knock out this 2 year old issue. https://github.com/kyleconroy/sqlc/issues/296

kyleconroy:
> This change could completely replace readschema for us, and would allow any improvements there to instead be contributed to directly sqlc. 

That's great to hear. Once this is merged, I'd love to hear if it's working out for you.

> And I also think this change will inspire someone (maybe me?) to just knock out this 2 year old issue. https://github.com/kyleconroy/sqlc/issues/296

There shouldn't anything blocking you (or anyone else) on implementing codegen for another language. The internal codegen interface is now stable. You can use this PR as a template for adding others.

PR #1569, opened 2022-04-25
add support for custom Go struct tags

### LINKED ISSUE #534: Add support for custom struct tags

josharian:
This PR adds a new type of override: go_struct_tag.
When provided for a field, it adds that struct tag to the generated code.
The provided struct tag is parsed according to the standard
package reflect rules, and its components are updated independently.
This allows struct tag overrides to be compatible with (and optionally
override) autogenerated json and db struct tags.

This PR also includes some minor preparatory cleanup and fixes.
I recommend reviewing it one commit at a time.
The early commits are small and easy; the last commit contains the new feature.

Fixes #534


nahwinrajan:
Hi there, thank you for implementing this feature.

I do not see this feature from documentation, and on original thread ( #534) there are two design that were discussed:
- author suggestion per field on the SQL Table
- Kyle's counter recommendation which is on override level.

may I know which design actually implemented and some example ?
Am hopeful it is the TS idea since it can really integrate well with validator library.

josharian:
@nahwinrajan search for `go_struct_tag` in https://docs.sqlc.dev/en/latest/reference/config.html

FrankFang:
How could a use it? Is there an example

adeyemialameen04:
> How could a use it? Is there an example
> 
> ```shell
> ❯ sqlc generate 
> error parsing sqlc.yaml: yaml: unmarshal errors:
>   line 12: field go_struct_tag not found in type config.SQLGo
> ```
> 
> sqlc.yaml
> 
> ```yaml
> version: "2"
> sql:
>   - engine: "postgresql"
>     queries: "config/sqls"
>     schema: "config/schema.sql"
>     gen:
>       go:
>         emit_json_tags: true
>         json_tags_case_style: snake
>         package: "queries"
>         out: "config/queries"
>         go_struct_tag: 'a:"b" x:"y,z"'
> ```

2yrs later, figured it out? 😂, honestly have no idea on how to use it

adeyemialameen04:
Honestly the first design would have been better, adding all of these to the config file is just nuts

Issue #814, opened 2020-11-25
null overrides `timestamp` to `*time.Time` not working


mvrhov:
I'm not sure and I might be missing something, but null overrides at least for `timestamp` to `*time.Time`, doesn't seem to wok, `text`, to *string works, `jsonb` to `*json.RawMessage`

I've also tried with `timestamptz, timestamp(0), timestamp without time zone` I'm unable to debug this as the go compiler fails when building debug version in Goland with OOM.. it seems that 8G of free ram is not enough.

```yaml
version: "1"
packages:
  - name: "db"
#....
overrides:
  - db_type: "timestamp"
    go_type:
      import: "time"
      package: "time"
      type: "Time"
      pointer: true
    nullable: true
  - db_type: "jsonb"
    go_type:
      import: "json"
      package: "json"
      type: "RawMessage"
      pointer: true
    nullable: true
```

```sql
CREATE TABLE office (
    id_office bigint NOT NULL GENERATED BY DEFAULT AS IDENTITY,

    created_at timestamp(0) DEFAULT now() NOT NULL,
    updated_at timestamp(0),
);

-- name: ListOffices :many
SELECT * FROM office
ORDER BY name;
```

This always generates

```go
type Office struct {
	IDOffice    int64
	CreatedAt   time.Time
	UpdatedAt   sql.NullTime
}
```

mvrhov:
Meh.. this seems to somewhat work but it's tedious and almost impossible to guess.

```yaml
  - db_type: "pg_catalog.timestamp"
    go_type:
      import: "time"
      type: "Time"
      pointer: true
    nullable: true

  - db_type: "pg_catalog.int2"
    go_type:
      type: "int16"
      pointer: true
    nullable: true
  - db_type: "pg_catalog.int4"
    go_type:
      type: "int"
      pointer: true
    nullable: true
  - db_type: "pg_catalog.int8"
    go_type:
      type: "int64"
      pointer: true
    nullable: true
  - db_type: "pg_catalog.bool"
    go_type:
      type: "bool"
      pointer: true
    nullable: true
  - db_type: "pg_catalog.float4"
    go_type:
      type: "float32"
      pointer: true
    nullable: true
  - db_type: "pg_catalog.float8"
    go_type:
      type: "float64"
      pointer: true
    nullable: true
```

jskelcy:
hmm I am trying to use this and its not working for me. Anything tricky you did here @mvrhov ?

[...]

PR #1571, opened 2022-04-26
Add support for emitting pointer types for nullable columns

nickbruun:
pgx/v4 supports passing and scanning into pointer types for nullable columns instead of using the standard library `Null<Type>` representation, which is super handy if you wind up interfacing a lot with, say, generated Protocol Buffer files using the proto3 `optional` flag.

This PR adds a new configuration option `emit_pointers_for_null_types`, which, as per the documentation line added:

> If true and `sql_package` is set to `pgx/v4`, generated types for nullable types are emitted as pointers (ie. `*string`) instead of `database/sql` null types (ie. `NullString`). Defaults to `false`.

This change would be a significant quality-of-life improvement for me, but I obviously don't know if it generally fits with the philosophy of sqlc, so consider this a suggestion and please feel free to provide any feedback (including, but not limited to bike-shedding of the choice of name for the configuration parameter, if we see this feature fit for sqlc.)

Lastly, thanks all for building this tool. It's genuinely saved me tons of hours writing the same pesky boiler plate with limited guarantees.


topheruk:
how can I try this out? I have been looking at the alternative which is to manually override but that is pretty tedious. would like to try this out just don't know how to include it.

mvrhov:
This was implemented before (named a bit differently), but it has been removed. #814 has a list of needed overrides. Which is a bit tedious to copy around I'd say.

topheruk:
@mvrhov so is it just recommend to add them manually then I'm guessing?

Thanks for this list as it does help.

mvrhov:
For now I guess. I do like this PR because then I don't have to copy a bunch of lines into each project.

nickbruun:
Certainly avoiding having to copying a bunch of lines around would be a huge improvement. Any chance this could be considered for a merge, despite prior history, @mvrhov?

mvrhov:
@nickbruun It's not my decision on what's getting merged.

nickbruun:
@mvrhov I understand – mostly just wondering what historical context you can provide since it sounds like this was tried before?

kyleconroy:
@nickbruun Sorry that I'm just getting around to getting this reviewed. I know it's been months since it was opened.

I'm onboard with this change and would like to get it merged for the pgx/v5 release https://github.com/kyleconroy/sqlc/issues/1823

kyleconroy (commented):
How should this work in conjunction with the new [pgtype](https://pkg.go.dev/github.com/jackc/pgx/v5@v5.0.0-beta.3/pgtype) package? I think we should only return pointers to non-pgtype types. Thoughts?

nickbruun:
Let me review the changes from pgx/v5 and I'm happy to update in accordance :)

brlala:
would be good to have this feature in v5

JamesArthurHolland:
Is this going to be merged soon? I'm using sqlc, gqlgen and gqlgenc, getting an error about String is incompatible with database/sql.NullString and I'd much rather use the pointers for null types.

nickbruun:
Oh wow, I totally got behind on developing a coherent view on pgx/v5. Thanks for merging this!

Issue #1097, opened 2021-07-28
Handle Postgres 'NULL' types


fresonn:
Hey!
For example, I have a schema
```sql
CREATE TYPE tag AS ENUM ('new', 'stock', 'top');

CREATE TABLE some_table (
    ...
    mainTag tag
    ...
)
```
You can see that the _mainTag_ field in the table can be **null**.
And for the corresponding table, 'sqlc' generates me the following code
```go
type TagType string

const (
	TagTypeNew   TagType = "new"
	TagTypeStock TagType = "stock"
	TagTypeTop     TagType = "top"
)

func (e *TagType) Scan(src interface{}) error {
	switch s := src.(type) {
	case []byte:
		*e = TagType(s)
	case string:
		*e = TagType(s)
	default:
		return fmt.Errorf("unsupported scan type for TagType: %T", src)
	}
	return nil
}
```

After I try to translate data from the database, I always get an error in the default case: "unsupported scan type for ...."

Question.
SQLC can track that this field can have a **NULL** value in the schema?

kyleconroy:
I think the best option here to generate a second type that can handle NULL values. This will also be backwards compatible with the existing type.

```go
type NullTagType struct {
    TagType TagType
    Valid bool
}
```

sbres:
I'm looking at this problem and I think I can try at least scope out the work needed.

In order to replicate the Nullable values form the SQL library, wee need the equivalent of `convertAssignRows`
After looking at some PR https://github.com/golang/go/issues/24258 for example it is clear the method will stay private.
It is not possible to only copy-paste the method as I found it not that easy to extract it.
This library has the equivalent function but is easy to copy paste:
https://github.com/denisenkom/go-mssqldb/blob/master/convert.go

1- We need an equivalent to convertAssign.
Should we add a convert.go file on the generation with convertAssign from go-mssqldb ?
Should we add convertAssign to sqlc as a public method ?

2- Once we have convertAssign we can create the NullXXX type needed with it's equivalent Scan and Value methods
Should we generate this type even when the type is never used? 
If we can just generate it no matter what becomes very straightforward.

3- Use the Nullable value instead of the raw type on the needed queries
I actually have not vision on this part at all.

@kyleconroy If you can tell me what you prefer for the point N-1 I can try to draft a PR with the changes, at least the points 1, 2

Jille:
[...]

PR #1485, opened 2022-03-14
Feat/nullable enums

sbres:
With this PR now custom Enums have a nullable structure like all the sql.NullXXXX

The Null structure is generated for every Enum.
When enums are present on the model we import `database/sql/driver`
The database type matching is a little bit dirty I think something better can be done, let me know if you have an idea.

Also, how should I go about creating tests for this?

PR #1582, opened 2022-05-02
feat: Nullable enums

Jille:
For every enum X, generate a type NullX which is a struct with X and
Valid bool.

issue #1097

Remaining work:

- [ ] Get agreement from @kyleconroy 
- [ ] Discuss whether to put in effort to skip generation of NullX when it's unused.
- [ ] Discuss whether the backwards compatibility break (because some inputs/outputs will become NullX) is a problem
- [ ] Test with pgx
- [ ] Test with database/sql postgres
- [ ] Test with database/sql mysql

### DIFF
--- a/examples/batch/postgresql/models.go
+++ b/examples/batch/postgresql/models.go
@@ -5,6 +5,7 @@
 package batch
 
 import (
+	"database/sql/driver"
 	"fmt"
 	"time"
 )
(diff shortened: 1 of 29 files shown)


Jille:
Hmm, looks like this does the same thing as #1485. Let's see if that PR gets merged and we can drop this one at that point.

@SnoozeThis https://github.com/kyleconroy/sqlc/pull/1485 <= 1 month

SnoozeThis:
The deadline passed while snoozing.

kyleconroy:
Closing as we merged https://github.com/kyleconroy/sqlc/pull/1485

PR #1601, opened 2022-05-09
feat: Split models into dedicated package

dkhenry:
This change allows models to be split into a different package. The
major benefit of this is to allow multiple sqlc querier packages to use
the same scheme, and models. This allows better code modularization and
help larger projects keep implementation details independent across
logical code modules


dkhenry:
@kyleconroy I am not sure how well this is going to work with the current tests, I have it set up so you can override the package of the model's and give a relative path where they should sit. That is how I am currently using it, but it breaks the tests.  Do the tests have a go.mod file and a package I could target ?

dkhenry:
@kyleconroy can you approve the workflows ? I think the tests should be passed

dkhenry:
@kyleconroy any feedback here ? I believe if you approve the workflows it will pass the tests

kyleconroy (commented):
Thanks for the PR and the tests.

This implementation feels off. I think the reason is that we'll regenerate the model code for each package. If these packages had different schema paths, this would lead to strange output that is order dependent. 

Instead, what if we split the schema and queries out into separate configuration options? A set of queries could depend on a set of schemas. This would allow you to define a set of schemas once and re-use it across different queries.

That said, we should start a proposal for this before doing any more work on this pull request.

dkhenry:
That sounds great. I will close this PR and get started on a proposal to implement the design

PR #1613, opened 2022-05-11
internal/codegen: add Enum.Valid and AllEnumValues

### LINKED ISSUE #1607: autogenerate enum validation methods

The codegen for enums generates a list of constants containing valid values:

```go
type EnumType string

const (
	EnumTypeA EnumType = "A"
	EnumTypeB EnumType = "B"
)
```

When writing developer tools, it can be helpful to convert from a string value to an enum type. However, there is no exhaustive list of valid enum values, so we have to wait until a query fails to discover that an invalid enum value has been provided.

I'd like the codegen to also emit something like:

```go
func (e EnumType) Valid() bool {
  switch e {
  case EnumTypeA, EnumTypeB:
    return true
  default:
    return false
  }
}
```

Then I can convert from string to enum type and check valid early on.

I know that this applies to PostgreSQL and Go. I'm not sure about MySQL and the other languages.

PostgreSQL

Go

josharian:
This commit adds a Valid method for enum types
and an All...Values function to get all enum values.

This makes it easier to work with enums.

Fixes #1607


kyleconroy (changes requested):
Thanks for making a first pass at this. A few things before this is ready to merge. 

For enums with many values, the generated code takes up a large amount of horizontal space. Can we put each enum on its own line for both methods?

```go
func (e BooksBookType) Valid() bool {
	switch e {
	case BooksBookTypeFICTION:
		return true
	case BooksBookTypeNONFICTION:
		return true
	}
	return false
}

func AllBooksBookTypeValues() []BooksBookType {
	return []BooksBookType{
		BooksBookTypeFICTION,
		BooksBookTypeNONFICTION,
	}
}
```

This feature has only been requested by a few people. I'd like to put both behind a feature flag. Let's go with `emit_enum_valid_method` and `emit_all_enum_values`. I know the emit prefixes are a bit weird, but it's what we use for other options.

josharian:
> Can we put each enum on its own line for both methods?

We can, although for the Valid method the compiler will generate better code if we can make them part of a single case statement, so it can unify the return statements. (That's also the reason not to write the code `return e == EnumA || e == EnumB || ... || e == EnumX`.)

What do you think about this:

```go
func (e BooksBookType) Valid() bool {
	switch e {
	case BooksBookTypeFICTION,
                 BooksBookTypeNONFICTION:
		return true
	}
	return false
}
```

kyleconroy:
I think it's more straightforward to put one on each line. The comma style with a single case statement looks good to me.

josharian:
OK, there's now one per line. Go's templates sure don't make that easy.

This is now gated by config parameters. There's a surprising amount required to add a config parameter; I hope I didn't miss anything.

There is also an additional commit to make debugging regeneration failures easier to diagnose. (I'd like to do https://twitter.com/commaok/status/1345977948453490688 as well, but this is a good first start.)

jakoguta:
The test data `emit_enum_valid_and_values` added is for PostgreSQL. I recommend we put them in the folder `postgresql/pgx`.

josharian:
> The test data emit_enum_valid_and_values added is for PostgreSQL. I recommend we put them in the folder postgresql/pgx.

Genuine question: Does it matter? This codegen change is DB-agnostic.

josharian:
Ping.

What expectations should I have for review and issue comment turnaround time? It’d be useful to know so that I can plan accordingly and also not pester pointlessly.

scripts/regenerate/main.go
```diff
@@ -21,9 +21,9 @@ func regenerate(dir string) error {
 			cwd := filepath.Dir(path)
 			cmd := exec.Command("sqlc-dev", "generate", "--experimental")
 			cmd.Dir = cwd
-			failed := cmd.Run()
+			out, failed := cmd.CombinedOutput()
 			if _, err := os.Stat(filepath.Join(cwd, "stderr.txt")); os.IsNotExist(err) && failed != nil {
-				return fmt.Errorf("%s: sqlc-dev generate failed", cwd)
+				return fmt.Errorf("%s: sqlc-dev generate failed\n%s", cwd, out)
```
kyleconroy, line 26: out is a byte slice, so I think this will print out gibberish.

kyleconroy (approved):
One small issue in the regen script, going to merge anyways

scripts/regenerate/main.go
josharian: If the verb were `%v` then it would. With `%s` it prints as a string.

PR #1623, opened 2022-05-16
Fix postgresql time

jld3103:
Go's `time.Time` type is not compatible with Postgres' `pg_catalog.time` type. Therefore this fix uses `pgtype.Time` which is the correct one. `pg_catalog.timetz` would also need to be fixed, but there is no `pgtype.TimeTZ` (see https://github.com/jackc/pgtype/issues/150 and https://github.com/jackc/pgx/issues/759).

### DIFF
--- a/internal/codegen/golang/imports.go
+++ b/internal/codegen/golang/imports.go
@@ -150,6 +150,7 @@ var pgtypeTypes = map[string]struct{}{
 	"pgtype.Macaddr":   {},
 	"pgtype.Numeric":   {},
 	"pgtype.Numrange":  {},
+	"pgtype.Time":      {},
 	"pgtype.Tsrange":   {},
 	"pgtype.Tstzrange": {},
 }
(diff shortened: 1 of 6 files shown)


kyleconroy (commented):
The standalone pgtype package has been deprecated as of pgx/v5. The types will also no longer be compatible with lib/pq.

With that in mind, adding pgx/v5 support will fix this. But we're going to do that all at once.

kyleconroy:
Going to close this PR. We'll address this when adding support for pgx/v5

GeertJohan:
Having it fixed with pgx/v5 sounds good. But is there any reason why we cannot merge this (realtively small change) now so that it already works with this version?

kyleconroy:
The reason I don't want to merge this is the unclear future of the pgtype package. I grieve it will be deprecated in less than a week.

StevenACoffman:
Hey, so jackc has continued to maintain pgtype, and I would love to get this merged until we do the pgx/v5 swap.

PR #1626, opened 2022-05-17
[FIX] copyfrom imports

positiveblue:
Depending on the commands used in your `query.sql` the `copyfrom.go`
may end up with unused package imports.

There is a patch for this problem that was merged in the #1386 PR.
However, it looks like some edge cases were left out of the fix.

This commit ensures that `buildImports` only has `copyfrom` queries
when called from `copyfromImports`.

### DIFF
--- a/internal/endtoend/testdata/copyfrom_imports/postgresql/pgx/query.sql
+++ b/internal/endtoend/testdata/copyfrom_imports/postgresql/pgx/query.sql
@@ -6,3 +6,7 @@ INSERT INTO myschema.foo (a, b) VALUES ($1, $2);
 
 -- name: InsertSingleValue :exec
 INSERT INTO myschema.foo (a) VALUES ($1);
+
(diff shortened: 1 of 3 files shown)


internal/endtoend/testdata/copyfrom_imports/postgresql/pgx/query.sql
```diff
@@ -6,3 +6,7 @@ INSERT INTO myschema.foo (a, b) VALUES ($1, $2);
 
 -- name: InsertSingleValue :exec
 INSERT INTO myschema.foo (a) VALUES ($1);
+
+-- name: DeleteValues :execresult
```
positiveblue, line 10: This was the main offender, any file generated from a query file with queries of type `metadata.CmdExecResult` would include the `github.com/jackc/pgconn` package

kyleconroy (approved):

PR #1629, opened 2022-05-19
fix citext resolution issues

### LINKED ISSUE #1581: citext on column resulted to interface{} as type
### Version

Other

### What happened?

Version
Other

What happened?
Hint: I am using v1.13.0 with `PostgreSQL` DB to generate the Go code. 

### Result:
```json
type AddNameParams struct {
	Name interface{}
}
```

### Expectation:
```json
type AddNameParams struct {
	Name string
}
```

Type `citext` should be supported as a `String` 

### Relevant log output

_No response_

### Database schema

```sql
CREATE TABLE IF NOT EXISTS temp(id uuid NOT NULL PRIMARY KEY DEFAULT gen_random_uuid(), name citext NOT NULL)
```

### SQL queries

```sql
INSERT INTO (name) VALUES('name')
```

### Configuration

_No response_

### Playground URL

_No response_

macOS

PostgreSQL

Go

ryan-berger:
Fixes #1581 .

I just added `citext` to the switch so that it would resolve to `string`, although this might not be the best solution.

I looked around, and it seems that `hstore` has similar code and since `hstore` and `citext` both are extensions that live in `pg_catalog` this seems to be in the spirit of the previous solutions.


kyleconroy (commented):
Could you please add a test case? There are many examples inside internal/endtoend/testdata.

ryan-berger:
@kyleconroy I've added a few e2e tests, sorry it took me so long to get around to this. This should be good to go

colbee1:
Cannot use sqlc because of citext not supported as string. Is there a chance to see patch merged soon ?

eserdk:
@colbee1 I think for now you could just use the following (I actually use this override):
```yaml
version: "2"
sql:
  - engine: "postgresql"
    schema: ...
    queries: ...
    gen:
      go:
        out: ...
        overrides:
          - db_type: "citext"
            go_type: "string"
```

Issue #1574, opened 2022-04-27
Min/Max on Datetimes result in interface{} as type


Neokil:
### Version

Other

### What happened?

Hint: I am using v1.13.0 (installed via brew) but that is not available in the form.

I currently have a query that looks like this:
```
SELECT COUNT(*) as NumOfActivities,
        MIN(event_time) as MinDate, 
        MAX(event_time) as MaxDate 
FROM activities 
WHERE account_id = $1
```
which creates a struct like this:
```
type GetActivityMetaDataForAccountRow struct {
	Numofactivities int64
	Mindate         interface{}
	Maxdate         interface{}
}
```
but I would expect
```
type GetActivityMetaDataForAccountRow struct {
	Numofactivities int64
	Mindate         time.Time
	Maxdate         time.Time
}
```
so it looks like after putting a date through an aggregate-function sqlc can no longer determine its type? but it seems only to be the case for datetimes as far as I can see it.

### Database schema

```sql
CREATE TABLE activities (
    account_id BIGINT NOT NULL,
    event_time TIMESTAMP WITH TIME ZONE NOT NULL, 
);
```

### SQL queries

```sql
SELECT COUNT(*) as NumOfActivities,
        MIN(event_time) as MinDate, 
        MAX(event_time) as MaxDate 
FROM activities 
WHERE account_id = $1
```

### What operating system are you using?

macOS

### What database engines are you using?

PostgreSQL

### What type of code are you generating?

Go

jose-zenledger:
Yeah I am experiencing this too. So then I changed it to cast the `min` or `max` so like:
 `MIN(event_time)::timestamp as MinDate`. However it fails when it returns null.

I then thought maybe I could use nulltime as a custom type but and I can't figure out how to use a custom type for custom queries like this.

kdubovikov:
This problem reproduces not only for timestamp fields, but for int fields as well.

```
with t1 as (
	select
		id,
		min(col) as min_col
	from
		table1
	group by
		id
)
select
	t1.min_col
from
	t2
left outer join t1 on t2.id = t1.id
```

May be related to https://github.com/kyleconroy/sqlc/issues/1334

kdubovikov:
[...]

PR #1631, opened 2022-05-19
fix: #1574 nullable return types for groupby functions

kdubovikov:
This is a proposed fix for https://github.com/kyleconroy/sqlc/issues/1574. I am unsure if this change breaks any tests because I don't have a correctly setup postgres instance to generate `pg_catalog` correctly with all extensions. However, I have checked that this change fixes the bug.

### DIFF
--- a/internal/compiler/output_columns.go
+++ b/internal/compiler/output_columns.go
@@ -213,12 +213,26 @@ func outputColumns(qc *QueryCatalog, node ast.Node) ([]*Column, error) {
 			}
 			fun, err := qc.catalog.ResolveFuncCall(n)
 			if err == nil {
-				cols = append(cols, &Column{
-					Name:       name,
-					DataType:   dataType(fun.ReturnType),
-					NotNull:    !fun.ReturnTypeNullable,
-					IsFuncCall: true,
-				})
+				if len(fun.Args) > 0 {
+					ref := n.Args.Items[0].(*ast.ColumnRef)
+					columns, err := outputColumnRefs(res, tables, ref)
+					if err != nil {
+						return nil, err
+					}
+					cols = append(cols, &Column{
+						Name:       name,
+						DataType:   columns[0].DataType,
+						NotNull:    !fun.ReturnTypeNullable,
+						IsFuncCall: true,
+					})
+				} else {
+					cols = append(cols, &Column{
+						Name:       name,
+						DataType:   dataType(fun.ReturnType),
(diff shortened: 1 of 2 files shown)


kdubovikov:
I haven't found any docs that mention what postgres instance I should use to generate pg_catalog. This is needed to complete the pull request and fix tests. Could anyone direct me on where to look?

internal/tools/sqlc-pg-gen/main.go
```diff
@@ -23,7 +23,8 @@ SELECT p.proname as name,
   format_type(p.prorettype, NULL),
   array(select format_type(unnest(p.proargtypes), NULL)),
   p.proargnames,
-  p.proargnames[p.pronargs-p.pronargdefaults+1:p.pronargs]
+  p.proargnames[p.pronargs-p.pronargdefaults+1:p.pronargs],
+  CASE WHEN p.prokind = 'a' THEN TRUE ELSE FALSE END
```
ryan-berger, line 27: I mentioned this in the discussion we were having, but I thought it is worth mentioning here. 

We probably should do a `CASE WHEN p.prokind = 'a' AND p.name != 'COUNT ....'` since `COUNT` is an exception. 

Also, I don't think there needs to be a `CASE` since you should just be able to use `p.prokind = 'a' AND p.name != 'COUNT'` as a boolean expression

kyleconroy (changes requested):
sqlc-pg-gen has been updated since this pull request was open. Can you rebase on main and see if you're still seeing this issue?

internal/compiler/output_columns.go
```diff
@@ -213,12 +213,26 @@ func outputColumns(qc *QueryCatalog, node ast.Node) ([]*Column, error) {
...
-				})
+				if len(fun.Args) > 0 {
+					ref := n.Args.Items[0].(*ast.ColumnRef)
+					columns, err := outputColumnRefs(res, tables, ref)
+					if err != nil {
+						return nil, err
+					}
+					cols = append(cols, &Column{
+						Name:       name,
+						DataType:   columns[0].DataType,
```
skabbes, line 224: I think that unfortunately the fix is a lot more complex than this, this is saying that any function with more than 0 args the return type should the type of its first arg?

This is true for most aggregates, (min, max, avg, every) but is not generally true of any function call - in particular `count`, `array_agg` and probably any user-defined aggregate.

https://www.postgresql.org/docs/15/functions-aggregate.html

I think special cases for these common aggregates is probably a useful change.

PR #1633, opened 2022-05-19
Validate sqlc function arguments

### LINKED ISSUE #1632: `sqlc.(n?)arg` panics when 0 arguments are passed in
### Version

1.13.0

### What happened?

I accidentally passed 0 arguments into `sqlc.narg` and received the following panic:

```
goroutine 1 [running]:
github.com/kyleconroy/sqlc/internal/sql/validate.ParamStyle({...., ......})
```

Which comes from this line:

https://github.com/kyleconroy/sqlc/blob/c2dcd56a5d8b2a3b2fa9c7f6545d0b1d5555be29/internal/sql/validate/param_style.go#L19

Although this is invalid input, it would be nice to have an error message printed out.

### Relevant log output

_No response_

### Database schema

_No response_

### SQL queries

_No response_

### Configuration

_No response_

### Playground URL

_No response_

Linux

PostgreSQL

Go

ryan-berger:
Fix #1632

### DIFF
--- a/internal/endtoend/testdata/sqlc_arg_invalid/mysql/query.sql
+++ b/internal/endtoend/testdata/sqlc_arg_invalid/mysql/query.sql
@@ -9,6 +9,9 @@ select id, first_name from users where id = sqlc.argh(target_id);
 -- name: TooManyArgs :one
 select id, first_name from users where id = sqlc.arg('foo', 'bar');
 
+-- name: TooFewArgs :one
+select id, first_name from users where id = sqlc.arg();
+
 -- name: InvalidArgFunc :one
 select id, first_name from users where id = sqlc.arg(sqlc.arg(target_id));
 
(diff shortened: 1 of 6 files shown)


kyleconroy:
Looks like there's a test failure. Sorry that it took a few days to run.

ryan-berger:
Oops, thanks for catching that. I forgot to rerun tests.

I took a look and it seems a bit of an interesting problem.

The error message that the test expects comes from the Visitor which
returns an error when it visits an invalid function.

If I return the error in there, it will crash because the ParamStyle
function is called before the Visitor is invoked


If I keep it how it is, tests fail and I have to duplicate logic.

I'll take a look at how to deal with this tomorrow. If you have any
preferences, let me know and I'll make sure to implement it that way

kyleconroy:
It looks like we already validate the arguments at a later spot. How about we just return early if there are no args to avoid the panic?

PR #1641, opened 2022-05-24
Add type support for pgtype.Point

samverrall-er:
This PR adds support for the `pgtype.Point` type to postgresql_type.go.


kyleconroy (commented):
We'll be adding support for all pgtype types when pgx/v5 support lands. 

Since that will be done as one large PR, I'm going to close this now.

Issue #1648, opened 2022-05-28
Tuple comparisons generate invalid code for the MySQL engine


ryanpbrewster:
### Version

1.13.0

### What happened?

A query like
```
SELECT * FROM foo WHERE (a, b) < (?, ?)
```
is valid in MySQL, but sqlc is generating the incorrect code for it. In particular, the query should accept two parameters, but sqlc generates:
```
func (q *Queries) GetFoos(ctx context.Context) ([]Foo, error) {
```

The correct code is generated for the Postgresql engine.

I believe that the issue is that `convertRowExpr` is unimplemented [here](https://github.com/kyleconroy/sqlc/blob/b0d6f13646d52287a6eb9be623f827227bfd6c7b/internal/engine/dolphin/convert.go#L1119).

### Relevant log output

```shell
$ SQLCDEBUG=1 sqlc generate
2022/05/28 13:23:17 dolphin.convert: Unknown node type *ast.RowExpr
2022/05/28 13:23:17 dolphin.convert: Unknown node type *ast.RowExpr
2022/05/28 13:23:17 dolphin.convert: Unknown node type *ast.RowExpr
2022/05/28 13:23:17 dolphin.convert: Unknown node type *ast.RowExpr
2022/05/28 13:23:17 dolphin.convert: Unknown node type *ast.RowExpr
2022/05/28 13:23:17 dolphin.convert: Unknown node type *ast.RowExpr
```

### Database schema

```sql
CREATE TABLE foo (a INT NOT NULL, b INT NOT NULL)
```

### SQL queries

```sql
-- name: GetFoos :one
SELECT 1 FROM foo WHERE (a, b) < (?, ?);
```

### Configuration

```yaml
{
  "version": "1",
  "packages": [
    {
      "path": "db",
      "engine": "mysql",
      "schema": "query.sql",
      "queries": "query.sql"
    }
  ]
}
```

### Playground URL

https://play.sqlc.dev/p/c3ac4a18aa93305c6fe39ed29e80ec9a9b3060a7d11c2b52fe79ed20224e343a

### What operating system are you using?

Linux

### What database engines are you using?

MySQL

### What type of code are you generating?

Go

ryanpbrewster:
@kyleconroy I have a fix for this here: https://github.com/kyleconroy/sqlc/pull/1649

Are you amenable to incremental approaches like that? If so I think I can very quickly close a bunch of related MySQL issues.

PR #1649, opened 2022-05-28
Add barebones support for converting mysql rowexprs

ryanpbrewster:
This is a _very_ barebones implementation of support for converting mysql row expressions into the AST.

Fixes https://github.com/kyleconroy/sqlc/issues/1648

~I could not find any obvious place to add a test for this. Should I be adding an example of tuple comparisons to the `examples/` directory? Or would you like me to add some basic unit-tests?~

EDIT: I went ahead and added a basic unit test that will catch regressions specifically around how many ParamRefs we extract.


ryanpbrewster:
@kyleconroy I'd like to merge or close this PR. If there is a way for me to structure these contributions that is easier to review/accept, I'm happy to revise. If there is not, I don't want to clog up your PR list.

Issue #1494, opened 2022-03-18
Postgresql 2D array in database schema generated as 1D array in models.go


colli173:
### Version

1.12.0

### What happened?

go version: 1.18 (equal results with 1.17.8)

Using TEXT[][] as the column type in postgresql generates a corresponding model with field of type []string, when expected model should contain field of type [][]string. Apologies if this is a known limitation of sqlc.

### Database schema

```sql
CREATE TABLE authors (
  id   BIGSERIAL PRIMARY KEY,
  names text[][]      NOT NULL,
  bio  text
);
```

### Configuration

```yaml
{
  "version": "1",
  "packages": [
    {
      "path": "db",
      "engine": "postgresql",
      "schema": "query.sql",
      "queries": "query.sql",
      "sql_package": "pgx/v4"
    }
  ]
}
```

### Playground URL

https://play.sqlc.dev/p/c29f1f2e8507033785a2753a7526650bb2c593370d66c5eed44cbc07e5aca2ea

### What operating system are you using?

macOS

### What database engines are you using?

PostgreSQL

### What type of code are you generating?

Go

PR #1651, opened 2022-05-30
feat: add support for Postgres multi dimensional arrays in Go

colli173:
This PR adds support for generating model fields from multi-dimensional postgresql array columns. The approach here is to pass around an additional column attribute `array_bounds` representing the dimensions of the postgres array column. When generating code, the `array_bounds` is checked along `is_array` to determine the dimensions of the resulting struct field. Would be happy to hear alternative approaches or solutions here as well. More info: [Postgres Arrays](https://www.postgresql.org/docs/14/arrays.html)


kyleconroy (dismissed):
This looks like a solid improvement over our current multi-dimension support. The code generation looks correct, but I do want to verify that this works against a running PostgreSQL instance before merging.

Can you fix the merge conflicts? Happy to merge after that.

colli173:
> We should only change the multi-dimension support for lib/pq.

Thanks for taking a look. Could you clarify the ask here? Is there an aspect of this that should change for the lib/pq package?

colli173:
@kyleconroy soft ping for another review when you have a chance.

Issue #1663, opened 2022-06-07
`SELECT COALESCE()` results are always assumed non-null


mpyw:
### Version

1.13.0

### What happened?

Possibly related to:

- #937
- #1096

`SELECT COALESCE(x, y)` results are always assumed non-null.
If all fields are nullable, result typing should not be `string` but `NullString`.

### Database schema

```sql
CREATE TABLE examples(
  id BIGSERIAL PRIMARY KEY,
  nullable text,
  nonnull text NOT NULL
);
```

### SQL queries

```sql
-- name: GetExample :one
SELECT
  nullable,
  nonnull,
  COALESCE(nonnull) AS coalesce_nonnull,
  COALESCE(nullable) AS coalesce_nullable,
  COALESCE(nonnull, nonnull) AS coalesce_nonnull_nonnull,
  COALESCE(nonnull, nullable) AS coalesce_nonnull_nullable,
  COALESCE(nullable, nonnull) AS coalesce_nullable_nonnull,
  COALESCE(nullable, nullable) AS coalesce_nullable_nullable
FROM examples WHERE id = $1 LIMIT 1;
```

### Configuration

```json
{
  "version": "1",
  "packages": [
    {
      "path": "db",
      "engine": "postgresql",
      "schema": "query.sql",
      "queries": "query.sql"
    }
  ]
}
```

#### Expected:

```go
type GetExampleRow struct {
	Nullable                 sql.NullString
	Nonnull                  string
	CoalesceNonnull          string
	CoalesceNullable         sql.NullString
	CoalesceNonnullNonnull   string
	CoalesceNonnullNullable  string
	CoalesceNullableNonnull  string
	CoalesceNullableNullable sql.NullString
}
```

#### Actual:

```go
type GetExampleRow struct {
	Nullable                 sql.NullString
	Nonnull                  string
	CoalesceNonnull          string
	CoalesceNullable         string
	CoalesceNonnullNonnull   string
	CoalesceNonnullNullable  string
	CoalesceNullableNonnull  string
	CoalesceNullableNullable string
}
```

### Playground URL

https://play.sqlc.dev/p/c9bba8ca287f65b7ce5f282bec5da570b0bd7693223e1d61e8b6619f4ba92a42

### What operating system are you using?

macOS

### What database engines are you using?

PostgreSQL

### What type of code are you generating?

Go

mpyw:
[...]

PR #1664, opened 2022-06-07
fix: 🐛 Correctly switch `coalesce()` result `.NotNull` value

mpyw:
This change follows the previous implementation that mutates and returns the first column, but switches only the `.NotNull` value depending on the situation.

- Is it sufficient that asserting only types of columns and constants?
- Also I'm concerned that the `A_Const` node may contain `Null`.

Also, I am new to Golang OSS, so I may be making some large mistakes. Thank you in advance for your review.

BREAKING CHANGE: 🧨 `coalesce()` may assign nullable results

✅ Closes: #1663

### DIFF
--- a/internal/compiler/output_columns.go
+++ b/internal/compiler/output_columns.go
@@ -148,9 +148,11 @@ func outputColumns(qc *QueryCatalog, node ast.Node) ([]*Column, error) {
 			if res.Name != nil {
 				name = *res.Name
 			}
-			var found bool
+			var firstColumn *Column
+			var shouldNotBeNull bool
 			for _, arg := range n.Args.Items {
-				if found {
+				if _, ok := arg.(*ast.A_Const); ok {
+					shouldNotBeNull = true
 					continue
 				}
 				if ref, ok := arg.(*ast.ColumnRef); ok {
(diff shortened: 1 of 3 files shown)


internal/compiler/output_columns.go
```diff
@@ -159,14 +161,18 @@ func outputColumns(qc *QueryCatalog, node ast.Node) ([]*Column, error) {
...
+						shouldNotBeNull = shouldNotBeNull || c.NotNull
 					}
 				}
 			}
-			if !found {
+			if firstColumn != nil {
+				firstColumn.NotNull = shouldNotBeNull
+				firstColumn.skipTableRequiredCheck = true
+				cols = append(cols, firstColumn)
+			} else {
```
mpyw, line 175: Question: Is this really correct that using first column on this way?

internal/compiler/output_columns.go
```diff
@@ -148,9 +148,11 @@ func outputColumns(qc *QueryCatalog, node ast.Node) ([]*Column, error) {
...
 			}
-			var found bool
+			var firstColumn *Column
+			var shouldNotBeNull bool
 			for _, arg := range n.Args.Items {
-				if found {
+				if _, ok := arg.(*ast.A_Const); ok {
+					shouldNotBeNull = true
 					continue
 				}
```
mpyw, line 157: Question: Is it sufficient that asserting only `A_Const` and `ColumnRef`?

internal/compiler/output_columns.go
```diff
@@ -159,14 +161,18 @@ func outputColumns(qc *QueryCatalog, node ast.Node) ([]*Column, error) {
...
 					}
 					for _, c := range columns {
-						found = true
-						c.NotNull = true
-						c.skipTableRequiredCheck = true
-						cols = append(cols, c)
+						if firstColumn == nil {
+							firstColumn = c
+						}
+						shouldNotBeNull = shouldNotBeNull || c.NotNull
```
mpyw, line 167: Question: What are the differences between `n.Args.Items` and their inner `columns`?

kyleconroy (approved):
Thanks for the PR! Sorry that it's take so long to review. Took me a few times through to understand the changes, but I think it's the right change and actually quite backwards compatible.

PR #1686, opened 2022-06-17
SQLite SELECT parameters

russellhaering:
Experimenting with SQLite and this query wasn't working as intended:

```sql
/* name: GetUser :one */
SELECT * FROM users WHERE id = ? LIMIT 1;
```

This PR adds some code to pass `WHERE` clauses on `SELECT` statements through to the internal AST, which seems to resolve the issue - although is no means comprehensive (for example `sqlc.arg()` still does not work).


russellhaering:
Closing this out. #1687 generates identical code, and handles many more use-cases.

PR #1687, opened 2022-06-17
feat: More SQL Syntax Support for SQLite

hakobera:
This pull request includes multiple SQL Syntax supports for SQLite for #161 

## Features

* `SELECT`
  * Support `WHERE`, `GROUP BY`, `HAVING`, `LIMIT`, `OFFSET`
  * Support nested `SELECT` like `select_nested_count`
* `INSERT`
  * Support both values and `SELECT` insert
* `UPDATE`
  * Support simple update set
* DDL
  * Support `CREATE VIEW` and `DROP VIEW`
  * Support `bool` type
* Function
  * `coalesce` function

## Examples

* authors
* booktest
* ondeck

Also add `endtoend` test data as much as possible.

## Breaking Change

* `numeric` and `decimal` are mapped to `float64` (currently these are mapped to `int64`


russellhaering:
This looks awesome, let me test a little bit but I'm definitely inclined to close out  #1686 in favor of this.

DBarney:
Wow, yeah lets go with this. I just got the Update statement working on my local branch after some work this morning, but this covers all that and more. So nice work @hakobera!

russellhaering:
Ran this against my (still pretty minimal) codebase and results look great. Its generating identical code for SELECT compared to #1686, identical code for INSERT compared to #1653, plus is generating correct-looking code for UPDATEs which I didn't have a solution for yet.

Closed my PR in favor of this.

javiercbk:
I executed some tests over a small project and it works very good.

Unrelated to this PR, but I will submit another PR fixing the following issue:

PR #1712, opened 2022-06-26
Support sqlite stdlib functions

hakobera:
Add SQLite function catalogs listed in below pages.

* https://www.sqlite.org/lang_aggfunc.html
* https://www.sqlite.org/lang_mathfunc.html
* https://www.sqlite.org/lang_corefunc.html

## Task

- [x] Move existing `builtins` end to end test to `builtins/postgresql`
- [x] Add function catalogs for SQLite
  - [x] Aggregate functions
  - [x] Mathematical functions
  - [x] Scalar functions
- [x] Add end to end test

## Remarks

Due to current SQLite parser limitation, we can't get correct type of arguments of function, so we can't overload function based on argument type like below.

```sql
SELECT max(1); => integer
SELECT max(1.0); => real
SELECT max('text'); => text
```

So arguments and return types of these functions like `max` are defined as `any` type due to this limitation.


internal/engine/sqlite/stdlib.go
```diff
@@ -5,18 +5,29 @@ import (
...
+				},
+			},
+			ReturnType:         &ast.TypeName{Name: "real"},
+			ReturnTypeNullable: true,
+		},
 		{
 			Name:       "COUNT",
 			Args:       []*catalog.Argument{},
-			ReturnType: &ast.TypeName{Name: "bigint"},
+			ReturnType: &ast.TypeName{Name: "integer"},
```
kyleconroy, line 30: This will change quite a bit of generated code. Is it documented that SQLite will always return an int here?

hakobera: In SQLite world, `bigint` and `integer` is totally same.

https://www.sqlite.org/datatype3.html

PR #1729, opened 2022-07-07
fix(postgresql): Add quotes for CamelCase columns

### LINKED ISSUE #1718: No quotes around column names in Postgres
### Version

1.14.0

### What happened?

SQLC is not adding quotes around columns in postgresql.

Query:
```sql
SELECT * FROM "Article";
```

Generated Query
```sql
SELECT id, title, publishDate FROM "Article";
```

No `""` quotes around `publishDate` since it is camelCase 🤷‍♂️.

How can I get sqlc to generate query like this:

```sql
SELECT "id", "title", "publishDate" FROM "Article";
```

### Relevant log output

_No response_

### Database schema

_No response_

### SQL queries

```sql
SELECT * FROM "Article";
```

### Configuration

```yaml
version: 2
sql:
- engine: "postgresql"
  schema: "./pkg/db/migrations"
  queries: "./pkg/db/queries.sql"
  gen:
    go:
      package: "dbgen"
      out: "./pkg/db/dbgen"
      json_tags_case_style: camelCase
      emit_db_tags: true
```

_No response_

### Playground URL

_No response_

macOS

PostgreSQL

Go

akutschera:
If columns are created with CamelCase, they have to be quoted or else
postgres won't find them.

This PR only applies to ```SELECT * FROM tablename``` queries. When the column names are specified individually with quotes, it already works.

fixes #1718

### DIFF
--- a/internal/compiler/expand.go
+++ b/internal/compiler/expand.go
@@ -45,6 +45,12 @@ func (c *Compiler) quoteIdent(ident string) string {
 			return "\"" + ident + "\""
 		}
 	}
+	if c.conf.Engine == config.EnginePostgreSQL {
+		// camelCase means the column is also camelCase
+		if strings.ToLower(ident) != ident {
+			return "\"" + ident + "\""
+		}
+	}
 	return ident
 }
 
--- a/internal/compiler/compile_test.go
+++ b/internal/compiler/compile_test.go
@@ -1,0 +1,38 @@
+package compiler
+
+import (
+	"testing"
+
+	"github.com/kyleconroy/sqlc/internal/config"
+)
+
+func TestQuoteIdent(t *testing.T) {
+	type test struct {
+		engine config.Engine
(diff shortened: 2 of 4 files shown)


internal/compiler/expand.go
```diff
@@ -45,6 +45,12 @@ func (c *Compiler) quoteIdent(ident string) string {
 			return "\"" + ident + "\""
 		}
 	}
+	if c.conf.Engine == config.EnginePostgreSQL {
+		// camelCase means the column is also camelCase
+		if strings.ToLower(ident) != ident {
+			return "\"" + ident + "\""
```
gurleensethi, line 51: 🎉 This will solve the problem where columns are in camel case and should not break anything.

gurleensethi:
@kyleconroy Can we release a new version of SQLC that contains this functionality?

Issue #1724, opened 2022-07-05
Add support for RETURNING


Arsen6331:
### Version

Other

### What happened?

I've been trying out the new beta sqlite support using the main branch of sqlc, and it seems to work perfectly for the most part, but the `RETURNING` statement doesn't work.

### Relevant log output

```shell
extraneous input 'RETURNING' expecting {<EOF>, ';', ALTER_, ANALYZE_, ATTACH_, BEGIN_, COMMIT_, CREATE_, DEFAULT_, DELETE_, DETACH_, DROP_, END_, EXPLAIN_, INSERT_, PRAGMA_, REINDEX_, RELEASE_, REPLACE_, ROLLBACK_, SAVEPOINT_, SELECT_, UPDATE_, VACUUM_, VALUES_, WITH_}
```

### Database schema

```sql
CREATE TABLE users (name text, id serial primary key);
```

### SQL queries

```sql
-- name: InsertUser :one
INSERT INTO users (name) VALUES (?) RETURNING id;

-- name: UpdateUser :one
UPDATE users SET name = ?
  RETURNING id;

-- name: DeleteUser :one
DELETE FROM users
  WHERE name = ?
  RETURNING id;
```

### What operating system are you using?

Linux

### What type of code are you generating?

Go

PR #1741, opened 2022-07-11
feat(sqlite) add support for returning

hakobera:
Implement #1724 .
Support both `RETURNING <column, column, ...>` and `RETURNING *`.

- [x] add [returning-clause](https://www.sqlite.org/syntax/returning-clause.html) support for SQLite parser
  - [x] Insert
  - [x] Update
  - [x] Delete
- [x] add end to end test for SQLite
- [x] update PostgreSQL end to end test to verify `RETURNING *` support

@kyleconroy Do you think we should update examples to use `RETURNING`?


kyleconroy (approved):

kyleconroy:
Let's update the examples in a separate PR

PR #1751, opened 2022-07-20
Add option to produce private structs for query params

kazhuravlev:
We use sqlc to generate some code, and on top of that, we write others. 

But unfortunately, generated code and structs are available by other packages. Also, different packages can see `New` and `Queries`. So in the next PR, I would like to add corresponding options.


kyleconroy:
If the query structs are private, how are you calling the methods?

kazhuravlev:
Good point. 

I try to explain what I am expecting.

```
query.sql
db.gen.go
models.gen.go
queries.gen.go

public_api_of_my_service.go
```

I want to generate queries by `sqlc` to some packages, like `storage`. Also, this package will contain some hand-written code (`public_api_of_my_service.go`). When my clients consume `storage` functionality - they will use only methods in `public_api_of_my_service.go`. 

So, they call some hand-write public method (like `GetSomeObjectByID()`), and this method will use code-generated code produced by `sqlc`.

kazhuravlev:
@kyleconroy, do you have any comments on that?

kyleconroy:
By using an `internal` package you can get the same level of encapsulation. Take a look at the attached example, but if you put an internal package inside your storage package, only the storage package can access it. 

```
example
├── go.mod
├── query.sql
├── sqlc.json
├── storage
│   ├── internal
│   │   └── db
│   │       ├── db.go
│   │       ├── models.go
│   │       └── query.sql.go
│   └── service.go
└── web
    └── main.go
```

[example.zip](https://github.com/kyleconroy/sqlc/files/9454178/example.zip)

kyleconroy (commented):
After looking over the code, I'm convinced that using the `internal` package is a cleaner way to solve this.

Issue #1208, opened 2021-09-21
Subselected value should be nullable


aarondl0:
### Version

1.10.0

### What happened?

```
-- name: ShouldHaveNull :many
select
   (select other_table.name from other_table where this_table.name = other_table.name) as "name"
from this_table;
```

sqlc will generate an output struct like this:

```
type ShouldHaveNullRow struct {
  Name string
}
```

But the struct should be:
```
type ShouldHaveNullRow struct {
  Name sql.NullString
}
```

The above example is with strings but I thought it was relevant to mention that this is happening with a replaced type that's configured similarly to:

```yaml
  - db_type: "pg_catalog.numeric"
    nullable: false
    go_type:
      import: "github.com/shopspring/decimal"
      type: "Decimal"
      pointer: false
  - db_type: "pg_catalog.numeric"
    nullable: true
    go_type:
      import: "github.com/shopspring/decimal"
      type: "NullDecimal"
      pointer: false
```

I know the types in this example are all messed up. I'm in a huge rush and don't have time to file the issue with extreme precision at this moment.

You can see in the playground link because the type is non nullable in the table, it incorrectly assumes that the subselect can never return null for that field.

### Playground URL

https://play.sqlc.dev/p/bb47e4143c86e1b45bc659b147c4072d0464d968332d8598e36f7c2e3e531f18

### What operating system are you using?

Linux

### What database engines are you using?

PostgreSQL

### What type of code are you generating?

Go

aarondl0:
I think you may be misunderstanding the complaint.

Let me explain more clearly (I have a bit more time now). There is a subselect query happening where we're retrieving a single value from it:

```sql
(select o.name from one o where one.name = 5 limit 1) as "name"
```

If you look at the struct it created to capture the result of `Test` query it is:
```go
type TestRow struct {
	Otherfield sql.NullString
	Name       decimal.Decimal
}
```

However it should be:
```go
type TestRow struct {
	Otherfield sql.NullString
	Name       decimal.NullDecimal // note the difference here
}
```

The first struct (the one sqlc generates and is in my playground link) is incorrect. This valid query cannot properly be bound to this struct for all possible return values. Does that make it more clear?

Your example is working I agree. However I could only reproduce when using type overrides which your example does not include. Please take another look.

voluntas:
I am also having the same problem. Even if I reflect the type in sqlc.yml in the subquery results, it is not applied. Could you please check this Issue again?

[...]

PR #1758, opened 2022-07-25
Add not null to expression sublink

WesleyMiller1998:
A simple sublink change to resolve a type generation issue.

THIS PR IS NOT THOROUGHLY TESTED AND SHOULD BE REVIEWED WITH A HIGH LEVEL OF SCRUTINY. WHILE IT IS A VERY SMALL CHANGE, I HAVE NO CONSIDERED THE IMPACT OF THIS CHANGE WHATSOEVER.

### DIFF
--- a/internal/compiler/output_columns.go
+++ b/internal/compiler/output_columns.go
@@ -241,6 +241,7 @@ func outputColumns(qc *QueryCatalog, node ast.Node) ([]*Column, error) {
 					return nil, err
 				}
 				first := subcols[0]
+				first.NotNull = false
 				if res.Name != nil {
 					first.Name = *res.Name
 				}


kyleconroy:
> THIS PR IS NOT THOROUGHLY TESTED AND SHOULD BE REVIEWED WITH A HIGH LEVEL OF SCRUTINY. WHILE IT IS A VERY SMALL CHANGE, I HAVE NO CONSIDERED THE IMPACT OF THIS CHANGE WHATSOEVER.

UNDERSTOOD THANK YOU VERY MUCH FOR YOUR CONTRIBUTION AND I WILL BE VERY CAREFUL MERGING IT IN I HOPE YOU AREN'T OFFENDED IF I DON'T MERGE IT BECAUSE IT MAY BREAK SOME THINGS

kyleconroy:
The tests failed which means this change isn't likely to be safe. Can you provide a bit more context on why you needed to do this?

WesleyMiller1998:
It had to do with issue #1208. I saw you had dismissed the issue as invalid, but I found that the issue was in fact valid. It has to do with how a subquery is processed. As pointed out in the issue, the subquery could return null which would mean that the return value of that subquery should also be null. I will have to check the tests to see where it went wrong and update

WesleyMiller1998:
Okay so I looked at the failed test. The issue is that when we do a subquery with SELECT COUNT, it is not a nullable value but if we do a subquery of (SELECT * FROM x WHERE id = 1 LIMIT 1) AS "name", like in the example, this query is certainly nullable

WesleyMiller1998:
Okay, so I dug into this problem a little further. Firstly, I was using the term "subquery" incorrectly. A subquery is a legitimate Postgresql term, see [here](https://www.postgresql.org/docs/9.0/functions-subquery.html), that returns a boolean value (meaning that it will always return a NOT NULL value. The other possibility, which I was again confusing the terminology, is an [aggregate function](https://www.postgresql.org/docs/9.0/functions-aggregate.html). This is what is happening in the test. Aggregate functions also return NOT NULL values. The problem highlighted in issue #1208 is that they are actually making an entirely new query, that is not a subquery or aggregate, which MUST be nullable despite the table definition.

WesleyMiller1998:
Apologies for my word vomit. I'm realizing now that the SubLink likely does handle most of those cases correctly because it is just a recursive call. The only case that it misses is when it recursively calls a plain SELECT x.id FROM x query because this will take the nullable value of the field id, but it should be always be nullable despite the value of the field

kyleconroy:
Since this PR doesn't have any tests, I'm going to go ahead and close it. @WesleyMiller1998 if you're still running into an issue, please open up a bug report with a complete reproduction of the issue. We can then decide on the fix there.

PR #1772, opened 2022-08-03
Add nowasm tag, fix nowasm inconsistency

SN9NV:
I found an inconsistency between the implementation of the wasm and nowasm files.

I thought that since it seems that this feature can be disabled, it would be nice to be able to control it from a tag. So I added that too.

The reason I came about this in the first place was that it seems that there is a Rust dependency in the wasm implementation, and that conflicts with my build system. I'm sure I could figure it out if this will not be possible in the future.

### DIFF
--- a/internal/ext/wasm/nowasm.go
+++ b/internal/ext/wasm/nowasm.go
@@ -1,16 +1,19 @@
-//go:build !(cgo && ((linux && amd64) || (linux && arm64) || (darwin && amd64) || (darwin && arm64) || (windows && amd64)))
+//go:build nowasm || !(cgo && ((linux && amd64) || (linux && arm64) || (darwin && amd64) || (darwin && arm64) || (windows && amd64)))
 
 package wasm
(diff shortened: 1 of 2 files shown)


kyleconroy:
> that it seems that there is a Rust dependency in the wasm implementation

I'm using wasmtime which is written in Rust. It comes with pre-built libraries though, so you shouldn't need a Rust toolchain to install sqlc. Can you give me a bit more information about your setup?

SN9NV:
Thanks for merging my PR.

My setup involves some restricted environments where glibc is not available, as well as developers machines running various distributions of Linux, Mac, ... so we decided to have a toolbox of our internally used tools, which is statically compiled where possible. For sqlc, we are using Zig as a C compiler, and it seems to be working for us.

Please don't feel too pressured to support such a specific use-case. We are used to putting in a little effort maintaining a fork or patch set where necessary to keep this system going.

PR #1775, opened 2022-08-05
docs: Update transactions how to example

jasonblanchard:
The example for transactions is a little confusing for a few reasons:

1. It mentions `WithTx`, but then it never actually calls it
2. Confusing use of the `db` package - I believe `db.Begin(ctx)` should be `p.Begin(ctx)`

My suggestion here does a few things: 
1. Fixes the above issues.
2. Uses naming conventions that match the “Getting started with PostgreSQL” tutorial (i.e `package tutorial`, `queries`, `schema.sql`, `query.sql`, `db`, etc).
3. Use `database/sql` and `lib/pq` to match the “Getting started with PostgreSQL” tutorial.


kyleconroy (approved):
Thanks for the improvements!

PR #1794, opened 2022-08-11
fix: prevent batch infinite loop with arg length

### LINKED ISSUE #1678: Errors in Batch cause infinite loop
### Version

1.14.0

### What happened?

If a query in the new `:batchexec` causes an error, batch.Exec runs in an infinite loop.
I think this is caused by this (generated code):
```
func (b *MyBatchResults) Exec(f func(int, error)) {
	for {
		_, err := b.br.Exec()
		if err != nil && (err.Error() == "no result" || err.Error() == "batch already closed") {
			break
		}
		if f != nil {
			f(b.ind, err)
		}
		b.ind++
	}
}
```

it only exits if no result/closed. if there is some error, but a "real error" that is not one of these two, the for loop will keep running forever.
a workaround i found is calling Close() on the batch BEFORE running exec. But i am not convinced this is the right way; pgx' tests do run Close only after the right amount of Exec() calls have been made : https://github.com/jackc/pgx/blob/master/batch_test.go

Possible fix could be: when creating the batch, the generated code saves the number of items in the batch into the batch struct. When running exec, the for loop exits if enough calls have been made.

Any opinions? I could probably contribute a fix, if it can be confirmed that my assumptions here are correct.

### Relevant log output

```shell
1103466 ERROR: duplicate key value violates unique constraint "<redacted>" (SQLSTATE 23505)
```
```


### Database schema

_No response_

### SQL queries

_No response_

### Configuration

_No response_

### Playground URL

_No response_

_No response_

_No response_

_No response_

### PR DESCRIPTION
**Pull request by chrisyxlee**
Because the batch queries are only being closed when checking for 2 errors (batch results closed or no result), other errors cause the generated code to infinite loop. This changes from a for loop that only breaks on checking the errors to a for loop that also considers the the total number of params passed in.

Some errors will not infinite loop if you close the results yourself. However, there are other errors that you can't close --

* Context cancelled during a batch query
* Connection is busy
* Bad query

Bad queries typically happen in tests, but the infinite loop has eaten up >20Gb of RAM within a few minutes for me before.

This is usually what happens. It infinite loops, so I can't interrupt the process, and then I have to do `ps aux | grep <>` to find and kill the process manually:

Screen Shot 2022-08-15 at 15 05 18

Fixes #1678 (which was closed by the reporter since they found a workaround)

### DIFF
--- a/examples/batch/postgresql/batch.go
+++ b/examples/batch/postgresql/batch.go
@@ -7,6 +7,7 @@ package batch
 
 import (
 	"context"
+	"errors"
 	"time"
 
 	"github.com/jackc/pgx/v4"
@@ -18,9 +19,9 @@ WHERE year = $1
 `
 
 type BooksByYearBatchResults struct {
-	br  pgx.BatchResults
-	ind int
-	tot int
+	br     pgx.BatchResults
+	tot    int
+	closed bool
 }
 
 func (q *Queries) BooksByYear(ctx context.Context, year []int32) *BooksByYearBatchResults {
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

**Comment by chrisyxlee** [comment]
🤔 This is the test failure -- 
At first glance, it seems like the DB connection was overloaded, but not sure if it's a symptom of something that I need to fix.

```
--- FAIL: TestBatchBooks (0.02s)
    db_test.go:15: db: ***localhost:5432/postgres?sslmode=disable
    db_test.go:94: error updating book 2: conn busy
FAIL
FAIL	github.com/kyleconroy/sqlc/examples/batch/postgresql	0.029s
```

**Review by kyleconroy — CHANGES_REQUESTED** [review]
Looking over the generated code, I think there's a cleaner way to implement the callbacks without having the check the return value of the error.

```go
func (b *UpdateValuesBatchResults) Exec(f func(int, error)) {
	defer b.br.Close() // Close is safe to call multiple times
	for i := 0; i < b.tot; i++ {
		if b.closed {
			f(i, errrors.New("batch already closed"))
			continue
		}
		_, err := b.br.Exec()
		if f != nil {
			f(i, err)
		}
	}
}

func (b *UpdateValuesBatchResults) Close() error {
	return b.closed = true
	return b.br.Close()
}
```

Thoughts?

chrisyxlee:
Oh yeah that's super clean, thanks! I'll circle back around and request another review after I've updated everything.

chrisyxlee:
* I chose `t` for the index looping through the batch results. The `i` variable is already being used and is being specified here: https://github.com/kyleconroy/sqlc/blob/0a1b04eb47c80236db176805fe43f709042b805c/internal/codegen/golang/result.go#L242-L248
  Is there a way I can verify that this variable won't collide with other generated code?
* Added a lot more error checking for the user function `f`.
* I also noticed there was a `defer rows.Close()` within a loop, so I pulled that into an anonymous function to allow the deferral to happen ASAP.

kyleconroy (approved):

PR #1798, opened 2022-08-14
docs: Add a getting started guide for SQLite

utamori:
SQLite support is fire!

Add SQLite guide document created from postgresql guide


kyleconroy (commented):
Thanks

PR #1809, opened 2022-08-21
pg-gen: make sqlc-pg-gen the complete source of truth for pg_catalog.go

skabbes:
This is sort of a yak-shaving task for https://github.com/kyleconroy/sqlc/pull/1748

Problem: there seem to have been some manual edits to pg_catalog.go over the course of time.  This makes it hard (almost impossible actually) to do any change there with any confidence.

Solution: Do some detective work on the manual edits performed, and incorporate them into pg-gen.  Once that is done the work on #1748 will be a lot easier.

Here are the 2 manual edits as far as I can tell:
https://github.com/kyleconroy/sqlc/commit/95131550616d8d06c9a484e948e91a6a7ed7b3cc (pg_catalog functions with variadic signatures - this is completely solved by this PR)

And

https://github.com/kyleconroy/sqlc/commit/846c6298ca0a53ab0cf176d16b4da56c52345d3c (manual edits to pg_catalog that I _believe_ will be fixed "right way" once "direct args" vs "aggregated args" are separated in #1748 ).

Commits in order tell a story, probably best to review that way.

### DIFF
--- a/internal/endtoend/testdata/valid_group_by_reference/postgresql/go/query.sql.go
+++ b/internal/endtoend/testdata/valid_group_by_reference/postgresql/go/query.sql.go
@@ -79,7 +79,7 @@ ORDER BY bucket DESC
 type ListMetricsRow struct {
 	Bucket   interface{}
 	CityName sql.NullString
-	Avg      string
+	Avg      float64
 }
 
 func (q *Queries) ListMetrics(ctx context.Context) ([]ListMetricsRow, error) {
--- a/internal/engine/postgresql/contrib/pgcrypto.go
+++ b/internal/engine/postgresql/contrib/pgcrypto.go
@@ -92,7 +92,7 @@ func Pgcrypto() *catalog.Schema {
 			Name: "digest",
 			Args: []*catalog.Argument{
 				{
-					Type: &ast.TypeName{Name: "bytea"},
+					Type: &ast.TypeName{Name: "text"},
 				},
 				{
 					Type: &ast.TypeName{Name: "text"},
@@ -104,7 +104,7 @@ func Pgcrypto() *catalog.Schema {
 			Name: "digest",
 			Args: []*catalog.Argument{
 				{
-					Type: &ast.TypeName{Name: "text"},
+					Type: &ast.TypeName{Name: "bytea"},
 				},
 				{
 					Type: &ast.TypeName{Name: "text"},
--- a/internal/engine/postgresql/contrib/pgcrypto.go
+++ b/internal/engine/postgresql/contrib/pgcrypto.go
@@ -206,6 +206,15 @@ func Pgcrypto() *catalog.Schema {
 			ReturnType: &ast.TypeName{Name: "bytea"},
 		},
 		{
+			Name: "pgp_armor_headers",
+			Args: []*catalog.Argument{
+				{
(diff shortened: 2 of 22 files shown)


skabbes:
@kyleconroy I think given the fact that you have been pretty diligent about adding endtoend tests whenever you change something, I think as long as this passes the test, it should be safe.  But if you can remember any other manual edits to pg_catalog as well, we'll definitely need to fix them - this could _possibly_ be a breaking change if we don't catch one of those edits.

internal/endtoend/testdata/valid_group_by_reference/postgresql/go/query.sql.go
```diff
@@ -79,7 +79,7 @@ ORDER BY bucket DESC
 type ListMetricsRow struct {
 	Bucket   interface{}
 	CityName sql.NullString
-	Avg      string
+	Avg      float64
```
skabbes, line 82: The query is:

```
SELECT time_bucket('15 days', time) AS bucket, city_name, AVG(temp_c)
FROM weather_metrics
WHERE DATE_SUB(NOW(), INTERVAL 6 MONTH)
GROUP BY bucket, city_name
ORDER BY bucket DESC;
```

So I think the float64 is actually correct here - the string was incorrect before.  They are also "both wrong" since the order of the function definitions in `pg_catalog.go` is actually what is determining the types here.

internal/engine/postgresql/contrib/pgcrypto.go
```diff
@@ -104,7 +104,7 @@ func Pgcrypto() *catalog.Schema {
 			Name: "digest",
 			Args: []*catalog.Argument{
 				{
-					Type: &ast.TypeName{Name: "text"},
+					Type: &ast.TypeName{Name: "bytea"},
```
skabbes, line 107: Changed the order of these, as they affected the end-to-end tests, and just wanted to preserve legacy behavior.

internal/engine/postgresql/pg_catalog.go
```diff
@@ -4150,16 +4150,6 @@ func genPGCatalog() *catalog.Schema {
...
-		{
-			Name: "concat",
-			Args: []*catalog.Argument{
-				{
-					Type: &ast.TypeName{Name: "any"},
-					Mode: ast.FuncParamVariadic,
-				},
-			},
-			ReturnType: &ast.TypeName{Name: "text"},
-		},
```
skabbes: This concat is not in the pg_catalog on the main branch, so I removed it here too.

internal/engine/postgresql/pg_catalog.go
```diff
@@ -6892,6 +6882,18 @@ func genPGCatalog() *catalog.Schema {
...
+			Args: []*catalog.Argument{
+				{
+					Type: &ast.TypeName{Name: "numeric"},
+				},
+				{
+					Type: &ast.TypeName{Name: "numeric"},
+				},
+			},
+			ReturnType: &ast.TypeName{Name: "numeric"},
+		},
```
skabbes, line 6896: Simply changed the order of the functions here to preserve the behavior of the e2e tests.

kyleconroy (approved):
Wow, amazing work. I'm kicking myself for editing this file manually over the last year. Especially appreciate the work at preserving test compatibility.

Issue #1807, opened 2022-08-21
`strict_function_checks` invalidates `sqlc.arg` and `sqlc.narg`


verygoodsoftwarenotvirus:
### Version

1.15.0

### What happened?

First, I want to say I very much appreciate everybody's work on this really cool project that is working out very well for my use case. Thank you so much for your time and dedication.

I was trying to follow along with [the documentation for nullable named parameters](https://docs.sqlc.dev/en/v1.15.0/howto/named_parameters.html#nullable-parameters), and kept getting the error:

```: function "sqlc.narg" does not exist```

I encountered the same output when using a plain `sqlc.arg`.

I found [an earlier issue](https://github.com/kyleconroy/sqlc/issues/1667) filed before the feature was formally released, but that was before version 1.14, and I was using the latest docker tag, which I assumed was 1.15.0. I tried:
- explicitly using the 1.15.0 tag in the docker image
- explicitly using the 1.14.0 tag in the docker image (thinking maybe it was released and then revoked lol)
- installing sqlc 1.15.0 via `go get` and running that manually

Eventually what "fixed" it was removing `strict_function_checks: true` from my configuration. I feel like this is either a bug in the functionality of the flag or of the documentation for not calling it out. I'm happy to implement either fix as well, but I'm definitely not familiar enough with the project to assess which is which. 

### Configuration

```yaml
version: "2"
sql:
  - schema: "internal/database/postgres/migrations"
    queries: "internal/database/postgres/queries"
    engine: "postgresql"
    strict_function_checks: false  # <-- this was true, previously
    gen:
      go:
        package: "generated"
        sql_package: "database/sql"
        out: "internal/database/postgres/generated"
        output_files_suffix: ""
        emit_interface: true
        emit_exported_queries: true
        emit_empty_slices: false
        emit_params_struct_pointers: true
        emit_result_struct_pointers: true
        emit_exact_table_names: true
        emit_enum_valid_method: true
        emit_all_enum_values: true
```

### What operating system are you using?

Linux

### What database engines are you using?

PostgreSQL

### What type of code are you generating?

Go

skabbes:
@verygoodsoftwarenotvirus can you try [this fix](https://github.com/kyleconroy/sqlc/pull/1814) ?

I believe it will fix your issue.

verygoodsoftwarenotvirus:
[...]

PR #1814, opened 2022-08-23
fix(validate): sqlc.arg & sqlc.narg are not "missing"

skabbes:
Fixes: https://github.com/kyleconroy/sqlc/issues/1807


skabbes:
@kyleconroy this is a pretty simple fix - I think just a slight oversight.

Issue #1800, opened 2022-08-16
Add support for queries against `information_schema` tables


yehudamakarov:
### Version

Other

### What happened?

I'm trying to generate a function from a query like this:

```
-- name: GetMyTableColumns :many
SELECT column_name,
       data_type
FROM information_schema.columns
WHERE table_name = 'my_table';
```

but I get the following error while generating:

`schema "information_schema" does not exist`

### Relevant log output

```shell
schema "information_schema" does not exist
```

### Database schema

```sql
information_schema, 
but in the sqlc.yaml I am using `schema: "migrations"` because I'm using `"github.com/golang-migrate/migrate/v4"`
```

### SQL queries

```sql
-- name: GetMyTableColumns :one
SELECT column_name,
       data_type
FROM information_schema.columns
WHERE table_name = 'my_table';
```

### Configuration

```yaml
see above
```

### What operating system are you using?

macOS

### What database engines are you using?

PostgreSQL

### What type of code are you generating?

Go

kyleconroy:
sqlc currently doesn't support this queries, but I'd love to in the future. One person is working on adding some similar tables in https://github.com/kyleconroy/sqlc/pull/1748. Probably best to wait for that to be merged, as we're still working on a few issues.

skabbes:
@yehudamakarov I put up a PR for this here: https://github.com/kyleconroy/sqlc/pull/1815 though you might have to write your query with casts for now.  LIke this:

```sql
SELECT column_name::text,
       data_type::text
FROM information_schema.columns;
```

That is to tell sqlc what datatypes you want to use, because the information_schema tables use a slightly _odd_ type, unless Kyle knows how to address that - I (currently) do not.

PR #1815, opened 2022-08-23
pg-gen: implement information_schema shema

skabbes:
Fixes: https://github.com/kyleconroy/sqlc/issues/1800

Was relatively straightforward after https://github.com/kyleconroy/sqlc/pull/1811

I recommend reviewing commit by commit - since I separated out the "auto-generated files commits" from the rest of the changes.


internal/endtoend/testdata/pg_timezone_names/sqlc.json
```diff
@@ -0,0 +1,28 @@
...
+      "schema": "query.sql",
+      "queries": "query.sql",
+      "gen": {
+        "go": {
+          "sql_package": "pgx/v4",
+          "package": "querytest",
+          "out": "go_pgx"
+        }
+      }
+    },
```
skabbes, line 15: @kyleconroy I found that undoing the nesting (and using multiple `sql` array elements, made these things a lot easier to test.

internal/endtoend/testdata/pg_timezone_names/query.sql
```diff
@@ -0,0 +1,8 @@
+-- name: GetTimezones :many
+SELECT * from pg_catalog.pg_timezone_names;
+
+-- name: GetTables :many
+SELECT table_name::text from information_schema.tables;
```
skabbes, line 5: I had to cast these to `::text` since I think the information_schema uses a database type that is not currently recognized by sqlc (it used `interface{}`).  It might be a simple fix to change that.

kyleconroy (approved):

PR #1840, opened 2022-09-08
fix: Let column names not invalidate Go field names

rollulus:
Characters in column names that do not qualify as a Go identifier are stripped. The title casing as applied to underscores is applied to stripped characters.


rollulus:
I wasn't sure about the e2e test naming convention. Also: thanks for this project!

PR #1854, opened 2022-09-19
docs: various readability improvements

sudomateo:
Various documentation improvements.

- Be explicit that sqlc does not perform migrations, but rather parses the up migrations and ignores the down migrations.
- Update the configuration reference to link to the `gen` key.
- Updated the PostgreSQL getting started documentation to use version 2 of the configuration.

### DIFF
--- a/docs/tutorials/getting-started-postgresql.md
+++ b/docs/tutorials/getting-started-postgresql.md
@@ -23,10 +23,21 @@ packages:
     engine: "postgresql"
     schema: "schema.sql"
     queries: "query.sql"
+
+version: "2"
+sql:
+  - engine: "postgresql"
+    queries: "query.sql"
+    schema: "schema.sql"
+    gen:
+      go:
+        package: "tutorial"
+        out: "tutorial"
 ```
 
-sqlc needs to know your database schema and queries. In the same directory,
-create a file named `schema.sql` with the following contents:
+sqlc needs to know your database schema and queries in order to generate code.
+In the same directory, create a file named `schema.sql` with the following
+content:
 
 ```sql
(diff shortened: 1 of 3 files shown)


docs/tutorials/getting-started-postgresql.md
```diff
@@ -23,10 +23,21 @@ packages:
...
+
+version: "2"
+sql:
+  - engine: "postgresql"
+    queries: "query.sql"
+    schema: "schema.sql"
+    gen:
+      go:
+        package: "tutorial"
+        out: "tutorial"
```
kyleconroy, line 35: It's confusing to have both version 1 and version 2 in the same YAML block. Since this is a getting started guide, let's only use version 2

docs/tutorials/getting-started-postgresql.md
```diff
@@ -60,25 +71,30 @@ DELETE FROM authors
...
-set name = $2,
-bio = $3
+  set name = $2,
+  bio = $3
 WHERE id = $1;
 ```
-Otherwise, to return the updated record back to the user, add this to the `query.sql` file:
+
+Otherwise, to return the updated record to the user, add this to the
+`query.sql`:
```
kyleconroy, line 86:
```suggestion
Otherwise, to return the updated record to the user, add this to `query.sql`:
```

kyleconroy (approved):

docs/tutorials/getting-started-postgresql.md
sudomateo: Whoops! I meant to remove version 1. Fixed!

kyleconroy (approved):

PR #1884, opened 2022-10-06
fix: Override types of aliased columns and named parameters

### LINKED ISSUE #1864: Column type overrides lost when renaming a parameter
### Version

1.15.0

### What happened?

When a type override is defined for a column, it only applies to parameters if they are positional or named the same as the column itself.

In the example below, I expected to generate

```go
func (q *Queries) Renamed(ctx context.Context, userEmail user.Email) (user.ID, error)
```

but with a named parameter I got 

```go
func (q *Queries) Renamed(ctx context.Context, userEmail string) (user.ID, error)
```

unless the parameter was named `email` (which matches the inferred name, so isn't very useful!).

### Relevant log output

_No response_

### Database schema

```sql
CREATE TABLE users (
  id bigserial PRIMARY KEY,
  email text NOT NULL CONSTRAINT unique_email UNIQUE
);
```

### SQL queries

```sql
-- name: Renamed :one
SELECT id
FROM users
WHERE email = @user_email;
```

### Configuration

```yaml
{
  "version": "2",
  "sql": [
    {
      "engine": "postgresql",
      "schema": "query.sql",
      "queries": "query.sql",
      "gen": {
        "go": {
          "out": "db",
          "overrides": [
            {
              "column": "users.id",
              "go_type": "example.com/user.ID"
            },
            {
              "column": "users.email",
              "go_type": "example.com/user.Email"
            }
          ]
        }
      }
    }
  ]
}
```

### Playground URL

https://play.sqlc.dev/p/f212e1ed3b1d4e565c9b65b599a017c20586d8169b10f89d88859723a95ce265

macOS

PostgreSQL

Go

haines:
Fixes #1864

Column type overriding fails when columns are aliased or parameter names do not match the underlying column.

This PR makes the compiler preserve the original column name if a column is aliased in a query or if a named parameter is used instead of a numbered one, and uses that name (if present) to check for a matching type override in the config.


StevenACoffman:
@haines Thanks for keeping this fresh! @kyleconroy We are using this patch and it is working well for us!

PR #1907, opened 2022-10-23
python: Port all Python tests to sqlc-gen-python

kyleconroy:


skabbes:
I gotta say, I really like this direction - and thank you for doing the yak-shaving tasks to make this happen.

PR #1914, opened 2022-10-27
[Go] flags to disable enums or table models output

danicc097:
Use case: when using sqlc alongside https://github.com/xo/xo, I'd want to keep all database generation in the same package. Some table names conflict with xo's generated ones, and all enums conflict. This would allow for more fine-grained output control

PS: flag names debatable


danicc097:
@kyleconroy sorry about how that diff turned out. Don't know if you prefer a separate PR that simply adds running goimports on all generated go files since I had to update all testdata

kyleconroy (commented):
I understand that you'd like to use sqlc along side xo/xo, but this specific feature only makes sense if you're using a tool that generates structs with the exact same name. Instead of this feature, you can delete the `models.go` file after generation.

danicc097:
@kyleconroy Deleting the models file leaves at least the issue that if you select all rows from a column (either `*` or explicitly) the generated method uses the models as a row result and won't compile (external structs not guaranteed to have the same fields/name). Would you accept a flag parameter to `always_emit_row_result_structs` or something like that? It's a one-liner in the codegen code as it is in the PR

Issue #1411, opened 2022-02-02
ORDER BY does not enforce column to exist


theFong:
### Version

1.11.0

### What happened?

When creating a query that orders by a non-existent column, sqlc **does not** emit a `query.sql:x:y: column "adfadsf" does not exist` error.

### Database schema

```sql
-- Example queries for sqlc
CREATE TABLE authors (
  id   BIGSERIAL PRIMARY KEY,
  name text      NOT NULL,
  bio  text
);
```

### SQL queries

```sql
-- name: ListAuthors :many
SELECT * FROM authors
ORDER BY adfadsf;
```

### Configuration

```yaml
version: 1
packages:
  - path: "sqlc"
    name: "sqlc"
    engine: "mysql"
    schema: "schema.sql"
    queries: "query.sql"
```

### Playground URL

https://play.sqlc.dev/p/9b40e9e704f08dc389049f02786f9d072d0c51839a432ff196dadcdfa6b19900

### What operating system are you using?

macOS

### What database engines are you using?

MySQL

### What type of code are you generating?

Go

ambientsound:
It would be great if sqlc could enforce validity of ORDER BY columns. We just had a regression because of dropped columns that were not removed from queries, and it passed straight through generation.

FWIW, we use Postgres.

akutschera:
You are right. While there was code for checking "GROUP BY" clauses, the "ORDER BY" checks were missing. The commit above should fix that.

PR #1915, opened 2022-10-27
fix: check column references in ORDER BY (#1411)

akutschera:
MySQL and Sqlite store the information for the order by columns in a WindowClause while PostgreSQL stores it in a SortClause.
That's why I added two new if-statements.

Refs: #1411


internal/compiler/parse_test.go
```diff
@@ -0,0 +1,104 @@
+package compiler
```
kyleconroy, line 1: I take a bit of a different approach to testing sqlc, opting mainly for end-to-end tests that don't verify internal behavior. Can you take these tests and move them into internal/endtoend/testdata/order_by_non_existing_column?

kyleconroy (changes requested):
Adding additional validation like this is great, but it will cause queries that used to work to fail. If the validation code is wrong, users won't be able to get their queries working. 

Let's add a new configuration option [here](https://github.com/kyleconroy/sqlc/blob/main/internal/config/config.go#L94) called `validate_order_by` that defaults to true (make it a pointer to a bool). Then we can update the error message telling users they can turn it off by setting `validate_order_by: false`.

internal/compiler/parse_test.go
akutschera: That I can do. We actually only need these tests for order by. The group by tests are in internal/endtoend/testdata/invalid_group_by_reference already

kyleconroy:
sqlc getting more strict over time is a good thing. `strict_function_checks` works in V1 as well, so we can put `validate_order_by` in the same spot.

akutschera:
I added the ```validate_order_by``` config option. It works even though it feels a bit awkward to pass this value through multiple function calls. I am open to suggestions to improve this before it gets merged.
I will do the documentation in a separate PR because I also want to add documentation for ```strict_function_checks``` in V1 and I don't think this belongs to this PR.

kyleconroy (approved):

Stumble:
Hey @akutschera ,

There might be a minor bug in checking ambiguous references when joning tables while using alias, see https://github.com/kyleconroy/sqlc/issues/2398

PR #1918, opened 2022-10-29
Only include comments below metadata as docstring

danicc097:


kyleconroy (commented):
Thanks for taking the time to put this pull request together. I'm inclined to not merge this, as it will change the output for existing files without much benefit. If this is causing your queries to break, please include an example and we can revisit this.

Going to close this for now.

danicc097:
Not causing breaking queries, but generates a lot of noise in methods, particularly in longer queries, considering every unrelated SQL comment in a query gets ultimately appended to the docstring. Some queries are obvious and I'm not that concerned but I would prefer a cleaner docstring for the rest, and for that I think it would be easier to know you can just put the actual docstring under the metadata and that's exactly what you get.

PR #1841, opened 2022-09-09
fix: named params contribute to batch parameter count

chrisyxlee:
The batch validation code was only allowing numbered params contribute to the total parameter count of a query, which prevented batch queries with only named parameters from successfully generating code.

Fixes #1625

Issue #1625, opened 2022-05-17
Batch queries with only named parameters fail with `:batch* commands require parameters`


chrisyxlee:
### Version

1.13.0

### What happened?

I have a query that is using solely named parameters (`sqlc.arg(...)`), and `sqlc generate` fails with that. I expected the query to successfully generate Go code. If there's at least 1 `$1`-type parameter, the query works. See this working [playground link](https://play.sqlc.dev/p/d4224c6f7182e5515ee5ace3ccca1527e52f2759c2d868977a1f37e9f5cdecd5).

https://github.com/kyleconroy/sqlc/blob/5c5ae9b1ff772b66c6c77ddb2c523cdfe010de1f/internal/sql/validate/cmd.go#L47-L53

I forked the repo to verify that it's not just failing for my queries. See `CreateBooks` query below.

Really appreciate the batch feature, it's simplified so much code. 👏 

### Relevant log output

```shell
in sqlc/examples/batch
❯ ~/bin/sqlc-dev generate
# package batch
postgresql/query.sql:18:1: :batch* commands require parameters
```

### Database schema

```sql
CREATE TABLE authors (
          author_id SERIAL PRIMARY KEY,
          name text NOT NULL DEFAULT ''
);

CREATE TYPE book_type AS ENUM (
          'FICTION',
          'NONFICTION'
);

CREATE TABLE books (
          book_id SERIAL PRIMARY KEY,
          author_id integer NOT NULL REFERENCES authors(author_id),
          isbn text NOT NULL DEFAULT '' UNIQUE,
          book_type book_type NOT NULL DEFAULT 'FICTION',
          title text NOT NULL DEFAULT '',
          year integer NOT NULL DEFAULT 2000,
          available timestamp with time zone NOT NULL DEFAULT 'NOW()',
          tags varchar[] NOT NULL DEFAULT '{}'
);
```

### SQL queries

```sql
-- name: CreateBook :batchone
INSERT INTO books (
    author_id,
    isbn,
    book_type,
    title,
    year,
    available,
    tags
) VALUES (
    sqlc.arg(named_author),
    sqlc.arg(named_isbn),
    sqlc.arg(named_book_type),
    sqlc.arg(named_title),
    sqlc.arg(named_year),
    sqlc.arg(named_available),
    sqlc.arg(named_tags)
)
RETURNING *;
```

### Configuration

```yaml
{
  "version": "1",
  "packages": [
    {
      "path": "postgresql",
[...]

PR #1924, opened 2022-11-04
codegen: surface batch validation errors

chrisyxlee:
As a result of #1841 (letting named params contribute to param count), the call to `ParamRef` will actually fail for queries with mixed usage of numbered and named parameters. 

This PR mainly ensures that when the failure occurs, the error surfaced to the user makes it obvious what the issue is. Previously, the error would say `:batch* commands require parameters` when the actual error is that the parameters are being mixed.

Now it says `path/to/query.sql:1:1: could not determine data type of parameter $1` or
`path/to/query.sql:15:1: can not mix $1 format with ? format`

With this change, I believe mixed parameters are still allowed for non-batch queries, and I'm not sure if I should change that (please let me know through review).

#1625

### DIFF
--- a/internal/sql/validate/cmd.go
+++ b/internal/sql/validate/cmd.go
@@ -45,7 +45,10 @@ func validateCopyfrom(n ast.Node) error {
 }
 
 func validateBatch(n ast.Node) error {
-	nums, _, _ := ParamRef(n)
+	nums, _, err := ParamRef(n)
+	if err != nil {
+		return err
+	}
 	if len(nums) == 0 {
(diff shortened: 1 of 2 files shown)


kyleconroy:
@chrisyxlee Can you include an example in the comments of a query that will produce the error? I want to make sure we add a test case for this.

kyleconroy (commented):
@chrisyxlee Can you include an example in the comments of a query that will produce the error? I want to make sure we add a test case for this.

chrisyxlee:
Thank you! Apologies for forgetting about this PR -- and sounds good to me. 👍

PR #1973, opened 2022-11-30
Improve support for SQL functions that return tables

### LINKED ISSUE #1322: Functions returning `TABLE` types do not work
### Version

1.11.0

### What happened?

When a function returns a `TABLE` type, you get an "empty list" error or something like that. I'm sorry I saw this earlier today and don't have the exact output, but the repro below should work.

`sqlc generate` fails due to an error.

### Relevant log output

_No response_

### Database schema

```sql
CREATE TABLE accounts (
  id         INTEGER GENERATED ALWAYS AS IDENTITY PRIMAR KEY,
  username   TEXT NOT NULL UNIQUE,
  password   TEXT NOT NULL
);

-- this is a useless and horrifying function cause we don't hash
-- the password, this is just to repro the bug in sqlc
CREATE OR REPLACE FUNCTION register_account(
    _username TEXT,
    _password VARCHAR(70)
)
RETURNS TABLE (
    account_id   INTEGER
)
AS $$
BEGIN
  INSERT INTO accounts (username, password)
       VALUES (
         _username,
         _password
       )
    RETURNING id INTO account_id;

  RETURN NEXT;
END;
$$ LANGUAGE plpgsql;
```

### SQL queries

```sql
-- name: Foo :one
SELECT * FROM register_account('a', 'b');
```

### Configuration

_No response_

### Playground URL

_No response_

Linux

PostgreSQL

Go

jrperritt:
This change adds an `IsArray` field for function parameters and creates structs for SQL functions that return tables.

Some notes about the implementation:
- Selecting from a function that returns a table with a single column will (continue to) return a scalar (or slice of scalar) type.
- The struct representing the returned table is still generated even if the table only has a single column. That is because function/table parsing necessarily happens before query parsing, and I think some of the query-processing logic (like star-expansion) needs to happen before a determination can be made. (This is speculation; it may be the case that unused structs don't have to be generated, but it seemed tricky to get right in my limited experiments).
- The name of the generated struct that represents the returned table is the sqlc-generated name of the SQL function.
- Functions from the built-in Postgresql catalogs are not processed. The reason isn't technical, but rather including them added something like 50,000 lines to the `internal/endtoend/testdata/process_plugin_sqlc_gen_json/gen/codegen.json` file.

Most of the line changes are from running `make regen`. The ~150 relevant line changes are almost entirely in:
- `internal/cmd/shim.go`
- `internal/compiler/output_columns.go`
- `internal/sql/catalog/func.go`
- `protos/plugin/codegen.proto`

There are a couple of tests in `examples/booktest/postgresql/db_test.go`, in addition to the existing function tests that didn't regress.

closes #1322


andrewmbenton:
I don't think it's out of scope necessarily. When @kyleconroy and I last discussed I think we had settled on support for queries using the [`RETURNS TABLE(columns)` syntax](https://www.postgresql.org/docs/current/xfunc-sql.html#XFUNC-SQL-FUNCTIONS-RETURNING-TABLE). He may correct me though.

I haven't had a chance to look closely through your PR to determine whether it meets that requirement though.

jrperritt:
The 2 force- pushes are for
1. Rebasing on `main` and,
2. Changing the enum names that were flagged during linting.

kyleconroy:
A few pieces of feedback on these changes. Instead of generating structs inside `models.go` for these functions, I'd like to use per-query row structs. The Go codegen package already handles this for queries that don't cleanly map to an existing table.

This change simplifies the code as well, as we no longer need to pass function information to the codegen plugin.

```go
const booksByTagsViaFunc = `-- name: BooksByTagsViaFunc :many
SELECT book_id, title, name, isbn, tags FROM sqlc_book_schema.books_by_tags_func($1::VARCHAR[])
`

type BooksByTagsViaFuncRow struct {
	BookID sql.NullInt32
	Title  sql.NullString
	Name   sql.NullString
	Isbn   sql.NullString
	Tags   sql.NullString
}

func (q *Queries) BooksByTagsViaFunc(ctx context.Context, dollar_1 []string) ([]BooksByTagsViaFuncRow, error) {
	...
}
```

As for the tests you've added, can you move all the changes for the examples directory to individual testcases into the end-to-end directory? The examples lean towards simplicity and most users of sqlc don't use stored procedures.

jrperritt:
> A few pieces of feedback on these changes. Instead of generating structs inside `models.go` for these functions, I'd like to use per-query row structs. The Go codegen package already handles this for queries that don't cleanly map to an existing table.
> 
> This change simplifies the code as well, as we no longer need to pass function information to the codegen plugin.

Indeed, that turned out to be a much better approach.

> As for the tests you've added, can you move all the changes for the examples directory to individual testcases into the end-to-end directory? The examples lean towards simplicity and most users of sqlc don't use stored procedures.

Sure, I've added an `internal/endtoend/testdata/func_return_table/` directory.

jrperritt:
@kyleconroy  This is ready for another look.

jrperritt:
I rebased this on `main` and added support for multidimensional array function parameters, which was implemented some time in the last year without me realizing it.

Issue #1977, opened 2022-12-01
Go renames don't work in config v2


josharian:
### Version

1.15.0

### What happened?

[I'm actually running 1.16.0, but the dropdown is missing that choice.]

I tried porting my config file from v1 to v2. Go renames and overrides stopped working.

To confirm that I didn't just make a config file mistake, I instrumented v2ParseConfig. I can confirm that my requested renames and overrides are present in the config field `conf.SQL[0].Gen.Go` (per the docs), but the downstream code is reading the configs from `conf.Gen.Go`, including later in v2ParseConfig.

### What type of code are you generating?

Go

Emyrk:
My PR should fix this: https://github.com/kyleconroy/sqlc/pull/2001

PR #2001, opened 2022-12-15
fix: Honor Package level renames in v2 yaml config

Emyrk:
When following the docs for renames for `pkg` scoped renames, the renames were not applying. This is my first look into the sqlc config, but it looked like things were just not plumbed through.

I just adds setting the renames from the yaml parse. In the combine settings I made sure to assign the renames from the pkg (the global was already set). This change does make pkg renames override global ones.

---

For unit testing, I noticed all the `endtoend/testdata` use a v1 config. Instead of adding more vectors to that to also do `v2` directories, what if we just added a `sqlc.v1.yaml` and `sqlc.v2.yaml` to each directory and run it twice? Or something of that nature to test v2 configs? At present this is not tested aside from some manual checking.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -213,13 +213,19 @@ func Combine(conf Config, pkg SQL) CombinedSettings {
 	cs := CombinedSettings{
 		Global:  conf,
 		Package: pkg,
+		Rename:  map[string]string{},
 	}
 	if conf.Gen.Go != nil {
-		cs.Rename = conf.Gen.Go.Rename
+		for k, v := range conf.Gen.Go.Rename {
+			cs.Rename[k] = v
+		}
 		cs.Overrides = append(cs.Overrides, conf.Gen.Go.Overrides...)
 	}
 	if pkg.Gen.Go != nil {
 		cs.Go = *pkg.Gen.Go
+		for k, v := range pkg.Gen.Go.Rename {
+			cs.Rename[k] = v
+		}
 		cs.Overrides = append(cs.Overrides, pkg.Gen.Go.Overrides...)
 	}
 	if pkg.Gen.JSON != nil {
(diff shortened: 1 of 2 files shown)


alicebob:
This seems to fix #1977.

Emyrk:
Is it possible to get this in soon? Seems small and trivial?

mpyw:
I noticed the same problem and wrote a patch, but then I found that this PR had already been submitted! 🤣 

```diff
diff --git a/internal/config/config.go b/internal/config/config.go
index 0dfd57bc..1c8942c8 100644
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -214,12 +214,22 @@ func Combine(conf Config, pkg SQL) CombinedSettings {
 		Global:  conf,
 		Package: pkg,
 	}
+	merge := func(dst, src map[string]string) map[string]string {
+		for k, v := range src {
+			if dst == nil {
+				dst = make(map[string]string)
+			}
+			dst[k] = v
+		}
+		return dst
+	}
 	if conf.Gen.Go != nil {
-		cs.Rename = conf.Gen.Go.Rename
+		cs.Rename = merge(cs.Rename, conf.Gen.Go.Rename)
 		cs.Overrides = append(cs.Overrides, conf.Gen.Go.Overrides...)
 	}
 	if pkg.Gen.Go != nil {
 		cs.Go = *pkg.Gen.Go
+		cs.Rename = merge(cs.Rename, pkg.Gen.Go.Rename)
 		cs.Overrides = append(cs.Overrides, pkg.Gen.Go.Overrides...)
 	}
 	if pkg.Gen.JSON != nil {
```

mpyw:
@kyleconroy There are a workaround with global `rename` config, but the bug may confuse many people. A patch version should be released if possible.

sgielen:
@kyleconroy respectful bump to make you aware of this tiny PR fixing #1977 :-)

Emyrk:
> @kyleconroy There are a workaround with global `rename` config, but the bug may confuse many people. A patch version should be released soon if possible.

That is what I did in the meantime.

PR #2006, opened 2022-12-24
Adding the ability to change the default scheme to generate DTO and queries

debugger84:
PostgreSQL allows you to store your tables in different schemas instead of the `public` one. 

I haven't found the solution to generate code for the tables in the schema that differs from `public`. So, I have created a fix to do it.

How it works:
1. Place your table `test` in the schema, for example, `test`.
2. Add the option `default_schema: "test"` to your sqlc.yaml
```
sql:
    engine: "postgresql"
    default_schema: "test"
```
3. In your queries file type the table name with the schema
```
-- name: GetTestItem :one
select * from "test"."test" where id=$1 limit 1;
```


kyleconroy:
Instead of adding support for a default schema, we should implement the logic that PostgreSQL uses for [schema search path](https://www.postgresql.org/docs/current/ddl-schemas.html#DDL-SCHEMAS-PATH).

PR #2018, opened 2023-01-13
 Manual fix for incorrect handling of "calories"

### LINKED ISSUE #2017: Table "calories" incorrectly generates model "calory", not "calorie"
### Version

1.16.0

### What happened?

[Singular("Calories") returns "Calory" · Issue #23 · jinzhu/inflection](https://github.com/jinzhu/inflection/issues/23)

### Relevant log output

_No response_

### Database schema

_No response_

### SQL queries

_No response_

### Configuration

_No response_

### Playground URL

https://play.sqlc.dev/p/c02cc2497dda3585969b11af42b8f4c827cad8eb387ec3dafd6a37415c141a98

_No response_

_No response_

_No response_

mpyw:
- Fixes #2017

### DIFF
--- a/internal/endtoend/testdata/inflection/mysql/go/models.go
+++ b/internal/endtoend/testdata/inflection/mysql/go/models.go
@@ -6,6 +6,10 @@ package querytest
 
 import ()
 
+type Calorie struct {
+	ID string
+}
+
 type Campus struct {
 	ID string
 }
--- a/internal/endtoend/testdata/inflection/mysql/go/query.sql.go
+++ b/internal/endtoend/testdata/inflection/mysql/go/query.sql.go
@@ -9,6 +9,33 @@ import (
 	"context"
 )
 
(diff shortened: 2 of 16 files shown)


mpyw:
@kyleconroy

I noticed that this change didn't make sense! The name in my project was `product_calories`, so the exact match has no effect...

Why not add the feature renaming the table name function on the top of the existing feature renaming field names?

mpyw:
I noticed that the feature has been already supported, but it looks to be broken.

Related:

- #1977
- #2001

mpyw:
Current workaround: Use global `rename` config

pexarkh:
The feature to rename tables would be very much appreciated. For example, in a project i'm currently on, there is a table called `EmailFTS`, where **FTS** stands for **Full-Text Search** (it's a separate sqlite table serving as a full text index). So sqlc makes `type EmailsFT struct` out of it which is not really fully meaningful...

Issue #1965, opened 2022-11-23
MIN() returns an interface{}


abh:
### Version

Other

### What happened?

(version 1.16.0)

`select min(id) from authors;` creates a function that returns an interface{} instead of a NULL-able version of the `id` type.

#1574 talks about this for DATETIME types, but other issues indicates that it should work for simpler types, maybe?

If not, what's the workaround? Parsing the interface{} is awkward, best I can tell.

### Playground URL

https://play.sqlc.dev/p/96e9cf0291f2e13cc784144f8abd4d5a556c549f2bfe909a26377703f671ff61

### What operating system are you using?

macOS

### What database engines are you using?

MySQL

### What type of code are you generating?

Go

andrewmbenton:
This is still an issue in v1.18.0: https://play.sqlc.dev/p/3931aa6d071068031e5eb03f4e40bd474d22afb2e47aabe0c34e68101e76d2f1.

Adding support for CAST seems like an important thing to do as well, so I'll reopen that issue.

abh:
It didn't work for my actual use case, but I noticed that in a simple use case doing a sub-select sometimes(?) makes sqlc figure out the correct type.

`select id from (select min(id) from authors) as min_author;`

https://play.sqlc.dev/p/e4b7f02a9c2ab5a4cea69f216758eb0d0a9ba42576804548e4f45b8c47e79910

francesconi:
@kyleconroy The same is true for `lead()` function. Even though I could cast it to, let's say `smallint`, it is not nullable.

sknig:
The workaround I use is to add the result to 0.  SELECT 0 + count(value) FROM table; gets correctly recognized as an int.

Issue #1901, opened 2022-10-17
sum in select result in model field type interface{}


sosolyht:
### Version

1.14.0

### What happened?

Hello, I found an error while using sqlc and mysql

return value must be int64 

but  it's an interface need cast to int64

```go
const totalVAT = `-- name: TotalVAT :one
SELECT sum(vat) as vatpence
FROM sales
`

func (q *Queries) TotalVAT(ctx context.Context) (interface{}, error) {
	row := q.db.QueryRowContext(ctx, totalVAT)
	var vatpence interface{}
	err := row.Scan(&vatpence)
	return vatpence, err
}
```

### Database schema

```sql
CREATE TABLE sales (
  vat integer      NOT NULL
);
```

### SQL queries

```sql
-- name: TotalVAT :one
SELECT sum(vat) as vatpence
FROM sales;
```

### Playground URL

https://play.sqlc.dev/p/6552b583f39af63c0d407f76b62cafdfe471b130c76d9066b9843b8c20b8cd90

### What operating system are you using?

macOS

### What database engines are you using?

MySQL

### What type of code are you generating?

Go

brlala:
use this as a workaround 
```
SELECT sum(vat) :: integer as vatpence
FROM sales
```

sosolyht:
@brlala thank you, but this solution does not work with mysql

sosolyht:
```go
func convertInterfaceToInt64(t interface{}) (int64, error) {
	switch t := t.(type) {
	case int64:
		return t, nil
	case int:
		return int64(t), nil
	case string:
		return strconv.ParseInt(t, 10, 64)
	case []byte:
		return strconv.ParseInt(string(t), 10, 64)
	default:
		return 0, fmt.Errorf("type %T not supported", t)
	}
}
```

query is return from interface to byte type 

so we can solve this problem

abh:
Related to #1622

andrewmbenton:
> use this as a workaround
> 
> ```
> SELECT sum(vat) :: integer as vatpence
> FROM sales
> ```

As noted, this workaround doesn't work for MySQL. The following does work though:

```sql
SELECT CAST(sum(vat) AS unsigned) vatpence
FROM sales;
```

https://play.sqlc.dev/p/a046db4043e5fb11a9135f6632d96bcd4506c754e9bab06b1706ef06c5ea6dc9

Issue #1866, opened 2022-09-28
Unable to Type Alias


christoph-pflueger:
### Version

1.15.0

### What happened?

Example Query:

```sql
-- name: CheckIfSomeRandomColumnIsNotNull :one
SELECT
  `some_random_column` IS NOT NULL AS `is_some_random_column_not_null`
FROM
  `example`
WHERE
  `id` = ?;
```

Results in:

```go
type CheckIfSomeRandomColumnIsNotNullRow struct {
	IsSomeRandomColumnNotNull interface{}
}
```

Every attempt to use `::bool` to type the struct field correctly resulted in a syntax error.

### Database schema

```sql
CREATE TABLE `example` (
  `id` bigint unsigned NOT NULL AUTO_INCREMENT,
  `some_random_column` bigint unsigned DEFAULT NULL,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB AUTO_INCREMENT=10 DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci;
```

### What operating system are you using?

Linux

### What database engines are you using?

MySQL

### What type of code are you generating?

Go

kyleconroy:
Fixed https://play.sqlc.dev/p/49ae28cb5dac3a2020f72fc17203a53921e9882990987c1a23329b8125c16729

Issue #1622, opened 2022-05-16
Using MYSQL functions ( CONVERT and CAST) result in removing column from struct


KarimEKhalifa:
### Version

1.13.0

### What happened?

When creating an insert query for a json column, sqlc does not generate the column name in the golang struct  if I try to add `CAST(CONVERT( ? USING utf8) AS JSON)`

From playing around with MySql and Postgres in the playground, I've noticed that it works fine in Postgres but not when using MySql

Postgres playground:
https://play.sqlc.dev/p/ea800eff2a7beb25bc960f7d839ea96039b5dad0c46f16dddd97f76dc28d3388

MySql playground:
https://play.sqlc.dev/p/38ec6bf15657aa79f4a3dd7c72c559934038a591625869b2dfe59a87498f7b1f

### Database schema

```sql
CREATE TABLE product_versions (
    id INT NOT NULL AUTO_INCREMENT PRIMARY KEY,
    product_id INT NOT NULL,
    product_key VARCHAR(255) NOT NULL UNIQUE,
    version INT NOT NULL,
    type ENUM ('LARGE', 'MEDIUM', 'SMALL') NOT NULL, 
    status ENUM ('ACTIVE', 'DRAFT') NOT NULL,
    attributes JSON,
    created_at TIMESTAMP NOT NULL,
    created_by INT NOT NULL
);
```

### SQL queries

```sql
-- name: CreateProductVersion :execresult
INSERT INTO product_versions (
        product_id,
        product_key,
        version,
        type,
        status,
        attributes,
        created_at,
        created_by
    ) VALUES (
        ?, ?, ?, ?, ?, CAST(CONVERT( ? USING utf8) AS JSON), NOW(), ?
);
```

### Configuration

```yaml
version: "1"
project:
  id: "project"
packages:
  - name: "sqlc"
    path: "./sqlc"
    queries: "./sqlc/query/"
    schema: "./migrations"
    engine: "mysql"
    emit_json_tags: true
    emit_interface: true
    emit_exact_table_names: false
```

### Playground URL

https://play.sqlc.dev/p/ea7294bd54c9eae950221ae124b4685690c2e5328a9d6ae444bb1a455924c14e

### What operating system are you using?

macOS

### What database engines are you using?

MySQL

### What type of code are you generating?

Go

ryanpbrewster:
[...]

PR #2020, opened 2023-01-14
Implement cast function parser for MySQL

RadhiFadlillah:
### What is this

As the title said, this PR wants to add support for `CAST` function in MySQL.

This PR is based from PR by @ryanpbrewster [here][ryan-pr] (which unfortunately he didn't send here, and only exist in his repository).

### Why is this PR created

Currently `sqlc` unable to infer the correct type from SQL function like `MAX`, `MIN`, `SUM`, etc. For those function, sqlc will return its value as `interface{}`. This behavior can be seen in this [playground].

As workaround, it advised to use `CAST` function to explicitly tell what is the type for that column, as mentioned in #1574.

Unfortunately, currently sqlc only support `CAST` function in PostgreSQL and not in MySQL. Thanks to this, right now MySQL users have to parse the `interface{}` manually, which is not really desirable.

### What does this PR do?

- Implement `convertFuncCast` function for MySQL.
- Add better nil pointer check in some functions that related to `convertFuncCast`.

I haven't write any test because I'm not sure how and where to put it. However, as far as I know the code that handle `ast.TypeCast` for PostgreSQL also don't have any test, so I guess it's fine :man_shrugging: 

### Related issues

- #687, which currently is the oldest MySQL issue that still opened.
- #1622 
- #1866
- #1901
- #1965

[ryan-pr]: https://github.com/ryanpbrewster/sqlc/pull/1
[playground]: https://play.sqlc.dev/p/6fefc08975fcd36229c89b16daad6358d1bcc8ca8db2f79470e4915ce044b806

### DIFF
--- a/internal/compiler/compat.go
+++ b/internal/compiler/compat.go
@@ -31,9 +31,7 @@ type Relation struct {
 }
 
 func parseRelation(node ast.Node) (*Relation, error) {
-	switch n := node.(type) {
-
-	case *ast.List:
+	if n, ok := node.(*ast.List); ok && n != nil {
 		parts := stringSlice(n)
 		switch len(parts) {
 		case 1:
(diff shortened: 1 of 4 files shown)


abh:
Why was this closed?

RadhiFadlillah:
Oh, my bad. It seems I've deleted my fork for this PR by accident when I was deleting several of my unused repositories.

TIL: The PR will be closed automatically if the fork is erased.

RadhiFadlillah:
I'm unable to recover the deleted repository so I decided to just recreate the PR. Sorry for the trouble :sweat:

PR #2026, opened 2023-01-18
cmd: Generate packages in parallel

kyleconroy:
Use `errgroup` to parallelize generating code for each SQL package. For configurations with many packages, this should dramatically speed up generation.

### DIFF
--- a/internal/cmd/cmd.go
+++ b/internal/cmd/cmd.go
@@ -44,6 +44,7 @@ func Do(args []string, stdin io.Reader, stdout io.Writer, stderr io.Writer) int 
 	rootCmd.SetIn(stdin)
 	rootCmd.SetOut(stdout)
 	rootCmd.SetErr(stderr)
+	rootCmd.SilenceErrors = true
 
 	ctx := context.Background()
 	if debug.Debug.Trace != "" {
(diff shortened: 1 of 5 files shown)


internal/cmd/cmd.go
```diff
@@ -44,6 +44,7 @@ func Do(args []string, stdin io.Reader, stdout io.Writer, stderr io.Writer) int
 	rootCmd.SetIn(stdin)
 	rootCmd.SetOut(stdout)
 	rootCmd.SetErr(stderr)
+	rootCmd.SilenceErrors = true
```
kyleconroy, line 47: When we moved to `RunE`, we accidentally added some additional error output. Turning that off for now.

internal/cmd/generate.go
```diff
@@ -159,71 +163,94 @@ func Generate(ctx context.Context, e Env, dir, filename string, stderr io.Writer
...
+			for _, file := range resp.Files {
+				files[file.Name] = string(file.Contents)
+			}
 
+			m.Lock()
+			for n, source := range files {
+				filename := filepath.Join(dir, out, n)
+				output[filename] = source
+			}
+			m.Unlock()
```
kyleconroy, line 239: Writing to a map isn't threadsafe, so put this in mutex. In the future might just use `sync.Map` instead.

internal/cmd/generate.go
```diff
@@ -159,71 +163,94 @@ func Generate(ctx context.Context, e Env, dir, filename string, stderr io.Writer
...
+			var name, lang string
+			parseOpts := opts.Parser{
+				Debug: debug.Debug,
+			}
 
-		result, failed := parse(ctx, name, dir, sql.SQL, combo, parseOpts, stderr)
-		if failed {
-			packageRegion.End()
-			errored = true
-			break
```
kyleconroy, line 203: If parsing failed, we'd return early. Instead, return errors from every package.

PR #2027, opened 2023-01-19
Implement ALTER TYPE SET SCHEMA

linux2647:
## Problem

Altering the schema for a given type is not reflected in generated code.

For example,

```sql
ALTER TYPE event SET SCHEMA new;
```

results in

```go
type Event string
```

when it should be

```go
type NewEvent string
```

## Solution

Implement `ALTER TYPE ... SET SCHEMA`.


linux2647:
@kyleconroy Is there anything additional you need to review this PR?

kyleconroy (approved):

PR #2046, opened 2023-02-03
fix: Add import statements even if only pointer types exist

zaneli:
Fixed missing import statement when only pointer types such as`*time.Time` or `*netip.Addr` exist.


kyleconroy (approved):

Stumble:
Hi @zaneli @kyleconroy 

I made [a similar change](https://github.com/Stumble/sqlc/commit/eebda8f8d4153bcd32a86341f02e9f6ecf5b6d39#diff-5888848f9ffbddb7b5d76142e6226de61375df0b151360152edae46b25965fdeR481) in my forked branch of sqlc. Changes are almose the same except for that in the 
trim function, `[]*` is also trimmed, to cover cases that when there exists a selected column of type of array of nullable types.
Is it possible?


```

// trimTypePrefix removes:
// *T => T
// []T => T
// []*T => T
func trimTypePrefix(t string) string {
	return strings.TrimPrefix(strings.TrimPrefix(t, "[]"), "*")
}
```

Issue #1956, opened 2022-11-17
SQLc failing to build due to upstream dependency due to sprintf deprecation (affects MacOS Ventura)


17twenty:
### Version

Other

### What happened?

Current versions of SQLc are failing to build on latest version of XCode / MacOS Ventura.

This is due to a sprintf deprecation in pg_query_go which is used by SQLc see https://github.com/pganalyze/pg_query_go/issues/67 for more.

The issue is fixed upstream which I've been able to reproduce locally by merging in the C code from https://git.postgresql.org/gitweb/?p=postgresql.git;a=blob_plain;f=src/port/snprintf.c;h=81d9c8c2740d1b90b69077cb3b5a1d9576e51dad;hb=450ee701261a7ff5fe35842b91384eea46e81c2e into my local build.

### Relevant log output

```shell
go install github.com/kyleconroy/sqlc/cmd/sqlc@v1.14.0
# github.com/pganalyze/pg_query_go/v2/parser
In file included from gram.y:46:
In file included from ../../../../pkg/mod/github.com/pganalyze/pg_query_go/v2@v2.1.0/parser/include/postgres.h:46:
In file included from ../../../../pkg/mod/github.com/pganalyze/pg_query_go/v2@v2.1.0/parser/include/c.h:1330:
In file included from ../../../../pkg/mod/github.com/pganalyze/pg_query_go/v2@v2.1.0/parser/include/port.h:17:
In file included from /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/netdb.h:91:
In file included from /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/netinet/in.h:81:
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/sys/socket.h:425:1: error: expected ';' after top level declarator
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/sys/constrained_ctypes.h:587:101: note: expanded from macro '__CCT_DECLARE_CONSTRAINED_PTR_TYPES'
[...]

PR #2068, opened 2023-02-14
Only enable postgresql support when cgo is enabled as it is required

### LINKED ISSUE #2050: Build tags to exclude postgresql engine

Would it be acceptable to add a build tag such as `disable_postgresql` to be able to exclude postgresql code from the build completely?

Currently, it is already stubbed out for windows

https://github.com/kyleconroy/sqlc/blob/58b0b6593fc665600619b0dfc69909c68ff2cda2/internal/engine/postgresql/parse_windows.go#L1

So the change would just be adding something like `|| disable_postgresl`

The postgresl uses heavyweight cgo, resulting in issues like #1956, and likely the reason for the lack of windows support in the first place. Having a build tag would allow avoiding the scary looking warnings when using `go run`. Hopefully it's ok given the one-line nature of the change (could prefix `experimental_` or something if not wanting to support indefinitely).

PostgreSQL

Go

anuraaga:
Fixes #2050

postgresql parser requires cgo to build the native parser library, which can cause build issues such as #1956. Currently there is no way to disable support when it is not needed and the build issues can cause a problem. With this change, `CGO_ENABLED=0` can be used to work with the other parsers and avoid build issues.

### DIFF
--- a/internal/engine/postgresql/parse_disabled.go
+++ b/internal/engine/postgresql/parse_disabled.go
@@ -1,5 +1,5 @@
-//go:build windows
-// +build windows
+//go:build windows || !cgo
+// +build windows !cgo
 
 package postgresql
 
--- a/internal/engine/postgresql/convert.go
+++ b/internal/engine/postgresql/convert.go
@@ -1,5 +1,5 @@
-//go:build !windows
-// +build !windows
+//go:build !windows && cgo
+// +build !windows,cgo
 
 package postgresql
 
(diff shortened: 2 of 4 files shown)


internal/engine/postgresql/parse_disabled.go
```diff
@@ -1,5 +1,5 @@
-//go:build windows
-// +build windows
+//go:build windows || !cgo
```
anuraaga, line 1: Had to rename this file since `_windows` was implicitly causing the file to only be used on windows

PR #2069, opened 2023-02-14
Propagate env to process plugin and support command args

anuraaga:
This is a strawman's proposal that I noticed while working on #2068 and feel free to reject it. Unit tests wouldn't pass because they expect installation of `sqlc-gen-json` on the running machine. I guess it's best if tests can be run without any external dependencies outside the codebase. Luckily, `go run` is basically guaranteed to work fine and could be an appropriate alternative.

I started by adding support for splitting args out of the cmd string. Then I also noticed that the user's environment is not propagated to the process when run - I think generally users would expect the environment to be available when an external process is run as the plugin may have dependencies on it, and it would otherwise be impossible to run the plugin. Though if not propagating was intentional, then that's fine too and definitely this PR can be closed.

### DIFF
--- a/internal/endtoend/testdata/process_plugin_sqlc_gen_json/sqlc.json
+++ b/internal/endtoend/testdata/process_plugin_sqlc_gen_json/sqlc.json
@@ -21,7 +21,7 @@
     {
       "name": "jsonb",
       "process": {
-        "cmd": "sqlc-gen-json"
+        "cmd": "go run ../../cmd/sqlc-gen-json"
       }
     }
   ]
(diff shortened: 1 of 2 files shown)


kyleconroy:
@anuraaga Thank you so much for the pull request. I'm sorry that it's taken a few months for me to respond.

As it is, I'm not comfortable merging this for a few reasons. Resetting the environment variables was a specific choice: this makes it a bit more difficult for process plugins to silently exfiltrate environment variables. As for the commands themselves, I'd like to continue following the protoc model for plugins when don't allow for custom arguments.

If you have more specific issues with the current setup, happy to hear what those are so that we can figure out a way forward.

PR #2089, opened 2023-02-16
fix: resolve duplicate fields generated when inheriting multiple tables

ethan-xiao:
For example, I created tables following:
create table example_a {
id integer not null,
name varchar(32),
};
create table example_b {} inherits (example_a);
create table example_c {} inherits (example_a, example_b);

and then sqlc generated golang struct for table example_c:
type ExampleC struct {
ID int32
Name string
ID int32
Name string
}

So I fixed this in this PR.

### DIFF
--- a/internal/sql/catalog/table.go
+++ b/internal/sql/catalog/table.go
@@ -240,12 +240,21 @@ func (c *Catalog) createTable(stmt *ast.CreateTableStmt) error {
 	}
 
 	tbl := Table{Rel: stmt.Name, Comment: stmt.Comment}
+	m := make(map[string]struct{}) // used to check for duplicate column names
 	for _, inheritTable := range stmt.Inherits {
 		t, _, err := schema.getTable(inheritTable)
 		if err != nil {
 			return err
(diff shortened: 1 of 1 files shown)


kyleconroy (changes requested):
Could you please add a new end-to-end test for this behavior? It's also helpful to see what the previous behavior looked like.

kyleconroy:
> A table can inherit from more than one parent table, in which case it has the union of the columns defined by the parent tables. Any columns declared in the child table's definition are added to these. **If the same column name appears in multiple parent tables, or in both a parent table and the child's definition, then these columns are “merged” so that there is only one such column in the child table**. To be merged, columns must have the same data types, else an error is raised. Inheritable check constraints and not-null constraints are merged in a similar fashion. Thus, for example, a merged column will be marked not-null if any one of the column definitions it came from is marked not-null. Check constraints are merged if they have the same name, and the merge will fail if their conditions are different.

https://www.postgresql.org/docs/current/ddl-inherit.html

PR #2090, opened 2023-02-16
Remove direct use of pgtype.Array with pgx/v5

### LINKED ISSUE #2083: Don't use pgtype.Array with pgx/v5 driver

`pgtype.Array` is very cumbersome to use and does not serialize to JSON well. It would be much more useful for `pgtype.FlatArray` to be the default.

PostgreSQL

Go

carlmjohnson:
Fixes #2083

### DIFF
--- a/internal/codegen/golang/go_type.go
+++ b/internal/codegen/golang/go_type.go
@@ -38,9 +38,6 @@ func goType(req *plugin.CodeGenRequest, col *plugin.Column) string {
 	}
 	typ := goInnerType(req, col)
 	if col.IsArray {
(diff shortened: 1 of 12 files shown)


kyleconroy:
```
I don't know the context for the sqlc thing, but I'm a little surprised whenever I see pgtype.FlatArray or especially pgtype.Array used directly. Typically, pgx can directly work with []T, [][]T, etc. I only anticipated pgtype.Array being used directly if the application actually needed to work with non-standard starting indexes.
```

What about nullable values? If I have a three strings and a NULL, I don't think those can be scanned into a `[]string`.

kyleconroy (approved):

PR #2099, opened 2023-02-22
fix: clarify error message for PostgreSQL Engine support

mikemherron:
### Problem
When trying to build the project under a fresh Ubuntu WSL install, tests would fail with the error "the PostgreSQL engine does not support Windows" - however, the actual reason was that `gcc` was not installed, therefore CGO was not available.  This occurred because postgresql/parse_disabled.go is included when building on windows _OR_ when CGO is not available - however it's possible (as in my case) to be running in a linux environment and still have CGO unavailable. 

### Solution
This PR updates the error message to clarify that both Windows and any environment without CGO is not supported.


kyleconroy:
Fixed by https://github.com/kyleconroy/sqlc/pull/2100

Issue #1833, opened 2022-09-07
AFTER and BEFORE is ignored when adding values to an ENUM


christeredvartsen:
### Version

1.15.0

### What happened?

PostgreSQL supports specifying `BEFORE` or `AFTER` when adding a value to an existing `ENUM`, see [docs](https://www.postgresql.org/docs/current/sql-altertype.html). sqlc does not pick this up and simply appends the new values, so when the `emit_all_enum_values` configuration parameter is set to `true`, the order of the values returned from the `All*Values()` function is not correct.

I have attached a playground URL where the necessary statements and configuration can be found. If you look at the generated implementation of `AllEnumTypeValues` in `db/models.go`, you can see the order is not correct.

Given the following statements:

```sql
CREATE TYPE enum_type AS ENUM ('first', 'last');
ALTER TYPE enum_type ADD VALUE 'second' AFTER 'first';
ALTER TYPE enum_type ADD VALUE 'third' BEFORE 'last';
```

I would expect the `AllEnumTypeValues` implementation to look like this:

```go
func AllEnumTypeValues() []EnumType {
	return []EnumType{
		EnumTypeFirst,
		EnumTypeSecond,
		EnumTypeThird,
		EnumTypeLast,
	}
}
```

but instead, it looks like this:

```go
func AllEnumTypeValues() []EnumType {
	return []EnumType{
		EnumTypeFirst,
		EnumTypeLast,
		EnumTypeSecond,
		EnumTypeThird,
	}
}
```

### Playground URL

https://play.sqlc.dev/p/746548c45a823783171d5a09b59680132882e1e6786137e2e0dba3f8bcf5b1e2

### What operating system are you using?

macOS

### What database engines are you using?

PostgreSQL

### What type of code are you generating?

Go

christeredvartsen:
@kyleconroy If you give me some pointers as to where in the code base to start looking I might be able to fix it myself.

christeredvartsen:
I have been poking around in the codebase and have an idea where to start, but, I'm unable to get debugging to work. I keep getting `internal compiler error: NewBulk too big: nbit=22255 count=1028433 nword=696 size=715789368`.

Any tips on how to get debugging to work I can probably figure out how to fix my initial issue.

akutschera:
It looks like you are also affected by #1849

PR #2110, opened 2023-02-24
Implement support for enum ordering

mikemherron:
Fix for #1833 

Also updated enum error messages to return the enum name (current code just outputs `*ast.TypeName` inside the error string) and used `slices.Index` to find existing values.


mikemherron:
Just realised `slices.Index` is an `x` package 🤦‍♂️ will re-open shortly...

PR #2114, opened 2023-02-24
Upgrade package, many build errors

kyleconroy:


oliverpool:
I have been using the `postgres` engine and then running my code on `sqlite` (since the sqlite engine is missing a few features).
So I just got bitten by this change (but since my workflow is hacky, this is not a real surprise).

Replacing all `?` is not so tedious and thanks to compile time type checking, I am pretty confident of update :+1:

PR #2141, opened 2023-03-11
dolphin: Support joins in DELETE statements

Ice3man543:
This PR is a contiuation of pr #883 and applies the hotfix for DELETE statements with JOIN in them. Earlier the generation used to panic but now the case is handled correctly.


Ice3man543:
Closing as unneeded

Issue #1596, opened 2022-05-09
Unnest into alias doesn't define the column


Jille:
### Version

1.12.0

### What happened?

I guess the parser doesn't understand that the alias after UNNEST defines a column, so when later trying to determine the type of the column, it thinks the column doesn't exist.

### Relevant log output

```shell
query.sql:2:8: column "x" does not exist
```

### SQL queries

```sql
-- name: FromUnnest :many
SELECT x FROM UNNEST(array[5,6]) x;
```

### Playground URL

https://play.sqlc.dev/p/01ee6a49034c018482526ca75aa29cf93aa447d3d93417543266cd4a6d1577b0

### What operating system are you using?

Linux

### What database engines are you using?

PostgreSQL

### What type of code are you generating?

Go

Jille:
```
$ SQLCDEBUG=dumpast=1 sqlc generate
([]interface {}) (len=1 cap=1) {
 (*ast.RawStmt)(0xc00052cb60)({
  Stmt: (*ast.SelectStmt)(0xc0003f1e40)({
   DistinctClause: (*ast.List)(0xc000642570)({
    Items: ([]ast.Node) <nil>
   }),
   IntoClause: (*ast.IntoClause)(<nil>),
   TargetList: (*ast.List)(0xc000642588)({
    Items: ([]ast.Node) (len=1 cap=1) {
     (*ast.ResTarget)(0xc00063f3e0)({
      Name: (*string)(<nil>),
      Indirection: (*ast.List)(0xc0006425a0)({
       Items: ([]ast.Node) <nil>
      }),
      Val: (*ast.ColumnRef)(0xc00052cae0)({
       Name: (string) "",
       Fields: (*ast.List)(0xc0006425b8)({
        Items: ([]ast.Node) (len=1 cap=1) {
         (*ast.String)(0xc0006445a0)({
          Str: (string) (len=1) "x"
         })
        }
       }),
       Location: (int) 33
      }),
      Location: (int) 33
     })
    }
   }),
   FromClause: (*ast.List)(0xc0006425d0)({
    Items: ([]ast.Node) (len=1 cap=1) {
     (*ast.RangeFunction)(0xc00052cb40)({
      Lateral: (bool) false,
      Ordinality: (bool) false,
      IsRowsfrom: (bool) false,
      Functions: (*ast.List)(0xc0006425e8)({
       Items: ([]ast.Node) (len=1 cap=1) {
        (*ast.List)(0xc000642600)({
[...]

PR #2156, opened 2023-03-17
feat: Allow use of table and column aliases for table functions returning unknown types

### LINKED ISSUE #1205: Column not found when using WITH ORDINALITY clause
### Version

1.10.0

### What happened?

When trying to generate a query using sqlc which includes WITH ORDINALITY then sqlc returns an error which says that the columns cannot be found.

### Relevant log output

```shell
column "id" does not exist
```

### Database schema

_No response_

### SQL queries

```sql
SELECT td.id, td.rn FROM unnest(@foo::uuid[]) WITH ORDINALITY AS td(id, rn);
```

### Configuration

_No response_

### Playground URL

https://play.sqlc.dev/p/298dce5f371c9759abe7a5beb3d0e5b6ec2a4b4f9cb9cbf96a3d8e4404a7f744

Linux

PostgreSQL

Go

haines:
At the moment, when a table function returns a type that isn't a known table, it's excluded from the output columns completely. This makes it impossible to use `unnest()` or `unnest() WITH ORDINALITY` with a table alias (https://github.com/kyleconroy/sqlc/issues/1205, https://github.com/kyleconroy/sqlc/issues/1596).

This PR allows constructs such as 

```sql
SELECT foo.bar FROM unnest(baz) AS foo (bar);
```

and

```sql
SELECT foo.bar, foo.i::bigint FROM unnest(baz) WITH ORDINALITY AS foo (bar, i);
```

I haven't attempted to solve this problem in general - it will only work for a table alias with column alias(es), and data types are `any`, necessitating casts. This seems like a good first step, at least.

Fixes #1205


StevenACoffman:
I really appreciate your continued work on this area, as you appear to be fixing our biggest pain points! This and your work on #1884 are much appreciated, and I hope both can get reviewed and merged soon!

PR #2157, opened 2023-03-19
feat(sqlite) adding support for nullable types

aloisbarreras:


aloisbarreras:
Hey @kyleconroy friendly bump :) any thoughts on this?

laytan:
I would love something like this, also for the functions `sqlc.arg` and `sqlc.slice`, @aloisbarreras would you also add the cases for those functions if the approach is agreed with by @kyleconroy ?

PR #2215, opened 2023-04-18
Avoid some unnecessary buffering when reading config file

andrewmbenton:


golightlyb (changes requested):
Is this now missing a `defer file.Close()`? Best wishes

andrewmbenton:
Good catch, I've updated this PR with that addition.

kyleconroy (approved):

PR #2218, opened 2023-04-19
fix panic caused by invalid shorthand for `--no-remote` flag

andrewmbenton:


kyleconroy (approved):
We should add a test to make sure that the CLI works.

Issue #2179, opened 2023-03-30
[proposal] Support LOAD DATA LOCAL INFILE for MySQL


Jille:
https://github.com/go-sql-driver/mysql#load-data-local-infile-support supports LOAD DATA LOCAL INFILE with an io.Reader.

We could implement the :copyfrom query annotation to use LOAD DATA LOCAL INFILE when using go-sql-driver. The generated code would call https://pkg.go.dev/github.com/go-sql-driver/mysql#RegisterReaderHandler with a new reader for this call (and deregister it afterwards).

I propose to keep the same :copyfrom name and semantics, even though that's technically incorrect for MySQL.

I've looked at the three MySQL drivers listed on https://github.com/golang/go/wiki/SQLDrivers and the other two don't seem to support LOAD DATA [LOCAL] INFILE, though they might support it in the future.

We'd probably want to add a setting where people can indicate which MySQL driver they're using so we can generate the code using the relevant package.

Jille:
Some more thoughts:

* We need to encode the sqlc rows to TSV (or another CSV-like format) that is accepted by LOAD DATA INFILE. I propose to create an external library for that and let the generated code depend on it. (It felt like too much to generate inline.)

* We need to configure sqlc which mysql driver is used, because database/sql doesn't natively support LOAD DATA INFILE. I think `sql_package` is the right place to put that. Passing `go-sql-driver/mysql` would behave mostly the same as passing `database/sql`, except that it allows `:copyfrom` support

Jille:
I have two PRs ready for this, and a documentation branch I'll send later.

https://github.com/kyleconroy/sqlc/pull/2220
https://github.com/kyleconroy/sqlc/pull/2219
https://github.com/Jille/sqlc/compare/loaddata2...loaddata3

Jille:
I need go-sql-driver/mysql to expose the *time.Location to encode correctly.

@SnoozeThis https://github.com/go-sql-driver/mysql/issues/1416 || https://github.com/go-sql-driver/mysql/pull/1419

SnoozeThis:
(https://snoozeth.is/NFrLngXswx4) I will either wait until https://github.com/go-sql-driver/mysql/issues/1416 is closed or until https://github.com/go-sql-driver/mysql/pull/1419 is merged and then add a comment.

SnoozeThis:
An error occurred while snoozing: not found

PR #2219, opened 2023-04-20
Allow sql_package: "go-sql-driver/mysql"

Jille:
In preparation of #2179 for LOAD DATA LOCAL INFILE


internal/codegen/golang/driver.go
```diff
@@ -35,7 +39,18 @@ func (d SQLDriver) Package() string {
...
 }
+
+func (d SQLDriver) UsesStandard() bool {
+	switch d {
+	case SQLDriverPGXV4, SQLDriverPGXV5:
+		return false
+	default:
+		return true
+	}
+}
```
kyleconroy, line 56: Can we move this to another PR? This appears to be unused in this PR.

internal/codegen/golang/driver.go
```diff
@@ -3,14 +3,16 @@ package golang
 type SQLDriver int
 
 const (
-	SQLPackagePGXV4    string = "pgx/v4"
-	SQLPackagePGXV5    string = "pgx/v5"
-	SQLPackageStandard string = "database/sql"
+	SQLPackagePGXV4            string = "pgx/v4"
+	SQLPackagePGXV5            string = "pgx/v5"
```
kyleconroy, line 7: I'm starting to think we should have called this `jackc/pgx/v5` or `github.com/jackc/pgx/v5`. Happy to keep `go-sql-driver/mysql` for now.

kyleconroy (changes requested):
I think we want to add a new configuration option for this instead of re-using `sql_package`. The current value changes the package to use, either database/sql or pgx/v4 or pgxv5. In this case we're still using database/sql as a base, but using the mysql driver to do some additional work.

I think adding a new `sql_driver` override is the way to go and require it to be set to the full import path of `github.com/go-sql-driver/mysql`.

internal/codegen/golang/driver.go
Jille: Easy enough to start accepting the new values as well as the old values at some point, right?

internal/codegen/golang/driver.go
Jille: Good call, I didn't even need it in the next PR. Removed.

Jille:
Done

kyleconroy (approved):

PR #2220, opened 2023-04-20
:copyfrom for MySQL LOAD DATA INFILE

Jille:
Similar to PostgreSQL :copyfrom support, this adds support for MySQL's LOAD DATA INFILE.

Usage is the same as PostgreSQL: by writing an INSERT INTO query with :copyfrom, and we'll generate a method that internally calls LOAD DATA INFILE.

The only MySQL driver I found that supports LOAD DATA LOCAL INFILE at all is github.com/go-sql-driver/mysql. In able to support other drivers that might add support in the future, the `sql_package` config setting must be set to `go-sql-driver/mysql`.


kyleconroy:
@Jille I'm assuming this PR is still in draft because of some upstream mysql driver issues?

Jille:
Yeah, I need go-sql-driver/mysql to expose the *time.Location to encode correctly. But the maintainers don't seem to want to expose it in a way sqlc can use it.

How would you feel about an extra argument to the :loaddata methods that receives the *time.Location with the timezone configured for the connection? It's ugly, but I'm not sure I see another way yet.

kyleconroy:
> How would you feel about an extra argument to the :loaddata methods that receives the *time.Location with the timezone configured for the connection? It's ugly, but I'm not sure I see another way yet.

I don't think that will play nicely with the `query_parameter_limit` configuration option. We could create a separate struct that has a location as a field? Not sure honestly.

Jille:
We haven't found a way to expose the timezone from go-sql-driver/mysql yet. Let's move ahead with this and simply block `time.Time` values for now.

internal/codegen/golang/templates/go-sql-driver-mysql/copyfromCopy.tmpl
```diff
@@ -0,0 +1,45 @@
...
+// Use this in a transaction and use SHOW WARNINGS to check for any problems and roll back if you want to.
+// This is a MySQL limitation, not sqlc. Check the documentation for more information: https://dev.mysql.com/doc/refman/8.0/en/load-data.html#load-data-error-handling
+func (q *Queries) {{.MethodName}}(ctx context.Context{{if $.EmitMethodsWithDBArgument}}, db DBTX{{end}}, {{.Arg.SlicePair}}) (int64, error) {
+	pr, pw := io.Pipe()
+	defer pr.Close()
+	rh := fmt.Sprintf("{{.MethodName}}_%d", atomic.AddUint32(&readerHandlerSequenceFor{{.MethodName}}, 1))
+	mysql.RegisterReaderHandler(rh, func() io.Reader { return pr })
+	defer mysql.DeregisterReaderHandler(rh)
+	go convertRowsFor{{.MethodName}}(pw, {{.Arg.Name}})
+	result, err := {{if (not $.EmitMethodsWithDBArgument)}}q.{{end}}db.ExecContext(ctx, fmt.Sprintf("LOAD DATA LOCAL INFILE '%s' INTO TABLE {{.TableIdentifierForMySQL}} %s ({{range $index, $name := .Arg.ColumnNames}}{{if gt $index 0}}, {{end}}{{$name}}{{end}})", "Reader::" + rh, mysqltsv.Escaping))
```
kyleconroy, line 36: Let's add a comment here explaining that LOAD DATA can't be used with parameters and link to this part of the docs

```
The LOAD DATA INFILE statement reads rows from a text file into a table at a very high speed. The file name must be given as a literal string.
```

kyleconroy (changes requested):
A few small changes, but other than that I think this is ready to go. I wanted to generate static queries, but it turns out that `LOAD DATA` doesn't work with parameters. Maybe something we can do in the future.

kyleconroy:
Going to get this ready to merge in https://github.com/sqlc-dev/sqlc/pull/2545/. Thanks for the initial implementation :)

kyleconroy:
Closed via #2545

PR #2253, opened 2023-05-02
Update bufbuild plugin references

mfridman:
Small cleanup to the `buf.gen.yaml` plugin references:

- https://buf.build/community/planetscale-vtprotobuf
- https://buf.build/protocolbuffers/go

We've made some [changes to Remote Plugins](https://buf.build/blog/remote-packages-remote-plugins-approaching-v1) and figured it'd be nice to update references in popular open-source projects.


buf.gen.yaml
```diff
@@ -2,9 +2,9 @@ version: v1
 managed:
   enabled: true
 plugins:
-  - remote: buf.build/protocolbuffers/plugins/go:v1.28.1-1
+  - plugin: buf.build/protocolbuffers/go:v1.30.0
```
mfridman, line 5: 1. Newer versions of the buf CLI use the `plugin` key for both remote/local plugin references
2. I updated to the latest versions, although can also omit the versions so it always uses the latest:

```yaml
plugins:
  - plugin: buf.build/protocolbuffers/go
```

kyleconroy:
Thanks!

Issue #1881, opened 2022-10-05
sqlc.narg does not work for sqlite3


FrankFang:
### Version

1.15.0

### What happened?

```sql
-- name: UpdateChapter :one
UPDATE chapters
SET name = coalesce(sqlc.narg('name'), name),
    content = coalesce(sqlc.narg('content'), content)
WHERE id = sqlc.arg('id')
RETURNING *;
```
It does not work.

### Relevant log output

```shell
$ sqlc generate
line 26:29 no viable alternative at input ';\n\n-- name: UpdateChapter :one\nUPDATE chapters\nSET name = coalesce(sqlc.narg('
line 26:29 no viable alternative at input 'UPDATE chapters\nSET name = coalesce(sqlc.narg('
# package main
D:\Fang\repos\gobook\db\queries\chapters.sql:1:1: no viable alternative at input 'UPDATE chapters\nSET name = coalesce(sqlc.narg('
```

### Database schema

```sql
CREATE TABLE IF NOT EXISTS books (
  id   INTEGER PRIMARY KEY AUTOINCREMENT,
  name text    NOT NULL,
  author text,
  created_at DATETIME DEFAULT CURRENT_TIMESTAMP NOT NULL,
  updated_at DATETIME DEFAULT CURRENT_TIMESTAMP
);
CREATE TABLE IF NOT EXISTS chapters (
  id   INTEGER PRIMARY KEY AUTOINCREMENT,
  name text    NOT NULL,
  book_id INTEGER NOT NULL,
  content text,
  created_at DATETIME DEFAULT CURRENT_TIMESTAMP NOT NULL,
  updated_at DATETIME DEFAULT CURRENT_TIMESTAMP
);

ALTER TABLE books ADD summary text;

ALTER TABLE books ADD deleted_at DATETIME;

CREATE TABLE IF NOT EXISTS chapters (
  id INTEGER PRIMARY KEY AUTOINCREMENT,
  book_id INTEGER NOT NULL,
  name TEXT NOT NULL,
  content TEXT NOT NULL,
  created_at DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP,
  updated_at DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP,
  deleted_at DATETIME
);

ALTER TABLE chapters
ADD COLUMN parent_id
INTEGER DEFAULT 0
NOT NULL;

ALTER TABLE chapters
ADD COLUMN deleted_at DATETIME;
```

[...]

PR #2259, opened 2023-05-04
feat: improve support for sqlite named parameters

orisano:
sqlite supports the named parameter, but currently sqlc does not rewrite it as it does in postgresql.
Because named.IsParamSign judges whether named parameter is named or not by the content of ast.A_Expr, but sqlite3 converts from BindContext to ast.ParamRef and loses the name information.
In sqlite3, it is modified to return ast.A_Expr or ast.ParamRef depending on the contents of BindContext when converting.

And, parameter functions do not work in sqlite.
This is because names containing a dot, such as `sqlc.arg`, cause a syntax error.
So functions starting with `sqlc_` are interpreted as starting with `sqlc.` in sqlite.

related:
* https://github.com/kyleconroy/sqlc/issues/1881 
* https://github.com/kyleconroy/sqlc/pull/2157


kyleconroy:
Closed via https://github.com/kyleconroy/sqlc/pull/2274

Issue #263, opened 2020-01-12
Incorrect line and column numbers when a query is missing a semicolon


kevinburke1:
If you forget to end a query with a semicolon you get:

```sql
-- Example queries for sqlc
CREATE TABLE authors (
  id   BIGSERIAL PRIMARY KEY,
  name text      NOT NULL,
  bio  text
);

-- name: GetAuthor :one
SELECT * FROM authors
WHERE id = $1 LIMIT 1;

-- name: ListAuthors :many
SELECT * FROM authors
ORDER BY name

-- name: CreateAuthor :one
INSERT INTO authors (
  name, bio
) VALUES (
  $1, $2
)
RETURNING *;

-- name: DeleteAuthor :exec
DELETE FROM authors
WHERE id = $1;
```

```
# package querytest
query.sql:1:1: syntax error at or near "INSERT"
```

Which is confusing because I start looking at the next query instead of the one above it, and the `1:1` makes me think it's at the beginning of the file which it's not.

Maybe just append this error message with "add a semicolon"?

cmoog:
@kevinburkemeter For which database engine?

kevinburkemeter:
Ah, Postgres

kyleconroy:
We can't provide this information right now due to the lack of structured error messages being returned from pg_query_go https://github.com/lfittl/pg_query_go/issues/24

dskarataev:
yes please! it is very confusing. I would NEVER realize what is the root cause of the error if I would not find this issue on GitHub...

So @kyleconroy would be really good if you would still find some way to add the structured error messages

videate-karim:
pg_query_go was moved. here is the correct link: https://github.com/pganalyze/pg_query_go/issues/24

I looked at this, @kyleconroy - is there any reason we can't track the line number of the source file? In the postgresql `Parse()` function, when `p.Parse()` returns err, couldn't that be emitted and/or returned with the line of source file?

I am looking here: https://github.com/kyleconroy/sqlc/blob/e2c60566caec2ce7276824e46da792cd98fa05c3/internal/engine/postgresql/parse.go#L165

kyleconroy:
> is there any reason we can't track the line number of the source file? 

Yes. The error returned from `pg_query.Parse` does not include a line number nor location. Parse is called once on the contents of a file. This won't be fixed until pg_query_go returns error with this information.

StevenACoffman:
[...]

PR #2289, opened 2023-05-28
Feat: PostgreSQL capture correct line and column numbers for parse error

StevenACoffman:
Fixes issue #263

Recently pg_query_go began to return the full structured error info instead of just error message in https://github.com/pganalyze/pg_query_go/pull/76 which has released this functionality in [v4.2.1](https://github.com/pganalyze/pg_query_go/releases/tag/v4.2.1). This sqlc PR #2288 updates to use v4.2.1 for pg_query_go.

After that PR is merged, sqlc will still not correctly display the line and column numbers, as pg_query_go  `parser.Error` type has an `Error()` method that **only** includes the **message**, even though that same `parser.Error` type has all the other fields public.

**_This_** PR attempts to normalize the pg_query_go  `parser.Error` type  to a sqlc `*sqlerr.Error`.

+ Parse [here](https://github.com/pganalyze/pg_query_go/blob/main/parser/parser.go#LL106C15-L106C38)
Appears to call:
+ ParseToProtobuf [here](https://github.com/pganalyze/pg_query_go/blob/910915bfda5eb7b51f330613c335842ecd0a67d0/parser/parser.go#L113)
+ Appears to call  https://github.com/pganalyze/pg_query_go/blob/7986455fd7e1752b8399930d49beac2e8a139bd5/parser/pg_query_parse.c#L110
+ Appears to call pg_query_parse_protobuf [here](
https://github.com/pganalyze/pg_query_go/blob/7986455fd7e1752b8399930d49beac2e8a139bd5/parser/pg_query_parse.c#L110)

So it looks like a PgQueryError error has:
```
		error = malloc(sizeof(PgQueryError));
		error->message   = strdup(error_data->message);
		error->filename  = strdup(error_data->filename);
		error->funcname  = strdup(error_data->funcname);
		error->context   = NULL;
		error->lineno    = error_data->lineno;
		error->cursorpos = error_data->cursorpos;

		result.error = error;	
```
Which then is read by 
```
func newPgQueryError(errC *C.PgQueryError) *Error {
	err := &Error{
		Message:   C.GoString(errC.message),
		Lineno:    int(errC.lineno),
		Cursorpos: int(errC.cursorpos),
	}
	if errC.funcname != nil {
		err.Funcname = C.GoString(errC.funcname)
	}
	if errC.filename != nil {
		err.Filename = C.GoString(errC.filename)
	}
	if errC.context != nil {
		err.Context = C.GoString(errC.context)
	}
	return err
}
```

So the pg_query_go `*parser.Error` should be fully populated:
```
type Error struct {
	Message   string // exception message
	Funcname  string // source function of exception (e.g. SearchSysCache)
	Filename  string // source of exception (e.g. parse.l)
	Lineno    int    // source of exception (e.g. 104)
	Cursorpos int    // char in query at which exception occurred
	Context   string // additional context (optional, can be NULL)
}
```

Which is not exactly the same as the sqlc `*sqlerror.Error` type:
```
type Error struct {
	Err      error
	Code     string
	Message  string
	Location int
	Line     int
	Column   int
	// Hint     string
}
```

I believe that:
+ pg_query_go`*parser.Error.Lineno` is equivalent to sqlc `*sqlerror.Error.Line`
+ pg_query_go `*parser.Error.Cursorpos` is equivalent to sqlc `*sqlerror.Error.Location`
+ pg_query_go `*parser.Error.Message` is equivalent to sqlc `*sqlerror.Error.Message`

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,7 +14,7 @@ require (
 	github.com/jinzhu/inflection v1.0.0
 	github.com/lib/pq v1.10.9
 	github.com/mattn/go-sqlite3 v1.14.16
-	github.com/pganalyze/pg_query_go/v4 v4.2.0
+	github.com/pganalyze/pg_query_go/v4 v4.2.1
 	github.com/riza-io/grpc-go v0.2.0
 	github.com/spf13/cobra v1.7.0
 	github.com/spf13/pflag v1.0.5
--- a/go.sum
+++ b/go.sum
@@ -118,8 +118,8 @@ github.com/mattn/go-isatty v0.0.12/go.mod h1:cbi8OIDigv2wuxKPP5vlRcQ1OAZbq2CE4Ky
 github.com/mattn/go-sqlite3 v1.14.16 h1:yOQRA0RpS5PFz/oikGwBEqvAWhWg5ufRz4ETLjwpU1Y=
 github.com/mattn/go-sqlite3 v1.14.16/go.mod h1:2eHXhiwb8IkHr+BDWZGa96P6+rkvnG63S2DGjv9HUNg=
 github.com/niemeyer/pretty v0.0.0-20200227124842-a10e7caefd8e h1:fD57ERR4JtEqsWbfPhv4DMiApHyliiK5xCTNVSPiaAs=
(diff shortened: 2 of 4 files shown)


kyleconroy (approved):
Thank you! This fixes a really old annoying issue with misleading line numbers.

Jille:
Thank you indeed! That sure was annoying.

PR #2250, opened 2023-05-02
feat: to allow spaces between function name and arguments of functions to be rewritten

orisano:
close #2130

PR #2298, opened 2023-06-06
fix(engine/sqlite): add location info to sqlite ast

orisano:
Sorry, #2250 will panic if Location is not set properly.
Fixed sqlite convert because many nodes don't have Location set properly.


kyleconroy:
Thanks for the quick fix! We should also probably make sure that we never pass -1 to strings.Repeat

PR #2302, opened 2023-06-07
Local templates

jwc-clinnection:
This pull request enables a user specified local template directory that allows templates to be customized. This feature adds a new command line argument "inittmpl" that copies the embedded templates locally. To use this feature:

- Add "template_path" to sqlc.yaml. For example: template_path: user_templates
- Run: sqlc inittmpl
- Edit the copied templates
- Run: sqlc generate


kyleconroy:
Okay, some good news and bad news. I'll start with the bad news. I'm unlikely to merge this change, as I don't want to support user-defined templates for code generation. The reason is maintainability and produce surface area. While templates are used to generate Go code today, I'd like to eventually move to a pure AST approach. Maintaining the current set of templates is difficult. Just take a look at a few of the codegen templates and you can see what I'm talking about.

I promised some good news though. I'd love to help you get this working as a plugin. This type of feature should be perfect for it and will give users complete flexibility. I think it would start as a process-based plugin and then could be ported to WASM/WASI once Go 1.21 comes out. Our plugin docs aren't great right now, hence the offer to help make it a plugin.

Let me know if you're interested.

jwc-clinnection:
Yes, I am very interested. Please advise as to how to proceed. Thank you.

jwc-clinnection:
I created a pristine fork

fdietze:
Was this plugin ever created?

PR #2307, opened 2023-06-07
fix: Let column names not result in invalid Go field names

kyleconroy:
Column names are less restricted than Go struct field names, which should be proper Go identifiers. Hence, column names cannot be translated directly into field names. This commit tries to address this.

Characters in column names that do not qualify as a Go identifier are stripped. The title casing as applied to underscores is applied to stripped characters.

Replaces https://github.com/kyleconroy/sqlc/pull/1840


andrewmbenton:
Is generating testdata with sqlc v1.15.0 intentional?

PR #2309, opened 2023-06-07
feat: add support for Postgres multi dimensional arrays in Go

kyleconroy:
This PR adds support for generating model fields from multi-dimensional postgresql array columns. The approach here is to pass around an additional column attribute array_bounds representing the dimensions of the postgres array column. When generating code, the array_bounds is checked along is_array to determine the dimensions of the resulting struct field. Would be happy to hear alternative approaches or solutions here as well. More info: [Postgres Arrays](https://www.postgresql.org/docs/14/arrays.html)

Replaces https://github.com/kyleconroy/sqlc/pull/1651 https://github.com/kyleconroy/sqlc/pull/2308


colli173:
I don't have push rights to this branch @kyleconroy, so I updated the original branch and made a new PR https://github.com/kyleconroy/sqlc/pull/2338 with the fixes.

PR #2310, opened 2023-06-07
codegen: Use catalog's default schema

kyleconroy:
Remove references to hard-coded "public" schema.

Also update some tests added in https://github.com/kyleconroy/sqlc/pull/1986


andrewmbenton (approved):
:+1:

PR #2319, opened 2023-06-09
build: Configure dependencies via devenv.sh

kyleconroy:
Create a consistent development environment for sqlc using https://devenv.sh. This setup is completely optional.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,10 @@
 /.idea/
 __pycache__
 .DS_Store
+
+# Devenv
+.envrc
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -1,3 +1,10 @@
 /.idea/
 __pycache__
 .DS_Store
+
+# Devenv
+.envrc
```
kyleconroy, line 6: I don't want to enable automatic entry into devenv.sh as it requires quite a bit of setup locally.

PR #2325, opened 2023-06-14
Add json_tags_id_camelcase configuration option

jwc-clinnection:
If json_tags_id_camelcase is true, "ID" in json tags will be camelcase. If false, will be uppercase. Defaults to `false`

Example where false:
```
type Example struct {
	ID           uuid.UUID  `json:"id"`
	Name         string     `json:"name"`
	CreatedAt    time.Time  `json:"createdAt"`
	UpdatedAt    time.Time  `json:"updatedAt"`
	CreatedByID  *uuid.UUID `json:"createdByID"`
	UpdatedByID  *uuid.UUID `json:"updatedByID"`
}
```

Example where true:
```
type Example struct {
	ID           uuid.UUID  `json:"id"`
	Name         string     `json:"name"`
	CreatedAt    time.Time  `json:"createdAt"`
	UpdatedAt    time.Time  `json:"updatedAt"`
	CreatedByID  *uuid.UUID `json:"createdById"`
	UpdatedByID  *uuid.UUID `json:"updatedById"`
}
```

### DIFF
--- a/docs/reference/config.md
+++ b/docs/reference/config.md
@@ -111,6 +111,8 @@ The `gen` mapping supports the following keys:
 - `emit_all_enum_values`:
   - If true, emit a function per enum type
     that returns all valid enum values.
+- `json_tags_id_camelcase`:
+  - If true, "ID" in json tags will be camelcase. If false, will be uppercase. Defaults to `false`
 - `json_tags_case_style`:
   - `camel` for camelCase, `pascal` for PascalCase, `snake` for snake_case or `none` to use the column name in the DB. Defaults to `none`.
 - `output_batch_file_name`:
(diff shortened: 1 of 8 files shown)


Killian-Grieg:
I need this so much! Thank you

kyleconroy:
Discussed this a bit with Andrew, I think we want to go ahead and making a breaking change here. By default, I think we want `updated_by_id` to map to `updatedById`. Let's rename the configuration option to be `json_tags_id_uppercase` and output `updatedByID` when that's true.

jwc-clinnection:
> Discussed this a bit with Andrew, I think we want to go ahead and making a breaking change here. By default, I think we want `updated_by_id` to map to `updatedById`. Let's rename the configuration option to be `json_tags_id_uppercase` and output `updatedByID` when that's true.

I have pushed the requested changes.

PR #2329, opened 2023-06-15
Emit query method arg params struct

jwc-clinnection:
Add configuration option "emit_methods_arg_with_struct" where if true, generated methods will always have struct argument even if number of arguments is 1. Defaults to `false`.

Example when emit_methods_arg_with_struct: false (default behavior):
```
const createBlog = `-- name: CreateBlog :one
insert into blog (name)
values ($1)
RETURNING id, name
`

func (q *Queries) CreateBlog(ctx context.Context, name string) (Blog, error) {
	row := q.db.QueryRow(ctx, createBlog, name)
	var i Blog
	err := row.Scan(&i.ID, &i.Name)
	return i, err
}
```

Example when emit_methods_arg_with_struct: true
```
const createBlog = `-- name: CreateBlog :one
insert into blog (name)
values ($1)
RETURNING id, name
`

type CreateBlogParams struct {
	Name string `json:"name"`
}

func (q *Queries) CreateBlog(ctx context.Context, arg CreateBlogParams) (Blog, error) {
	row := q.db.QueryRow(ctx, createBlog, arg.Name)
	var i Blog
	err := row.Scan(&i.ID, &i.Name)
	return i, err
}
```


kyleconroy:
As usual, thanks for the pull request. This option isn't needed because we already have this functionality using the `query_parameter_limit` option. Setting this to `0` will always generate a struct.

https://play.sqlc.dev/p/e95df5c9a6b538ce5d55846d9dc0c1e4801521c8c4e987418863a34daeb86e57

PR #2332, opened 2023-06-16
fix: MySQL slice shadowing database/sql import

LudovicDebost:
The generated code use a var named `sql` but when the query return a `database/sql` type like `sql.NullString` then the var is shadowing the import.

Fixes #1816


kyleconroy:
We'll have the same issue is someone uses a type in a `query` package, but that's not going to happen nearly as much as `sql`.

PR #2335, opened 2023-06-17
improve init command to create v2/v1 yaml file

dangquangdon:
`sqlc init` now would generate v2 format yaml file by default, user also can use `--v1` to generate v1 format if needed.

```
Create an empty sqlc.yaml settings file

Usage:
  sqlc init [flags]

Flags:
  -h, --help   help for init
      --v1     generate v1 config yaml file
      --v2     generate v2 config yaml file (default true)
```


kyleconroy:
Thanks! This has been on the todo list for a while.

Issue #710, opened 2020-09-11
Unused struct is being generated into models.go


yeka:
sqlc.yaml
```
version: "1"
packages:
  - name: "authors"
    path: "generated/repo/authors"
    queries: "./files/queries/authors.sql"
    schema: "./files/schema/"
    engine: "postgresql"
    emit_json_tags: true
    emit_prepared_queries: false
    emit_interface: false
    emit_exact_table_names: false
  - name: "books"
    path: "generated/repo/books"
    queries: "./files/queries/books.sql"
    schema: "./files/schema/"
    engine: "postgresql"
    emit_json_tags: true
    emit_prepared_queries: false
    emit_interface: false
    emit_exact_table_names: false
```

The `./files/schema/` directory contains [golang-migrate](https://github.com/golang-migrate/migrate) files.

`000001_authors.up.sql`
```
CREATE TABLE authors (
    id   BIGSERIAL PRIMARY KEY,
    name text      NOT NULL,
    bio  text
);
```

`000002_books.up.sql`
```
CREATE TABLE books (
    id SERIAL PRIMARY KEY,
    title VARCHAR(120) NOT NULL
);
```

And directory `./files/queries` contains sql queries for `sqlc`.

`authors.sql`
```
-- name: ListAuthors :many
SELECT * FROM authors
ORDER BY name;
```

`books.sql`
```
-- name: ListBooks :many
SELECT * FROM books;
```

When I run `sqlc generate`, it will generate `./generated/repo/authors/` and `./generated/repo/books` as expected.
When I see the `./generated/repo/authors/models.go` and `./generated/repo/books/models.go`, both have 2 struct types which is `Author` and `Book`.

`models.go`
```
// Code generated by sqlc. DO NOT EDIT.

package books

import (
	"database/sql"
)

type Author struct {
	ID   int64          `json:"id"`
	Name string         `json:"name"`
	Bio  sql.NullString `json:"bio"`
}

type Book struct {
	ID    int32  `json:"id"`
	Title string `json:"title"`
}
```

In `books` package, `Author` type in unused, and in `authors` package `Book` type is unused.
I assumed it's generated from database schema because both package use the same schema in `sqlc.yaml`.
Is it possible not to generate unused struct in `models.go`?

kyleconroy:
> Is it possible not to generate unused struct in models.go?

Yes, we could change the code generation to not output models for unused tables. Note that you still may run into issues if your a query in `books.sql` returns results from the authors table.

[...]

PR #2336, opened 2023-06-17
Feature filter unused structs in models

yank0vy3rdna:
Added config option `emit_filter_unused_models`.
If true, add only used by queries models to models.go.

https://github.com/kyleconroy/sqlc/issues/710


andrewmbenton:
Thanks! As you know this has been a feature request for awhile.

I'm not sure how Kyle wants to handle this but at the very least, in order to merge, we would need to address the failing test and write at least one new test that covers usage of this new config option.

andrewmbenton:
There are a few issues with this implementation, so I took a crack at one of my own in https://github.com/kyleconroy/sqlc/pull/2369. It should behave similarly to this one, but covers a few more edge cases.

I'll close this PR and we can continue discussion on the new one.

Issue #2316, opened 2023-06-08
Malformed imports for array overrides


aitva:
### Version

1.18.0

### What happened?

I tried to upgrade to sqlc v1.18.0 and the Go compiler reports `malformed import path` in the generated sources. I am using column overrides for array types and the generated code keeps the `[]` the import URL.

Here is the code that was working in v1.17.2: https://play.sqlc.dev/p/17e05c48de88123b99ffff7b585291a45f7db6099cbba3627888c1f238f6c6fe

Here is the same code in v1.18.0: https://play.sqlc.dev/p/5c3833f312ae1596952586f263eecc4ce56b2736aa45e90a1441e04b6a99b8db

If you look at the generated `model.go`, you will notice the `"database/[]sql"` import that produces the error.

### Playground URL

https://play.sqlc.dev/p/5c3833f312ae1596952586f263eecc4ce56b2736aa45e90a1441e04b6a99b8db

### What operating system are you using?

Linux

### What database engines are you using?

PostgreSQL

### What type of code are you generating?

Go

andrewmbenton:
The fact that an override configuration like `"go_type": "database/[]sql.NullInt64"` ever worked is a little surprising. We just merged a change I made to add a `"slice"` boolean property to the object definition format for the override `"go_type"` property. I also added some documentation to the bottom of the Type Overriding section here: https://docs.sqlc.dev/en/latest/reference/config.html#type-overriding

Basically you just need to do something like the following in your overrides list:
```json
{
  "column": "authors.tags",
  "go_type": {
    "import": "database/sql",
    "type": "NullInt64",
    "slice": true
}
```

This should work with a build off of the main branch, and it will be in the next release.

aitva:
Thank you @andrewmbenton. I will wait for the next release to upgrade. :pray:

PR #2339, opened 2023-06-18
internal/config: support golang overrides for slices

andrewmbenton:
While investigating a fix for https://github.com/kyleconroy/sqlc/issues/2316 I discovered some generally unexpected behavior when attempting to get sqlc to generate slices of types in overrides for Go code gen. So I added a boolean `"slice"` property to the `"go_type"` property of the overrides object array, which behaves similarly to the `"pointer"` property.

This doesn't "fix" the above-referenced issue per se, but it does provide a more explicit way to tell sqlc how to achieve the desired result.

The docs I wrote may be wrong, so feel free to advise on any changes and I'll incorporate them.


internal/codegen/golang/imports.go
```diff
@@ -462,15 +460,9 @@ func (i *importer) batchImports() fileImports {
...
-	v = strings.TrimPrefix(v, "[]")
-	v = strings.TrimPrefix(v, "*")
-	return v
-}
-
 func hasPrefixIgnoringSliceAndPointerPrefix(s, prefix string) bool {
-	trimmedS := trimSliceAndPointerPrefix(s)
-	trimmedPrefix := trimSliceAndPointerPrefix(prefix)
+	trimmedS := strings.TrimLeft(s, "*[]")
+	trimmedPrefix := strings.TrimLeft(prefix, "*[]")
```
kyleconroy, line 465: Do we validate these prefixes anywhere? With the change to `TrimLeft`, I think a string like this `****[[]]]database/foo` will work?

andrewmbenton: We do not validate these anywhere as far as I can see. You are right about this change trimming more than is perhaps desirable, so I'll revert.

kyleconroy (approved):

Issue #2061, opened 2023-02-11
Support dynamic order by clause


go-aegian:
### What do you want to change?

the order by clause in the :many definition query should be set dynamically 

```
-- name: ListOrders :many
  select * from order where status = $1 order by @orderBy::text;
```
**Generated**

`const listOrders = "select * from order where status = $1 order by @orderBy::text"
`
 ```
type ListOrdersParams struct {
       Status string
       OrderBy string
}
```

```
func (q *Queries) ListOrders(ctx context.Context, arg ListOrdersParams) ([]*ListOrdersRow, error) {
        orderBy:=arg.OrderBy
        if orderBy=="" {
           orderBy= "0"
        }
        listOrdersWithOrderBy = strings.Replace(listOrders, "@orderBy::text", orderBy, -1)

	rows, err := q.db.Query(ctx, listOrdersWithOrderBy, arg.Status)

...

```

### What database engines need to be changed?

PostgreSQL, MySQL

### What programming language backends need to be changed?

Go, Python, Kotlin

jwc-clinnection:
Thank you!

jwc-clinnection:
I have implemented this in #2343

danielbraun89:
Any update on this?

kianooshaz:
Any update on this?

EfogDev:
Any update on this?

ckucukjc:
Any update on this?

ptman:
https://github.com/sqlc-dev/sqlc/discussions/364

Issue #2060, opened 2023-02-11
Support dynamic where clauses


go-aegian:
### What do you want to change?

apart from the predefined where clause, it needs somehow to be able to support passing a string containing a filter clause that it's parameters are already resolved prior to the Query function, I proposed a possible solution to be able to declare in the sql definition something like this 

```
-- name ListOrders :many
  select * from order where status = $1 and (@filter::text);

```
Generated 

```const listOrders = "select * from order where status = $1 and (@filter::text);"```

```
 type ListOrderParams struct {
       Status string
       Filter string
}
```

```
func (q *Queries) ListOrders(ctx context.Context, arg ListOrderParams) ([]*OrderRow, error) {
        filter:=arg.Filter
        if filter=="" {
           filter = "1=1"
        }
        listOrdersWithFilter = strings.Replace(listOrders, "@filter::text", filter, -1)

	rows, err := q.db.Query(ctx, listOrdersWithFilter, arg.Status)

...
```

### What database engines need to be changed?

PostgreSQL, MySQL

### What programming language backends need to be changed?

Go, Python, Kotlin

**Comment by jwc-clinnection** [comment]
I have implemented this in #2343

**Comment by go-aegian** [comment]
Thank you @jwc-clinnection once the release is available I'll check it out

**Comment by jwc-clinnection** [comment]
It is failing some tests. For example this is an expected change given the pull request:
```
                -               rows, err := q.db.Query(ctx, joinTextArray)
                +               sql := joinTextArray
                +       
                +               rows, err := q.db.Query(ctx, sql)

```
I tried:
```
make build-endtoend
cd ./internal/endtoend/testdata && go build ./...
```
But, I still get errors when I run:
```
make test
```

Should I be doing something to regenerate the test samples? If so, may I have some guidance on the procedure?

jwc-clinnection:
> Thank you @jwc-clinnection once the release is available I'll check it out

@go-aegian Can you give any direction? Thank you.

go-aegian:
> 

@jwc-clinnection unfortunately I can't help you on this as I am not familiar how is it compiled, tested. That is why I wait for a released version.

jwc-clinnection:
@kyleconroy The tests that this is failing are expected given the changes. Can you give me advice as to how to proceed to correct the tests?

iancharters:
Any movement on this?  I would really like to be able to leverage this in our codebases that use sqlc.

go-aegian:
[...]

PR #2343, opened 2023-06-20
Dynamic query

jwc-clinnection:
Implemented dynamic query as suggested by [go-aegian](https://github.com/go-aegian) in issues:

[ Support dynamic where clauses #2060 ](https://github.com/kyleconroy/sqlc/issues/2060)
[Support dynamic order by clause #2061](https://github.com/kyleconroy/sqlc/issues/2061) 

Given the desire to move away from templates, I suspect that this would be a temporary fix but a much desired one.

### DIFF
--- a/internal/codegen/golang/query.go
+++ b/internal/codegen/golang/query.go
@@ -120,7 +120,7 @@ func (v QueryValue) Params() string {
 			} else if !v.EmitStruct() && v.IsStruct() {
 				out = append(out, toLowerCase(f.Name))
 			} else {
-				if !f.Column.IsFilter {
+				if !f.Column.IsFilter && !f.Column.IsOrderBy {
 					out = append(out, v.Name+"."+f.Name)
 				}
 			}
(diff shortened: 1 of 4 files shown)


jwc-clinnection:
It is failing some tests. For example this is an expected change given the pull request:
```
                -               rows, err := q.db.Query(ctx, joinTextArray)
                +               sql := joinTextArray
                +       
                +               rows, err := q.db.Query(ctx, sql)

```
I tried:
```
make build-endtoend
cd ./internal/endtoend/testdata && go build ./...
```
But, I still get errors when I run:
```
make test
```

Should I be doing something to regenerate the test samples? If so, may I have some guidance on the procedure?

jwc-clinnection:
We have discovered a bug in this pull request. If "query_parameter_limit: 0" is not set, then it fails to generate the correct filter code. We are working to to resolve this problem.

jwc-clinnection:
We have corrected the issue, and it now generates correct code for all cases of query_parameter_limit

go-aegian:
@jwc-clinnection have you got any advice from @kyleconroy on how to move this forward to the release?

In the meantime I did manage a workaround process until your work is released though.

andrewmbenton:
There is still an active discussion regarding `sqlc`'s path forward for dynamic SQL, so I'm closing this for now until that discussion is resolved. Follow along with issues [#2061](https://github.com/sqlc-dev/sqlc/issues/2061) and [#2060](https://github.com/sqlc-dev/sqlc/issues/2060) or this discussion: https://github.com/sqlc-dev/sqlc/discussions/364

PR #2344, opened 2023-06-20
cmd/sqlc: Add the vet subcommand

kyleconroy:
## vet

This is a proposal to add a new subcommand that runs queries through a set of lint rules.

### Rules

Rules are defined in the configuration file. They consist of a name, message, and an expression. If the expression evaluates to `true`, an error is reported. These expressions are evaluated using [cel-go](https://github.com/google/cel-go).

Each expression has access to a query object, which is defined as the following struct:

```proto
message Config
{
  string version = 1;
  string engine = 2 ;
  repeated string schema = 3;
  repeated string queries = 4;
}

message Query
{
  // SQL body
  string sql = 1;
  // Name of the query
  string name = 2; 
  // One of :many, :one, :exec, etc.
  string cmd = 3;
  // Query parameters, if any
  repeated Parameter params = 4;
}


message Parameter
{
  int32 number = 1;
}
```

This struct would be expanded in the future to include more query information. We may also add information from a running database, such as the result from `EXPLAIN`.

### Example

While these examples are simplistic, they give you an idea on what types of rules you can write.

```yaml
version: 2
sql:
  - schema: "query.sql"
    queries: "query.sql"
    engine: "postgresql"
    gen:
      go:
        package: "authors"
        out: "db"
    rules:
      - no-pg
      - no-delete
      - only-one-param
      - no-exec
rules:
  - name: no-pg
    message: "invalid engine: postgresql"
    rule: |
      config.engine == "postgresql"
  - name: no-delete
    message: "don't use delete statements"
    rule: |
      query.sql.contains("DELETE")
  - name: only-one-param
    message: "too many parameters"
    rule: |
      query.params.size() > 1
  - name: no-exec
    message: "don't use exec"
    rule: |
      query.cmd == "exec"
```


go.mod
```diff
@@ -24,6 +25,8 @@ require (
 	gopkg.in/yaml.v3 v3.0.1
 )
 
+require github.com/stoewer/go-strcase v1.2.0 // indirect
```
andrewmbenton, line 28: This should have ended up in the longer list of indirect dependencies below right?

internal/cmd/vet.go
```diff
@@ -0,0 +1,189 @@
...
+	)
+	if err != nil {
+		return fmt.Errorf("new env; %s", err)
+	}
+
+	checks := map[string]cel.Program{}
+	msgs := map[string]string{}
+
+	for _, c := range conf.Rules {
+		if c.Name == "" {
```
andrewmbenton, line 81: Did you consider putting some of this basic validation inside of a shared config parse/validation area? Not sure if config.Validate() would be the right spot, or somewhere else, or maybe this idea just isn't correct if you think each subcommand should be responsible for validating its own "section" of the config (if it has one)?

go.mod
kyleconroy: Not sure honestly. I'll make sure to run `go mod tidy` as that will put it in the right spot.

PR #2349, opened 2023-06-21
Support unbounded parameter limit

deryrahman:
In version 2, setting up `query_parameter_limit` to `-1` is not working. As per explanation in [this reference](https://github.com/kyleconroy/sqlc/blob/8d365df161778947e36394838727ce68e40d1f69/docs/reference/config.md?plain=1#L127), it should accepts limit of positional arguments with value `-1`.

And I think there should be a correction of what value `-1` must behave. Instead of always emit struct (like value `1` does), it should generates the code with complete positional arguments.

### DIFF
--- a/docs/reference/config.md
+++ b/docs/reference/config.md
@@ -124,8 +124,8 @@ The `gen` mapping supports the following keys:
 - `output_files_suffix`:
   - If specified the suffix will be added to the name of the generated files.
 - `query_parameter_limit`:
-  - Positional arguments that will be generated in Go functions (>= `1` or `-1`). To always emit a parameter struct, you would need to set it to `-1`. `0` is invalid. Defaults to `1`.
-`rename`:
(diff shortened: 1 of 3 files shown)


kyleconroy:
`0` is actually a valid setting. When zero is set, queries will always have a struct. Can we update this to only make changes for `-1`?

deryrahman:
hi @kyleconroy , sure. Just want to clarify, so what is the definition of `query_parameter_limit` ? In my interpretation, it's used to set the maximum number of parameters that would be generated.

So, set the query parameter limit to `n`, where `n` > 0, if the number of parameter is `p` and `p`<=`n`, then, it uses data type per individual parameter, otherwise (`p`>`n`), it uses a struct.

Set to `-1` to make it always uses data type per individual parameter. And `0` would be logically invalid I believe(?)

deryrahman:
Just seeing your latest fix. Understood now, sqlc doesn't support always emit all positional arguments (`-1`) as I expected (though I've just realized that emitting all args is bad practice). Sorry for misinterpretation. Agree to close this PR 👍

PR #2361, opened 2023-06-22
A few small staticcheck fixes

Jille:


kyleconroy:
How are you running staticcheck? I'd like to run it as part of CI

Jille:
Simply `staticcheck ./...`. There's still a few findings that I didn't fix, they should be fixed before enabling it in CI.

PR #2362, opened 2023-06-22
Don't defer rows.Close() if pgx.BatchResults.Query() failed 

Jille:
To appease staticcheck


Jille:
cc @chrisyxlee I think you introduced this in 0ebe855870753129ab619c4c82b9dc7391c088c5

chrisyxlee:
thanks for catching!

Issue #2131, opened 2023-03-06
Unable to pass in variable for limit in update queries


bbody:
### Version

1.16.0

### What happened?

I want to allow my application to limit how many records get updated but SQLC does not expose it in the code. I don't have the option to use a subquery with a limit (the code generation works) due to my MySQL version.

### Relevant log output

```shell
const test = `-- name: Test :exec
UPDATE authors
SET bio="test"
WHERE name IS NULL
LIMIT ?
`

func (q *Queries) Test(ctx context.Context) error {
	_, err := q.db.ExecContext(ctx, test)
	return err
}
```

### Database schema

```sql
CREATE TABLE authors (
  id   INT PRIMARY KEY,
  name text      NOT NULL,
  bio  text
);
```

### SQL queries

```sql
-- name: Test :exec
UPDATE authors
SET bio="hello"
WHERE name IS NULL
LIMIT ?;
```

### Configuration

```yaml
{
  "version": "1",
  "packages": [
    {
      "path": "db",
      "engine": "mysql",
      "schema": "query.sql",
      "queries": "query.sql"
    }
  ]
}
```

### Playground URL

https://play.sqlc.dev/p/732de1ddc031aba2c5bfce20367a16435eedd43745324af4249b2b6803862430

### What operating system are you using?

macOS

### What database engines are you using?

MySQL

### What type of code are you generating?

Go

Jille:
As a workaround you can use `sqlc.arg(limit)` instead of `?`.

Jille:
Kyle added this to the milestone, and I'd like to have the next release, so I wrote a PR :P

@SnoozeThis https://github.com/kyleconroy/sqlc/pull/2365 -> close

SnoozeThis:
I don't have privileges to execute this action. Please install the SnoozeThis GitHub app (https://github.com/apps/snoozethis).

Jille:
@SnoozeThis #2365

SnoozeThis:
(https://snoozeth.is/j3WmMD-oYu4) I will wait until https://github.com/kyleconroy/sqlc/issues/2365 is closed and then add a comment.

SnoozeThis:
Resolved.

bbody:
Thank you @kyleconroy

PR #2365, opened 2023-06-24
Support "LIMIT ?" in UPDATE and DELETE for MySQL

Jille:
PostgreSQL doesn't support UPDATE-LIMIT

issue #2131


kyleconroy:
Fixes #2131

Issue #2277, opened 2023-05-12
Support for naming hints via sqlc.arg in BETWEEN expressions


dnwe:
### What do you want to change?

Currently if you use a `BETWEEN` expression such as `SELECT id, name, date FROM events WHERE date BETWEEN ? AND ?;` then sqlc simply names the struct fields as `Date` and `Date_2` a'la:
```go
type GetEventByDateParams struct {
	Date   sql.NullTime
	Date_2 sql.NullTime
}
```

It would be great if naming hints could be passed to sqlc as-in `LIKE` expressions so you could put `SELECT id, name, date FROM events WHERE date BETWEEN sqlc.arg(from_date) AND sqlc.arg(to_date);` and that would result in the generated struct fields being `FromDate` and `ToDate`. At the moment if you attempt to do that it breaks the parser and the Params struct is omitted entirely ([playground link](https://play.sqlc.dev/p/7ecf294cb7aa0b3ce1025657e091cce22970d3b4f5c4eb2baa578e5083cd4219))

### What database engines need to be changed?

PostgreSQL, MySQL

### What programming language backends need to be changed?

Go

Jille:
Kyle added this to the milestone, and I'd like to have the next release, so I wrote a PR :P

The change was small, but finding where to make it was harder :)

@SnoozeThis https://github.com/kyleconroy/sqlc/pull/2366

SnoozeThis:
(https://snoozeth.is/E8ZK_PmKi-g) I will wait until https://github.com/kyleconroy/sqlc/pull/2366 is merged and then add a comment.

Jille:
I just realized I misread your feature request. I thought you wanted to have better default names than Date and Date_2. However, I made a drive-by fix that fixed reusing the same parameter number for both left and right and that probably fixed your actual request.

I've added a test case for your request to my PR.

dnwe:
@Jille perfect, testing with sqlc from your branch and it is working well, thank you!

As you mention, actually the "better defaults" behaviour actually satisfies the original bug without even needing the sqlc.arg support, but thanks for enabling that too

```go
type GetEventByDateParams struct {
	FromDate time.Time
	ToDate   time.Time
}
```

SnoozeThis:
Resolved.

PR #2366, opened 2023-06-24
Improve default names for BETWEEN ? AND ? to have prefixes from_ and to_

Jille:
issue #2277

### DIFF
--- a/internal/endtoend/testdata/between_args/mysql/go/query.sql.go
+++ b/internal/endtoend/testdata/between_args/mysql/go/query.sql.go
@@ -16,12 +16,12 @@ WHERE   price BETWEEN ? AND ?
 `
 
 type GetBetweenPricesParams struct {
-	Price   int32
-	Price_2 int32
+	FromPrice int32
+	ToPrice   int32
 }
 
 func (q *Queries) GetBetweenPrices(ctx context.Context, arg GetBetweenPricesParams) ([]Product, error) {
-	rows, err := q.db.QueryContext(ctx, getBetweenPrices, arg.Price, arg.Price_2)
+	rows, err := q.db.QueryContext(ctx, getBetweenPrices, arg.FromPrice, arg.ToPrice)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


Jille:
My change to the number was just random code cleanup, but I think that fixed support for BETWEEN sqlc.arg(min) AND sqlc.arg(max). Added a test case to keep it that way.

internal/endtoend/testdata/between_args/mysql/go/query.sql.go
```diff
@@ -16,12 +16,41 @@ WHERE   price BETWEEN ? AND ?
...
+	return items, nil
+}
+
+const getBetweenPricesNamed = `-- name: GetBetweenPricesNamed :many
+SELECT  id, name, price
+FROM    products
+WHERE   price BETWEEN sqlc.arg(min_price) AND sqlc.arg(max_price)
+`
+
+func (q *Queries) GetBetweenPricesNamed(ctx context.Context) ([]Product, error) {
```
kyleconroy, line 52: I think your change has inadvertently broken this code. Notice that this function now has zero arguments. I think you need to keep the code that picked the number for the argument.

kyleconroy (changes requested):

internal/endtoend/testdata/between_args/mysql/go/query.sql.go
Jille: Actually, that never worked. #2277 was filed about that including https://play.sqlc.dev/p/7ecf294cb7aa0b3ce1025657e091cce22970d3b4f5c4eb2baa578e5083cd4219 which shows it doesn't work.

I thought I had fixed it, but maybe I looked wrong. I'll drop the second commit [the one adding this broken test case] from this PR for now and will have a look.

Jille: It needed an extra fix. I made that in https://github.com/kyleconroy/sqlc/pull/2373.

kyleconroy (approved):

PR #2369, opened 2023-06-27
internal/codegen/golang: omit unused structs from output

andrewmbenton:
This is a feature that has been requested from time to time, most-recently in https://github.com/kyleconroy/sqlc/pull/2336.

This implementation introduces a new configuration parameter `omit_unused_structs` which, when set to `true`, prevents inclusion of table and enum structs that aren't used in queries for a given package.

Closes https://github.com/kyleconroy/sqlc/issues/710


internal/codegen/golang/gen.go
```diff
@@ -288,3 +293,42 @@ func usesBatch(queries []Query) bool {
...
+		}
+		if query.hasRetType() {
+			keepTypes[strings.TrimPrefix(query.Ret.Type(), "Null")] = struct{}{}
+			if query.Ret.IsStruct() {
+				for _, field := range query.Ret.Struct.Fields {
+					keepTypes[strings.TrimPrefix(field.Type, "Null")] = struct{}{}
+				}
+			}
+		}
+	}
```
kyleconroy, line 317: The trimming behavior here has a subtle bug. Type names come from user-defined table names. A table named `null_foo` will have a struct named `NullFoo`. Instead of trimming the null prefix, The enum code below should check for any use of the enum struct or the nullable enum struct.

kyleconroy (changes requested):

internal/codegen/golang/gen.go
andrewmbenton: Good point, will update

kyleconroy (approved):

PR #2373, opened 2023-06-27
Add support for BETWEEN sqlc.arg(min) AND sqlc.arg(max)

Jille:
issue #2277


kyleconroy:
@Jille You'll need to rebase on main, but this is good to go once that happens.

Jille:
Rebased :)

PR #2377, opened 2023-06-28
Allow for DDL and unsupported statements as a query

Jille:
By adding 'no-inference' to your query comment, sqlc won't try to understand your query. (The dolphin/postgresql/sqlite parser still need to understand your syntax.) This means that params and output columns don't work, but you can run any query you want.


andrewmbenton:
Sorry for the delay @Jille. I think we shouldn't introduce new query comment syntax, but the idea here is good. I think what we should do is allow `:exec` to be more permissive. In other words, I think (to modify your test example a little) the following should "just work" instead of silently doing nothing like sqlc does currently.

```sql
CREATE TABLE foo (val text not null);

-- name: CreateTempTable :exec
CREATE TEMPORARY TABLE bar (LIKE foo);
```

andrewmbenton:
@Jille the new database-backed analyzer actually handles the test case in my last comment.

Issue #2398, opened 2023-06-30
incorrect ambiguous reference check


Stumble:
### Version

1.18.0

### What happened?

There seems to be a minor bug in checking ambiguous references when joning tables while using alias.

To reproduce, change the `sqlc/internal/endtoend/testdata/order_by_non_existing_column/postgresql` to
```
-- Example queries for sqlc
CREATE TABLE authors (
  id   INT
);

CREATE TABLE books (
  id   INT,
  author_id INT,
  price INT
);

-- name: ListAuthors :many
SELECT id FROM authors
ORDER BY adfadsf;

-- name: ListAuthorsByCheapestBook :many
SELECT
author_id, min(b.price) as min_price
From books b inner join authors a on a.id = b.author_id
GROUP BY b.author_id
ORDER BY min_price;
```

The column of min_price is not ambiguous but sqlc reports:
```
# package 
query.sql:13:1: column reference "adfadsf" not found: if you want to skip this validation, set 'strict_order_by' to false
query.sql:17:1: column reference "min_price" is ambiguous: if you want to skip this validation, set 'strict_order_by' to false

```

### Relevant log output

```shell
query.sql:13:1: column reference "adfadsf" not found: if you want to skip this validation, set 'strict_order_by' to false
query.sql:17:1: column reference "min_price" is ambiguous: if you want to skip this validation, set 'strict_order_by' to false
```

### Database schema

```sql
CREATE TABLE authors (
  id   INT
);

CREATE TABLE books (
  id   INT,
  author_id INT,
  price INT
);
```

### SQL queries

```sql
-- name: ListAuthorsByCheapestBook :many
SELECT
author_id, min(b.price) as min_price
From books b inner join authors a on a.id = b.author_id
GROUP BY b.author_id
ORDER BY min_price;
```

### What operating system are you using?

Linux

### What database engines are you using?

PostgreSQL

### What type of code are you generating?

Go

akutschera:
That's an interesting one. I believe this has been in the code since https://github.com/kyleconroy/sqlc/pull/1348
What happens seems to be this:
For each column name in a GROUP BY or ORDER BY clause, we check, if we have it in a table definition.
When we cannot find it, we check the select statement. When we find it there (e.g. in an alias) we say we found it.
Because we iterate over the number of tables in the statement, we will find it twice. That causes this problem.

The PR above should fix this (and also #1886 which seems to be very similar to this one here).

Stumble:
[...]

Issue #1886, opened 2022-10-07
[PostgresSQL] Failure to use GROUP BY for column output names when using subqueries


voluntas:
### Version

1.15.0

### What happened?

When using a subquery, if I specify an output column name for GROUP BY, sqlc cannot compile it. if I execute SQL directly, it works fine.

I am aware that when using PostgreSQL, column output names can be specified as GROUP BY. However, if I am mistaken, please let me know.

### Database schema

```sql
CREATE TABLE spam (
    id SERIAL PRIMARY KEY
);
```

### SQL queries

```sql
-- name: Spam :many
SELECT
  s.id AS "s_id"
FROM
  -- Using subqueries doesn't work.
  (
    SELECT
      spam.id
	FROM spam
  ) AS "s"
-- Change this to s.id and it will work.
GROUP BY s_id;
```

### Playground URL

https://play.sqlc.dev/p/b3ae0c313bd30c061316dc729e285eba780ba4c7c94e4a44eaac81c619ce9c99

### What operating system are you using?

macOS

### What database engines are you using?

PostgreSQL

### What type of code are you generating?

Go

PR #2399, opened 2023-07-01
fix: count column aliases only once

akutschera:
Alias column names don't exist in the table definition. When we take it from the select target list, we only need to do this once (and not once for every table in a join statement).

Refs: #2398 #1886


andrewmbenton:
Thanks for this and sorry for the delay. There is definitely a bug, but I don't know if restricting the restarget counting is the right fix (although your change does seem to have "fixed" the bug). The issue seems to be that sqlc doesn't understand what list of refs the aliases in `ORDER BY` and `GROUP BY` clauses should be checked against.

I'm going to dig in a little bit and see if I can come up with a recommendation for a different approach.

PR #2406, opened 2023-07-06
docs: Add release notes for v1.19.0

kyleconroy:


docs/reference/config.md
```diff
@@ -72,7 +80,32 @@ sql:
       emit_async_querier: true
       query_parameter_limit: 5
 ```
-  
+
+### database
+
+The `database` mapping supports the following keys:
+
+- `url`:
```
andrewmbenton, line 88: I know this isn't the ideal time to offer naming suggestions, but we can't really change this after release so it's worth a brief discussion. `uri` is better than `url` in this context I think, if we want to indicate that we are only supporting the URI-style connection string format for various drivers (and not other formats e.g. key/value pairs).

I took a quick read of the [postgres](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING) and [mysql](https://dev.mysql.com/doc/refman/8.0/en/connecting-using-uri-or-key-value-pairs.html) and [sqlite](https://www.sqlite.org/c3ref/open.html) docs on this and they all mention "URI."

docs/howto/ci-cd.md
```diff
@@ -0,0 +1,92 @@
...
+    steps:
+    - uses: actions/checkout@v3
+    - uses: sqlc-dev/setup-sqlc@v3
+      with:
+        sqlc-version: '1.19.0'
+    - run: sqlc diff
+```
+
+We also encourage running [`sqlc vet`](../reference/cli.html#vet). To get the most value out of `vet`,
+you'll want to set up a running database. See the [vet] documentation for a
```
andrewmbenton, line 58: [vet] is missing a link url here, but would probably be redundant given the link in the previous sentence so probably just remove square brackets.

PR #2426, opened 2023-07-10
Fix: Update sqlc/db-prepare command name in docs

ngutierrez3:
`sqlc/db-prepare` was misspelled, updating to match the rule name properly


kyleconroy (approved):

andrewmbenton:
Good catch, thanks!

PR #2427, opened 2023-07-10
[DRAFT] Support for PostgreSQL "RECORD" return types

jwc-clinnection:
In this draft change, we've added support for PostgreSQL functions that return the "RECORD" data type.

### DIFF
--- a/internal/compiler/output_columns.go
+++ b/internal/compiler/output_columns.go
@@ -515,13 +515,30 @@ func (c *Compiler) sourceTables(qc *QueryCatalog, node ast.Node) ([]*Table, erro
 			}
 
 			fn, err := qc.GetFunc(funcCall.Func)
-			if err != nil {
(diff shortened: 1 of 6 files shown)


andrewmbenton:
Hey John, thanks for this! There was some work done (but not merged) to handle functions that return tables which may be similar to what you've done here: https://github.com/kyleconroy/sqlc/pull/1973. If you have a chance to take a look, let me know if you think there's redundancy.

In any case, can you share an example schema and query you're using to develop this change? Or better yet, add a test case to `internal/endtoend/testdata` and push it to your branch?

andrewmbenton:
Since it hasn't been updated in awhile I'm going to close this PR. We have some ideas for supporting composite types when using `pgx/v5`which may overlap with this work: https://github.com/sqlc-dev/sqlc/issues/2760

PR #2443, opened 2023-07-12
feat: improved inference for parameter arguments passed to nullif.

### LINKED ISSUE #1821: sqlite: nullif(?,"") (and other builtin functions) does not generate the correct type
### Version

1.15.0

### What happened?

Tried both on v1.15.0 and HEAD (as of August 26 2022).

sqlite schema:

```sql
-- name: UpdateChild :one
CREATE TABLE IF NOT EXISTS Child(
    ID INTEGER NOT NULL PRIMARY KEY,
    PublicID TEXT NOT NULL,
    Name TEXT NOT NULL,
    DOB TEXT NOT NULL, 
    Photo BLOB
);
```

Given the following query:

```sql
-- name: UpdateChild :one
UPDATE Child SET
    name  = COALESCE(nullif(?,""), name),
    dob   = COALESCE(nullif(?,""), dob),
    photo = COALESCE(?,photo)
    WHERE
        publicid = ?
    RETURNING *;
```
sqlc generates:
```go
type UpdateChildParams struct{
  NULLIF   interface{}
  NULLIF_2 interface{}
  Photo    []byte
  Publicid string
}
```

It should instead generate:
```go
type UpdateChildParams struct{
  Name     interface{}
  Dob      interface{}
  Photo    []byte
  Publicid string
}
```

Which it does generate if I remove `nullif`:
```sql
-- name: UpdateChild :one
UPDATE Child SET
    name  = COALESCE(?, name),
    dob   = COALESCE(?, dob),
    photo = COALESCE(?,photo)
    WHERE
        publicid = ?
    RETURNING *;
```

I need the `nullif` because `COALESCE` only works with `NULL` values.

### Relevant log output

_No response_

### Database schema

_No response_

### SQL queries

_No response_

### Configuration

_No response_

### Playground URL

https://play.sqlc.dev/p/7208959872393525fe74f3f7250180226cda8da6e1d3002578a017b923ee3a11

Linux

_No response_

Go

orisano:
close #1821


andrewmbenton:
I think rather than special-casing the func call parameter naming, the preferred solution would be to use `sqlc.arg()` for this to force sqlc to use the specific name you want: https://docs.sqlc.dev/en/stable/howto/named_parameters.html

I will close this and the original issue, but feel free to comment or reopen if you disagree.

orisano:
I thought it was a good idea because it solves both the parameter name problem and the type problem.

andrewmbenton:
> I thought it was a good idea because it solves both the parameter name problem and the type problem.

Yeah I acted too soon without understanding the full scope of the problem. Sorry about that. I think we should reopen and probably implement this with the expectation that there may be more "passthrough" functions added to the list. Or is `nullif()` really the only one where this is the desired behavior?

PR #2448, opened 2023-07-13
Mention UUID type for pgx/v5

peterwillis:


andrewmbenton:
Nevermind, you can disregard my note above. I hadn't seen that your edit matches the style of similar sentences in other parts of the docs.

PR #2456, opened 2023-07-13
docs: Use downloads.sqlc.dev for install links

kyleconroy:


mfridman:
Will the GitHub release continue to work, or are we advised to switch to downloads.sqlc.dev ?

kyleconroy:
GitHub releases will continue to work, but I'd advise you to use https://downloads.sqlc.dev as it won't get rate limited and it's powered by Google's Cloud CDN, so I think might be faster.

PR #2462, opened 2023-07-14
docs: Add a description to the document for cases when a query result has no rows

n3xem:
I added a description to the document for cases when a query result has no rows because I found it unclear what the query returns when reading the document.


kyleconroy (approved):

kyleconroy:
Thanks!

Issue #2466, opened 2023-07-15
Include getters and setters with models so we may  work with interfaces instead of specific structs


blwsh:
### What do you want to change?

Selecting all fields and a subset of fields for the same table yields different structs.  Currently, we're required to write a mapper function which transforms the row(s) in to an abstraction our applications can understand - however, if sqlc was to provide getters and setters, we'd be able to define interfaces which the results conform to and leave a lot of the heavy lifting to go.

Please see the example below:

```sql
CREATE TABLE `products`
(
    `name`  varchar(255) NOT NULL,
    `price` int(11)      NOT NULL
)
```

```go
// generated models/queries

// select * from products
type Product struct {
  Name string
  Price int
}
func (p Product) GetName() string { ... }
func (p Product) GetPrice() int { ... }

// select `name` from products
type ProductNameRow struct {
  Name string
}
func (p ProductNameRow) GetName() string { ... }

// app

type ProductWithName interface {
  GetName() string
}

func DoesProductHaveCoolName[T ProductWithName](product T) bool { ... }

DoesProductHaveCoolName(&ProductNameRow{}) // good
DoesProductHaveCoolName(&Product{}) // also good
```

For this particular problem extending the template for structs to include getters and setters would be sufficient, however - it may be a better idea to allow the extension of the template via the config file instead?

### What database engines need to be changed?

PostgreSQL, MySQL

### What programming language backends need to be changed?

Go

kyleconroy:
Closed via https://github.com/sqlc-dev/sqlc/pull/2471#issuecomment-1656307541

PR #2471, opened 2023-07-17
feat(generation): Add new emit_getters option (Go Only)

blwsh:
Includes getters with Go Lang structs so they can be consumed as interfaces. Enabled with the flag `emit_getters` (`false` by default). Available in config versions v1 and v2.

Example:

```go
// sqlc generated

// select * from products
type Product struct {
  Name string
  Price int
}
func (p Product) GetName() string { ... }
func (p Product) GetPrice() int { ... }

// select `name` from products
type ProductNameRow struct {
  Name string
}
func (p ProductNameRow) GetName() string { ... }

// app

type ProductWithName interface {
  GetName() string
}

func DoesProductHaveCoolName[T ProductWithName](product T) bool { ... }

DoesProductHaveCoolName(&ProductNameRow{}) // good
DoesProductHaveCoolName(&Product{}) // also good
```

Resolves https://github.com/kyleconroy/sqlc/issues/2466


kyleconroy:
@blwsh Appreciate opening this up. This isn't something we're looking to support at the time. We're hoping to solve this specific issue by allowing you to bring your own models to sqlc, where you can specify getters or any other methods you'd like to have on your structs.

To solve this issue, another sqlc user @Jille has written https://github.com/Jille/go_get_set_generator which will generate Getter methods for you from Go source.

Issue #2454, opened 2023-07-13
Parse and handle additional query metadata via annotations in comments


andrewmbenton:
### What do you want to change?

Kyle and I have been discussing how to add support for more query-specific configuration. The particular need driving this is to have a way to opt individual queries out of [sqlc vet rules](https://docs.sqlc.dev/en/stable/howto/vet.html) (which are set globally for a sqlc package). But it would also be nice to provide a common configuration syntax and behavior for passing query-specific metadata to plugins (e.g. as requested here: https://github.com/kyleconroy/sqlc/issues/2270).

The proposal is:

In any SQL comment, a line starting with `@symbol` will have all text following the symbol parsed as yaml. If no text follows the symbol then we infer a boolean true value.

We would then pass that structured data to plugins, as well as make use of it internally.

Here's a hypothetical queries.sql:
```sql
-- name: GetAuthor :one
-- @sqlc-vet-disable # implies skip all vet rules
SELECT * FROM authors
WHERE id = $1 LIMIT 1;

-- name: ListAuthors :many
-- @sqlc-vet-disable [foo, bar] # disable the vet rules named foo and bar
-- @plugin-attr {"foo": "bar"} # a k/v map used by some plugin
SELECT * FROM authors
ORDER BY name;
```

If you have feedback on this please leave a comment. I'd especially like to hear from anyone who's aware of tooling that does something similar with comments in sql files, so that we can avoid inadvertent conflicts.

PR #2474, opened 2023-07-18
feat(vet): Introduce a query annotation to opt out of sqlc vet rules

andrewmbenton:
Introducing the `@sqlc-vet-disable` query annotation, to opt individual queries out of sqlc vet rules, based on https://github.com/kyleconroy/sqlc/issues/2454.

I'm not sure that printing a message to stdout when skipping a query is ideal, but I didn't want to skip silently or return an error and I couldn't use stderr obviously. If skipping silently is better for now I'm happy to remove that line.

Related to the above, I couldn't figure an easy way to test this new functionality without passing in an output stream separate from stderr where a test would expect the skip message to appear. I can do that, but I wasn't sure about the approach and I'm not a testing expert so maybe there's a better thing we could do. Anyway please advise if there's something I should add regarding testing.

### DIFF
--- a/internal/cmd/vet.go
+++ b/internal/cmd/vet.go
@@ -31,6 +31,7 @@ import (
 var ErrFailedChecks = errors.New("failed checks")
 
 const RuleDbPrepare = "sqlc/db-prepare"
+const QueryFlagSqlcVetDisable = "@sqlc-vet-disable"
 
 func NewCmdVet() *cobra.Command {
 	return &cobra.Command{
@@ -249,7 +250,6 @@ func (c *checker) checkSQL(ctx context.Context, s config.SQL) error {
 		return ErrFailedChecks
 	}
 
-	// TODO: Add MySQL support
 	var prep preparer
 	if s.Database != nil {
 		if c.NoDatabase {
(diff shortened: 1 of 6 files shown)


internal/cmd/vet.go
```diff
@@ -299,6 +299,10 @@ func (c *checker) checkSQL(ctx context.Context, s config.SQL) error {
 	req := codeGenRequest(result, combo)
 	cfg := vetConfig(req)
 	for i, query := range req.Queries {
+		if result.Queries[i].Flags[QueryFlagSqlcVetDisable] {
+			fmt.Printf("Skipping vet rules for query %s", query.Name)
```
kyleconroy, line 303: I think printing to stderr is fine and more correct than printing to stdout. Can we change the format to match error output? Also, if you want to suppress the output by default, you can check to see if `debug.Active` is true and only print it then.

```
fmt.Fprintf(c.Stderr, "%s: %s: %s: skipping\n", query.Filename, q.Name, name)
```

andrewmbenton: The problem with printing to stderr I think is that's where "normal" vet rule errors are reported, so skipping a rule would trigger a vet fail every time right?

The `debug.Active` check would suppress that except in cases where someone explicitly opted-in, but I'm still not sure it's the right thing to do.

PR #2482, opened 2023-07-20
fix(engine/sqlite): support for repeated table_option

### LINKED ISSUE #2481: SQLite: support STRICT, WITHOUT ROWID tables

In line with https://github.com/kyleconroy/sqlc/pull/2175, SQLite tables defined using both STRICT and WITHOUT ROWID aren't currently supported.

This works:

```sql
CREATE TABLE authors (
  id   BIGSERIAL PRIMARY KEY,
  name text      NOT NULL,
  bio  text
) WITHOUT ROWID;
```

This works:

```sql
CREATE TABLE authors (
  id   BIGSERIAL PRIMARY KEY,
  name text      NOT NULL,
  bio  text
) STRICT;
```

This doesn't work:

```sql
CREATE TABLE authors (
  id   BIGSERIAL PRIMARY KEY,
  name text      NOT NULL,
  bio  text
) STRICT, WITHOUT ROWID;
```

SQLite

I'm using Go, not sure if that matters for this feature request.

orisano:
fix #2481

ref: https://www.sqlite.org/syntax/create-table-stmt.html
ref: https://www.sqlite.org/syntax/table-options.html


kyleconroy:
Thanks for adding a new end-to-end test to verify that this works.

PR #2511, opened 2023-07-24
fix(vet): Split vet messages out of codegen.proto

kyleconroy:
We'll move it back into the codegen package once we're ready for vet WASM plugins.


andrewmbenton (approved):
This looks nice.

I'm just realizing though that I don't think the `cel.Types()` call is doing anything, or at least doesn't seem to be required?

Nothing else in the vet codepath references `PostgreSQLExplain` and `MySQLExplain` although they are types of fields on `vet.PostgreSQL` and `vet.MySQL`.

I guess if that call to `cel.Types()` isn't required we can remove it in a separate PR.

kyleconroy:
Okay I think this is ready to go.

PR #2513, opened 2023-07-24
Clarify option package support

johanbrandhorst:
The emit_pointers_for_null_types option supports both pgx/v4 and pgx/v5 according to the tests:

https://github.com/kyleconroy/sqlc/tree/main/internal/endtoend/testdata/emit_pointers_for_null_types

Ensure that this is clear in the documentation.


kyleconroy (approved):
Thanks! The original feature predated pgx/v5 I believe.

PR #2521, opened 2023-07-25
feat(debug): add debug flag and docs for dumping vet rule variables

andrewmbenton:


kyleconroy (approved):

internal/cmd/vet.go
```diff
@@ -430,6 +430,10 @@ func (c *checker) checkSQL(ctx context.Context, s config.SQL) error {
 				evalMap["mysql"] = engineOutput.MySQL
 			}
 
+			if debug.Debug.DumpVetEnv {
+				fmt.Printf("vars for rule '%s' evaluating against query '%s':\n%v\n", name, query.Name, evalMap)
```
kyleconroy, line 434: Can we dump this as JSON?

PR #2522, opened 2023-07-25
fix missing unsigned param

### LINKED ISSUE #2432: Unsigned types not present in query parameters
### Version

Other

### What happened?

**sqlc version**: `1.19.0`

When generating code for an SQL query against a table with fields marked as `UNSIGNED`, the query arguments should also be unsigned. The currently generated code for the query relies on signed types, which is not the desired behavior.

In the case presented below, the following code is generated:

```go
type UpsertMappingParams struct {
	LeftID  int64
	RightID int64
}
```

The expected and correct result should be:

```go
type UpsertMappingParams struct {
	LeftID  uint64
	RightID uint64
}
```

### Relevant log output

_No response_

### Database schema

```sql
CREATE TABLE IF NOT EXISTS `mappings` (
  `id` BIGINT(20) UNSIGNED NOT NULL,
  `left_id` BIGINT(20) UNSIGNED NOT NULL,
  `right_id` BIGINT(20) UNSIGNED NOT NULL,
  PRIMARY KEY (`left_id`, `right_id`),
  UNIQUE INDEX `id_UNIQUE` (`id` ASC) VISIBLE,
  INDEX `fk_left_id_idx` (`left_id` ASC) VISIBLE,
  INDEX `fk_right_id_idx` (`right_id` ASC) VISIBLE
);
```

### SQL queries

```sql
-- name: UpsertMapping :exec
INSERT INTO
    mappings (left_id, right_id)
VALUES
    (?, ?)
ON DUPLICATE KEY UPDATE
    id = id;
```

### Configuration

```yaml
{
  "version": "1",
  "packages": [
    {
      "path": "db",
      "engine": "mysql",
      "schema": "query.sql",
      "queries": "query.sql"
    }
  ]
}
```

### Playground URL

https://play.sqlc.dev/p/6316120fffa5029678410f26691f24670d912069e1345473c95c1f0d6825db6a

Linux

MySQL

Go

hiroebe:
Fixes #2432


internal/endtoend/testdata/unsigned_params/mysql/go/query.sql.go
```diff
@@ -0,0 +1,19 @@
...
+
+import (
+	"context"
+)
+
+const createFoo = `-- name: CreateFoo :exec
+INSERT INTO foo (id) VALUES (?)
+`
+
+func (q *Queries) CreateFoo(ctx context.Context, id uint32) error {
```
hiroebe, line 16: Note: The `id` parameter was previously generated as `int32` while it's unsigned in MySQL schema.

PR #2543, opened 2023-07-29
feat(vet): add default query parameters for vet explain with MySQL

kyleconroy:


internal/cmd/vet.go
```diff
@@ -170,16 +170,145 @@ func (p *pgxConn) Prepare(ctx context.Context, name, query string) error {
...
+	case "decimal": // "numeric", "dec", "fixed"
+		// No perfect choice here to avoid "Impossible WHERE" but I think
+		// 0.1 is decent. It works for all cases where `scale` > 0 which
+		// should be the majority. For more information refer to
+		// https://dev.mysql.com/doc/refman/8.1/en/fixed-point-types.html.
+		return 0.1
+	case "float", "double":
+		return 0.1
+	case "date":
+		t := time.Time{}
```
andrewmbenton, line 282: If we just want the zero time values I think we can dispense with instantiating the Go structs and just use hardcoded strings (see bottom of this page: https://dev.mysql.com/doc/refman/8.1/en/date-and-time-types.html). I wasn't sure that the zero values would be the best choice though.

internal/cmd/vet.go
```diff
@@ -170,16 +170,145 @@ func (p *pgxConn) Prepare(ctx context.Context, name, query string) error {
...
+	if col == nil {
+		return nil
+	}
+	if col.Type == nil {
+		return nil
+	}
+	switch col.Type.Name {
+	case "any":
+		return nil
+	case "bool":
```
andrewmbenton, line 269: There is no `"bool"` type for MySQL; it's just an alias for `tinyint` with `0`/`1` values (which my `Println()` testing bears out). But maybe I'm not thinking correctly about the possible ways the input `col` gets constructed, and we might see something come through this code path with `col.Type.Name == "bool"`?

internal/cmd/vet.go
```diff
@@ -170,16 +170,145 @@ func (p *pgxConn) Prepare(ctx context.Context, name, query string) error {
...
+		// 0.1 is decent. It works for all cases where `scale` > 0 which
+		// should be the majority. For more information refer to
+		// https://dev.mysql.com/doc/refman/8.1/en/fixed-point-types.html.
+		return 0.1
+	case "float", "double":
+		return 0.1
+	case "date":
+		return "0000-00-00"
+	case "datetime", "timestamp":
+		return time.Time{}
```
andrewmbenton, line 284: Why not use `"0000-00-00 00:00:00"` here?

internal/cmd/vet.go
kyleconroy: It's an alias, but a cool columns will have that type at this stage. When you pass false to the MySQL driver, it converts it to a 0. I found this case somewhere in the tests.

internal/cmd/vet.go
kyleconroy:
```suggestion
		return "0000-00-00 00:00:00"
```

PR #2545, opened 2023-07-30
feat(mysql): :copyfrom support via LOAD DATA INFILE

### LINKED ISSUE #2179: [proposal] Support LOAD DATA LOCAL INFILE for MySQL

kyleconroy:
This enables the :copyfrom query annotation for people using go-sql-driver/mysql that transforms it into a LOAD DATA LOCAL INFILE.

We don't have a way to get the timezone from the connection, so I've simply blocked people from using time.Times in their copyfrom.

Fixes #2179


internal/codegen/golang/gen.go
```diff
@@ -145,8 +145,15 @@ func generate(req *plugin.CodeGenRequest, enums []Enum, structs []Struct, querie
...
-		return nil, errors.New(":copyfrom is only supported by pgx")
+	if tctx.UsesCopyFrom && !tctx.SQLDriver.IsPGX() && golang.SqlDriver != SQLDriverGoSQLDriverMySQL {
+		return nil, errors.New(":copyfrom is only supported by pgx and github.com/go-sql-driver/mysql")
+	}
+
+	if tctx.UsesCopyFrom && golang.SqlDriver == SQLDriverGoSQLDriverMySQL {
+		if err := checkNoTimesForMySQLCopyFrom(queries); err != nil {
+			return nil, err
+		}
+		tctx.SQLDriver = SQLDriverGoSQLDriverMySQL
```
kyleconroy, line 156: This is a gross hack, but I don't want to change how parseDriver works right now.

PR #2549, opened 2023-07-31
feat(cmd/sqlc): Bump version to 1.20.0

kyleconroy:


docs/reference/changelog.md
```diff
@@ -1,6 +1,242 @@
...
+INSERT INTO foo (a, b, c, d) VALUES (?, ?, ?, ?);
+```
+
+```go
+func (q *Queries) InsertValues(ctx context.Context, arg []InsertValuesParams) (int64, error) {
+	...
+}
+```
+
+`LOAD DATA` support must be enabled in the MySQL client and server.
```
andrewmbenton, line 123: Isn't the MySQL client just `github.com/go-sql-driver/mysql`?

PR #2561, opened 2023-08-02
docs/howto: automatic migration planning with atlas

rotemtam:


iampeter:
Great timing, I was just looking for this.

PR #2572, opened 2023-08-03
fix(postgres): Fixed a problem with array dimensions disappearing when using "ALTER TABLE ADD COLUMN"

### LINKED ISSUE #2571: array column added from alter table are generated as non array
### Version

1.20.0

### What happened?

On alter tables, array columns are generated as non array. This doesn't happen in 1.19.

image

### Relevant log output

_No response_

### Database schema

```
-- Example queries for sqlc
CREATE TABLE authors (
  id   BIGSERIAL PRIMARY KEY,
  name text      NOT NULL,
  array1 BIGINT[],
  array2 BIGINT array
);

ALTER TABLE authors ADD COLUMN array3 BIGINT[];
ALTER TABLE authors ADD COLUMN array4 BIGINT array;
```

### SQL queries

```
-- name: GetAuthor :one
SELECT * FROM (select id from authors) t;
```

### Configuration

```
{
  "version": "1",
  "packages": [
    {
      "path": "db",
      "engine": "postgresql",
      "schema": "query.sql",
      "queries": "query.sql"
    }
  ]
}
```

### Playground URL

 https://play.sqlc.dev/p/09d13d4e8c68bda99da95e6e0af5adf85c60396277614faae148fd8b56a9dec4

_No response_

_No response_

_No response_

orisano:
close #2571

### DIFF
--- a/internal/sql/catalog/table.go
+++ b/internal/sql/catalog/table.go
@@ -57,6 +57,7 @@ func (table *Table) addColumn(cmd *ast.AlterTableCmd) error {
 		IsNotNull:  cmd.Def.IsNotNull,
 		IsUnsigned: cmd.Def.IsUnsigned,
 		IsArray:    cmd.Def.IsArray,
+		ArrayDims:  cmd.Def.ArrayDims,
 		Length:     cmd.Def.Length,
 	})
 	return nil


orisano:
@kyleconroy If it is array, primitive types are used. Is this behavior correct?

kyleconroy:
> Is this behavior correct?

Yep! Both pgx and lib/pq can scan into slices. With lib/pq you need to do a bit more work with `pq.Array`, but sqlc does that for you.

Issue #2589, opened 2023-08-08
emit_pointers_for_null_types does not work on default 'database/sql' go package


loicbefve:
### Version

1.20.0

### What happened?

While generating the go queries using the `emit_pointers_for_null_types` flag in a configuration file v2 nothing append and database/sql nullable types are used instead.
I think I've found where in the code that might happen and will propose a fix alongside this issue.
Thanks for reading me.

### Configuration

```yaml
version: "2"
sql:
  - engine: "postgresql"
    queries: "query.sql"
    schema: "../database/schema/1.2.9.sql"
    gen:
      go:
        emit_interface: true
        emit_enum_valid_method: true
        emit_all_enum_values: true
        emit_pointers_for_null_types: true
        package: "database_sqlc"
        out: "."
```

### What operating system are you using?

macOS

### What database engines are you using?

PostgreSQL

### What type of code are you generating?

Go

loicbefve:
#2590 To fix this issue

kyleconroy:
As discussed in the linked pull request, this is surprising but working as intended. By default, code that uses `database/sql` and `lib/pq` can't support scanning into pointer fields. In the future we can add a warning when this parameter is set to true and `lib/pq` is being used.

PR #2590, opened 2023-08-08
Allow flag emitPointersForNull to all drivers

loicbefve:
Here is a proposition to fix the issue #2589 simply allow to use pointers for all drivers.
If this is not possible, I'm curious to understand why :)


kyleconroy (commented):
Ahh, if only it were this simple!

Each database driver handles types different. The https://github.com/go-sql-driver/mysql and https://github.com/lib/pq drivers do not support scanning null values into pointers.

PR #2617, opened 2023-08-14
Add a config flag to add a context key with the current query name

Jille:
Multiple people have asked for more involved features like stats, but with this simple context injection people can implement their own DBTX and do whatever they want from there.

My use case is also measuring query durations per query


kyleconroy:
It's an interesting approach. Before merging this, I'd like to spend a bit more time thinking through a solution that uses the interceptor pattern from gRPC. I'm not sure it will map cleanly to sqlc-generated code, but it would solve many of the customization issues that we face today without requiring new configuration parameters.

kyleconroy:
**🤖 Automated message**

---

PR #2618, opened 2023-08-14
Enable custom handling of prepared queries

Jille:
Without prepared queries, one can use arbitrary wrappers as a DBTX. For example to measure query timing.

With prepared queries however, only DBTX.PrepareContext() is called, which must return a *sql.Stmt. That is a concrete type and there's no easy way to create one yourself with any overrides.

Even more difficult is when in a transaction, when we call (*sql.Tx).StmtContext() which also must return a *sql.Stmt. *sql.Tx is also a concrete type here that can't be overwritten.

Because a Go interface { StmtContext(ctx, *sql.Stmt) AnotherInterface } wouldn't match *sql.Tx, we can't easily just intercept even that.

So the best way I can think of is allowing a custom interceptor which people can use to do whatever they want in their wrapping.

### DIFF
--- a/examples/ondeck/mysql/db.go
+++ b/examples/ondeck/mysql/db.go
@@ -21,6 +21,12 @@ func New(db DBTX) *Queries {
 	return &Queries{db: db}
 }
 
+type PreparedStmtInterceptor interface {
+	ExecContext(context.Context, *sql.Stmt, *sql.Tx, ...interface{}) (sql.Result, error)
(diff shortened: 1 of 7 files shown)


Jille:
One alternative is a config flag that simply omits (*Queries).{exec,query,queryRow}. Those have pretty much the signature of the callback we should expose. (Only difference being the query string being passed in next to the prepared statement.)

The advantage is that it is less complexity on our side. The disadvantage is that users would have to add files to the generated package to implement these private methods.

PR #2619, opened 2023-08-15
docs: Reword phrase about pgx and multidimensional arrays

rollulus:
English is not my native language, so I could be wrong, but my interpretation of the text was that pgx can only do multidimensional arrays, and nothing else. I think that's not right. Hope my commit corrects this.

Thanks for sqlc and best of luck with your company around it.


andrewmbenton:
Good catch. I agree that your placement of "only" is much clearer.

But we actually added support for multidimensional arrays with other driver packages (in https://github.com/sqlc-dev/sqlc/pull/2338) and forgot to update this documentation, so the whole sentence can go.

rollulus:
Even better, thanks.

Issue #2607, opened 2023-08-12
Deduplicate generated structs


loicbefve:
### What do you want to change?

Let's consider those two SQL queries :

**Query 1:**
```sql
-- name: get_members :many
SELECT * FROM members; 
```

**Query 2:**
```sql
-- name: get_connected_members :many
SELECT * FROM members WHERE status = 'connected';
```

In both cases I get a `MemberRow` but sqlc will generate two different models `GetMembersRow` and `GetConnectedMembersRow` that will be identical.

It would be great that istdetects that they are the same and then generate a unique `MemberRow`.

### What database engines need to be changed?

PostgreSQL, MySQL, SQLite

### What programming language backends need to be changed?

Go, Python, Kotlin

loicbefve:
It could only works on query mapping exactly to a table to avoid naming problems and the name of the model could be the name of the table. In our example `Member`

andrewmbenton:
I think I understand what you're asking, but the specific example you've cited actually does the thing you want (for `SELECT * ...`). Here's a playground URL: https://play.sqlc.dev/p/dc32f36841126622fe128376a72ec1b084582d43762a06ddb0cd92f0d121716f

I am assuming there's an example where two different row structs are generated, so can you share that?

hezise:
I'd like to join this.
my example is more complexed, but the idea is the same.
the requested fields for the two different queries are the same 
but the row struct is different

kyleconroy:
Duplicate of https://github.com/sqlc-dev/sqlc/issues/781

PR #2624, opened 2023-08-15
Reuse queries struct

### LINKED ISSUE #755: Support for custom models

MartyHub:
Hello,

reading #2607, I can see that sqlc is able to reuse a table struct when generating queries output.
To fix #755, maybe the same process can be extended to allow a query output struct to be reused?

Regards

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,6 +10,8 @@ jobs:
     runs-on: windows-latest
     steps:
     - uses: actions/checkout@v4
+    # wasmtime can't be built with 12.2
+    - run: choco install mingw --version 11.2.0.07112021 --allow-downgrade
     - uses: actions/setup-go@v4
       with:
         go-version: '1.21.1'


andrewmbenton:
This is a nice solution I think, but we should have @kyleconroy weigh in as well.

It's a breaking change, so I think if we do merge this we'll want it behind a configuration parameter like `reuse_query_structs` or `reuse_structs`. But that brings up an interesting question I guess. Should we try to have sqlc reuse the query parameter structs as well? A quick search doesn't turn up anyone asking for this in GH issues, but I think it would make sense.

A minor suggestion would be to name the slice of query structs `queryStructs` instead of `structsQueries`.

And then one more question, after looking over the testdata output, are you certain that the choice of struct name is deterministic? That is, can we be sure that we're always iterating over the queries in the same order so that the reused struct will always have the same name? I think the answer is yes but just want to be sure.

MartyHub:
I agree a configuration parameter would be useful to keep compatibility!

Good point regarding deterministic structs iteration order, I'll try to have a look.

Since my PR, I also discovered that it does not play nicely with [sqlc.embed](https://docs.sqlc.dev/en/stable/reference/changelog.html#sqlc-embed), this needs more test cases.

fnsne:
Hello, @MartyHub
The PR error looks like not add `"reuse_structs": false` in the internal/endtoend/testdata/process_plugin_sqlc_gen_json/gen/codegen.json.
But I don't know how to help to fixed that.

截圖 2023-09-22 上午9 52 07

PR #2627, opened 2023-08-17
Add stand_alone_name configuration for each queries

michaelzx:
# Description
This PR adds the string configuration sql.gen.go.stand_alone_name in sqlc.yaml.
# Features
1. The program will not generate the db.go file
2. Merge the original code in db.go into each queries.go file, combine into one file
3. In the new queries.go file, stand_alone_name will be used to differentiate DBTX, New, Queries, and sql constants
# Examples
I have put examples in the examples/stand_alone_name directory, including schema.sql, queries.sql, sqlc.yaml, and the generated go files
Development environment:
Mac, go 1.19.3, PostgreSQL 14
# Note
I have not done complete testing, this is just to provide a reference idea


kyleconroy:
@michaelzx Appreciate you putting together a complete working example for reference. As of right now, we won't be accepting a change like this. Single configuration options that change that dramatically change the structure of the code are difficult to maintain, as they interact with the other configuration options in difficult to understand ways.

I hope that in a few versions codegen plugins will be a bit easier to write and you can implement your idea via that mechanism.

PR #2639, opened 2023-08-20
All not needed an alias when there is a subquery with the same column name

ovadbar:


andrei-dascalu:
@andrewmbenton the workaround doesn't play well with multiple joins on the same table, particularly when one of the joins may be nullable. It also doesn't work when you need to select all the fields - works for a particular field.

kyleconroy:
This should be fixed in [v1.23.0](https://docs.sqlc.dev/en/latest/reference/changelog.html#v1-23-0) by enabling the [database-backed query analyzer](https://docs.sqlc.dev/en/latest/howto/generate.html#enhanced-analysis-with-managed-databases). If not, please open a new issue.

PR #2661, opened 2023-08-26
Keep environment variables on calling plugin

walterwanderley:
Trying to convert https://github.com/walterwanderley/sqlc-grpc into a plugin

### DIFF
--- a/internal/ext/process/gen.go
+++ b/internal/ext/process/gen.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"os"
 	"os/exec"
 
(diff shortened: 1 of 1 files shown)


kyleconroy:
We intentionally limit the environment variables sent to process-based plugins to ensure that those plugins can't easily steal credentials stored in the environment. I'm happy to add an allowlist of variables though and include just those. Which environment variables does your plugin need access to?

walterwanderley:
Ok, Just need the PATH env.

kyleconroy:
Merged the other PR, will be available in the next release :)

PR #2668, opened 2023-08-28
fix(codegen): use correct Go types for bit strings and cid/oid/tid/xid with pgx/v4

andrewmbenton:
resolves https://github.com/sqlc-dev/sqlc/issues/2667


kyleconroy (approved):
These tests would have caught the issue, as the code wouldn't compile.

PR #2669, opened 2023-08-29
feat(codegen): Allow plugins to access environment variables

kyleconroy:
Allow plugins, both process-based and WASM-based, to read environment variables from the host, configured via a per-variable allowlist. Plugins can not declare the environment variables they have access to outside of the sqlc.yaml configuration file.

```
version: '2'
plugins:
- name: py
  env:
  - PATH
  wasm:
    url: https://github.com/sqlc-dev/sqlc-gen-python/releases/download/v0.16.0-alpha/sqlc-gen-python.wasm
    sha256: 428476c7408fd4c032da4ec74e8a7344f4fa75e0f98a5a3302f238283b9b95f2
```

Closes https://github.com/sqlc-dev/sqlc/pull/2661


andrewmbenton:
So the idea here is that the user can decide which variables go in the allow list, via sqlc.conf?

kyleconroy:
Exactly. This is very similar to how wasmtime allows environment variables to be accessed.

```yaml
version: '2'
plugins:
- name: py
  env:
  - PATH
  wasm:
    url: https://github.com/sqlc-dev/sqlc-gen-python/releases/download/v0.16.0-alpha/sqlc-gen-python.wasm
    sha256: 428476c7408fd4c032da4ec74e8a7344f4fa75e0f98a5a3302f238283b9b95f2
sql:
- schema: "schema.sql"
  queries: "query.sql"
  engine: postgresql
  codegen:
  - out: src/authors
    plugin: py
    options:
      package: authors
      emit_sync_querier: true
      emit_async_querier: true
      query_parameter_limit: 5
```

docs/reference/config.md
```diff
@@ -316,6 +316,8 @@ Each mapping in the `plugins` collection has the following keys:
 
 - `name`:
   - The name of this plugin. Required
+- `env`
+  - A list of environment variables to pass to the plugin. By default, no environment variables are passed.
```
andrewmbenton, line 320: There is the `SQLC_VERSION` var which is added (not from the actual shell environment though). That should be documented somewhere, but probably not here though.

docs/guides/plugins.md
```diff
@@ -59,39 +47,51 @@ the new files. The `plugin` key must reference a plugin defined in the
...
+      filename: codegen.json
 ```
 
 For a complete working example see the following files:
 - [sqlc-gen-json](https://github.com/sqlc-dev/sqlc/tree/main/cmd/sqlc-gen-json)
   - A process-based plugin that serializes the CodeGenRequest to JSON
 - [process_plugin_sqlc_gen_json](https://github.com/sqlc-dev/sqlc/tree/main/internal/endtoend/testdata/process_plugin_sqlc_gen_json)
   - An example project showing how to use a process-based plugin
+
+## Environment variables
```
andrewmbenton, line 74: Should mention the `SQLC_VERSION` var in this section I think

kyleconroy: Done

Issue #2690, opened 2023-09-04
MySQL :copyfrom is not working


p1ass:
### Version

1.20.0

### What happened?

#2545 introduced a :copyfrom feature for MySQL database thanks to @kyleconroy .
However, it is not woking well.

I think the cause of problem is `parseDriver` function never returns `SQLDriverGoSQLDriverMySQL`.
https://github.com/sqlc-dev/sqlc/blob/faa1c9d156272eea4d33a31de2ab94951ba6af58/internal/codegen/golang/driver.go#L11C1-L35

### Relevant log output

```shell
sqlc generate failed.
# package db
error generating code: :copyfrom is only supported by pgx and github.com/go-sql-driver/mysql
```
```


### Database schema

```sql
CREATE TABLE authors (
  id   bigint PRIMARY KEY,
  name text      NOT NULL,
  bio  text
);
```

### SQL queries

```sql
-- name: BulkInsertAuthors :copyfrom
INSERT INTO authors (id, name, bio) VALUES (?, ?, ?);
```

### Configuration

```yaml
{
  "version": "1",
  "packages": [
    {
      "path": "db",
      "engine": "mysql",
      "schema": "query.sql",
      "queries": "query.sql"
    }
  ]
}
```

### Playground URL

https://play.sqlc.dev/p/ffd4ea54087e3ade2e02ae5ea9c9d76f00a9f34d0f649544c8ac716a0c1cfed0

### What operating system are you using?

macOS

### What database engines are you using?

MySQL

### What type of code are you generating?

Go

p1ass:
I misunderstood the problem.
It was working well adding an option `"sql_driver": "github.com/go-sql-driver/mysql"` as below playground.

https://play.sqlc.dev/p/9bf65cf825740ae06e585a0a25bc6a48154a660080d8cbb1af1a45d813255b21

kyleconroy:
As you discovered, you need to setthe `sql_driver` configuration option. We'll update the docs to make this clearer.

PR #2692, opened 2023-09-04
fix `parseDriver` function to return a correct Driver when using MySQL

p1ass:
## Related Issue

#2690 

## Description

The former `parseDrive`r function only returns PostgreSQL drivers, so it never returns go-sql-driver.
This behavior makes it necessary to specify an option `sql_driver` whenever we use `:copyfrom`.

This PR changes the parseDriver function's behavior to return a correct driver according to the specified DB engine.
As a result of this change, we will not need to add the option.

### DIFF
--- a/internal/codegen/golang/driver.go
+++ b/internal/codegen/golang/driver.go
@@ -1,5 +1,7 @@
 package golang
 
+import "github.com/sqlc-dev/sqlc/internal/config"
+
 type SQLDriver string
 
 const (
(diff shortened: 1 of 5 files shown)


orisano:
@kyleconroy Why did you CLOSE this PR?
I think github.com/go-sql-driver/mysql should be inferred if the engine is mysql even if sql_driver is not explicitly stated.
I think it is just a workaround to specify sql_driver to avoid #2690.

PR #2708, opened 2023-09-06
fix(codegen/golang): Fix detection of querier used packages

pespantelis:
Hey there! First of all, I want to drop a quick thank you for this awesome project, it's a blast working with it!

After the `v1.21.0` release, I noticed that the `querier.go` file imports the `time` package, without any reference on that package.

This PR addresses the issue, but since I'm relatively new to this project, I'm not entirely certain if it's the ideal solution. Your feedback and guidance would be greatly appreciated!

Fixes https://github.com/sqlc-dev/sqlc/issues/2707


kyleconroy:
@pespantelis Could you build sqlc from main and verify that this has been fixed?

pespantelis:
@kyleconroy Yep, it works for my case. Thanks!

PR #2727, opened 2023-09-14
Enables the configuration of copyfrom.go similar to quierer and friends

XanderAtGoingDutch:
This pr fixes a TODO by `Jille` that enables the user to use a different filename for copyfrom.go.


XanderAtGoingDutch:
When all checks pass can this be merged then? or do i need to do something yet?

kyleconroy:
The checks have passed, but the added test doesn't verify that the configuration option works. Can you update that test to generate a copyfrom file?

PR #2735, opened 2023-09-15
Add `emit_sql_as_comment` option to Go code plugin

jasoncabot:
This option adds the raw SQL query as a comment to the generated query function, defaulted to false.

This is useful when working in an IDE that displays comments over functions, you are able to glance at the actual SQL query that will be executed without having to lose context of where you are in the current file you are working on.

If you have a query with comments already, it will add the query as a [Code Block](https://tip.golang.org/doc/comment) underneath 

Input:
```
-- name: ListBar :many
-- Lists all bars
SELECT id FROM (
  SELECT * FROM bar
) bar;
```
Output:
```
// Lists all bars
//
//	SELECT id FROM (
//	  SELECT id FROM bar
//	) bar
func (q *Queries) ListBar(ctx context.Context) ([]int32, error) {
```

Example of how this shows up in VS Code when referencing the function:

image

If you haven't already got a comment, it will add the name of the function - I wasn't 100% sure about this but it allows the godoc comment to still have valid formatting when running `gofmt`

Input:
```
-- name: RemoveBar :exec
DELETE FROM bar WHERE id = $1;
```

Output:
```
// RemoveBar
//
//	DELETE FROM bar WHERE id = $1
func (q *Queries) RemoveBar(ctx context.Context, id int32) error {
```

Example of how this shows up in VS Code when referencing the function:

image


andrewmbenton:
Sorry for the delayed response. I understand the improved ergonomics of having the query in a godoc for certain editors but adding yet another configuration flag for a small addition to our templates system isn't a trade-off we want to make at the moment.

Ideally our Go codegen would be extensible enough that you could add this without changing anything about `sqlc` itself, but sadly it isn't that flexible yet.

andrewmbenton:
We've now had at least one other request for this feature, so I'm going to reopen and attempt to rebase in order to merge.

andrewmbenton:
> Thanks for picking this up again. I actually had a rebased (on v1.24) version on my fork at [jasoncabot@aa15d6a](https://github.com/jasoncabot/sqlc/commit/aa15d6ab5a0f1626e6e6eb45d915ec1d3938eece)

Oh excellent. I will close my PR if you pull your rebased changes into this one.

PR #2740, opened 2023-09-16
fix(compiler): Improved type inference for concat operator

### LINKED ISSUE #2734: sqlc >= v1.18.0 detects a string concatenation as `sql.NullBool`
### Version

1.21.0

### What happened?

I've just done a few version bumps up to v1.21.0 and noticed that:

```sql
select
-- ...
db.engine || ' ' || db.engine_version as runtime
```

Ends up being converted to an `sql.NullBool`.

The upgrade diff for the generate structs can be found below:

```diff
 // Code generated by sqlc. DO NOT EDIT.
 // versions:
-//   sqlc v1.17.0
+//   sqlc v1.21.0
 // source: queries.sql
 
 package db
@@ -198,21 +198,11 @@ where
 `
 
 type RetrieveAWSAdvisoriesRow struct {
-       Arn            string
-       Name           string
-       Runtime        interface{}
-       AdvisoryType   interface{}
-       Description    interface{}
-       Arn_2          string
-       Name_2         string
-       Runtime_2      string
-       AdvisoryType_2 interface{}
-       Description_2  interface{}
-       Arn_3          string
-       Name_3         string
-       Runtime_3      interface{}
-       AdvisoryType_3 interface{}
-       Description_3  interface{}
+       Arn          string
+       Name         string
+       Runtime      sql.NullBool
+       AdvisoryType interface{}
+       Description  interface{}
 }
```

Previously it was an `interface{}` or a `string`, whereas now it's a `sql.NullBool` :thinking: 

I've had a look back on the Playground as far back as v1.18.0 ([link](https://play.sqlc.dev/p/4644276ab2d6a08adc69359a99b241c23a0ce108f0a01d9bb937a81d584501ef)) and it affects that as well as up to v1.21.0 (with #2729 applied locally) 

### Relevant log output

_No response_

### Database schema

_No response_

### SQL queries

_No response_

### Configuration

_No response_

### Playground URL

https://play.sqlc.dev/p/4644276ab2d6a08adc69359a99b241c23a0ce108f0a01d9bb937a81d584501ef

Linux

SQLite

Go

orisano:
close #2734

Fixed sqlite engine treating "||" as OR.
MySQL does not have a concat operator, so there is no need to support it.


andrewmbenton:
This might be a bigger issue that we need to handle in a separate PR, but based on my reading of the SQLite docs and experimenting with a running PostgreSQL database it appears the the string concat operator (and I think all operators) will return `NULL` if any of the inputs is null.

From https://www.sqlite.org/lang_expr.html:

> The result of any binary operator is either a numeric value or NULL, except for the || concatenation operator,
> and the -> and ->> extract operators which evaluate to either NULL or a text value.
>
> All operators generally evaluate to NULL when any operand is NULL, with specific exceptions as stated below.
> This is in accordance with the SQL92 standard.

So I guess I would expect `sqlc` to output nullable types for outputs, but that should perhaps be addressed for all operators in a new PR.

andrewmbenton:
@orisano I didn't mean to imply that we should close this. It might be the correct first step to at least fix the treatment of `||` to make it equivalent to other binary operators. And then we could create a new issue for nullability of all binary operators. What do you think?

PR #2744, opened 2023-09-20
test(endtoend): Verify all schemas in endtoend

kyleconroy:
For each test in the endtoend package, use the new sqlc managed databases to verify that the schemas are valid. In the future, we'll attempt to prepare all the queries as well.


internal/quickdb/rpc.go
```diff
@@ -0,0 +1,59 @@
...
+
+func NewClient(project, token string, opts ...Option) (pb.QuickClient, error) {
+	var o options
+	for _, apply := range opts {
+		apply(&o)
+	}
+
+	dialOpts := []grpc.DialOption{
+		grpc.WithTransportCredentials(credentials.NewTLS(&tls.Config{
+			InsecureSkipVerify: true,
```
andrewmbenton, line 43: You should be able to use a default `tls.Config{}` here

andrewmbenton (changes requested):

andrewmbenton (approved):

PR #2751, opened 2023-09-21
feat(vet): Run rules against a managed database

kyleconroy:
The newest `vet` feature is running rules against databases hosted by us. You don't need to have PostgreSQL running locally, we'll take care of everything for you. This feature will be included in the next release as a private beta. We're still working out kinks in our infrastructure.

```yaml
version: '2'
cloud:
  project: "01HAQMMECEYQYKFJN8MP16QC41"
sql:
- schema: postgresql/schema.sql
  queries: postgresql/query.sql
  engine: postgresql
  database:
    managed: true
  rules:
  - sqlc/db-prepare
```

### DIFF
--- a/internal/cmd/cmd.go
+++ b/internal/cmd/cmd.go
@@ -37,6 +37,7 @@ func Do(args []string, stdin io.Reader, stdout io.Writer, stderr io.Writer) int 
 	rootCmd.PersistentFlags().StringP("file", "f", "", "specify an alternate config file (default: sqlc.yaml)")
 	rootCmd.PersistentFlags().BoolP("experimental", "x", false, "DEPRECATED: enable experimental features (default: false)")
 	rootCmd.PersistentFlags().Bool("no-remote", false, "disable remote execution (default: false)")
+	rootCmd.PersistentFlags().Bool("remote", false, "enable remote execution (default: false)")
 	rootCmd.PersistentFlags().Bool("no-database", false, "disable database connections (default: false)")
 
 	rootCmd.AddCommand(checkCmd)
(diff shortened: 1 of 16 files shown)


internal/cmd/cmd.go
```diff
@@ -37,6 +37,7 @@ func Do(args []string, stdin io.Reader, stdout io.Writer, stderr io.Writer) int
 	rootCmd.PersistentFlags().StringP("file", "f", "", "specify an alternate config file (default: sqlc.yaml)")
 	rootCmd.PersistentFlags().BoolP("experimental", "x", false, "DEPRECATED: enable experimental features (default: false)")
 	rootCmd.PersistentFlags().Bool("no-remote", false, "disable remote execution (default: false)")
+	rootCmd.PersistentFlags().Bool("remote", false, "enable remote execution (default: false)")
```
kyleconroy, line 40: Now that `vet` is also using the project configuration, we can't default to remote generation. Instead, allow users to opt in with the `--remote` flag.

internal/endtoend/vet_test.go
```diff
@@ -51,12 +51,8 @@ func TestExamplesVet(t *testing.T) {
 			path := filepath.Join(examples, tc)
 
 			if tc != "kotlin" && tc != "python" {
-				if s, found := findSchema(t, filepath.Join(path, "postgresql")); found {
-					db, cleanup := sqltest.CreatePostgreSQLDatabase(t, tc, false, []string{s})
-					defer db.Close()
-					defer cleanup()
-				}
 				if s, found := findSchema(t, filepath.Join(path, "mysql")); found {
+					t.Skip("local")
```
kyleconroy, line 55: TODO: Remove this before merging

internal/config/config.go
```diff
@@ -69,7 +69,8 @@ type Project struct {
 }
 
 type Database struct {
-	URI string `json:"uri" yaml:"uri"`
+	URI     string `json:"uri" yaml:"uri"`
+	Managed bool   `json:"managed" yaml:"managed"`
```
andrewmbenton, line 73: Need to add this to our json schemas too right?

PR #2753, opened 2023-09-21
docs: clarify nullable override behavior

andrewmbenton:
* also remove some dead code

### DIFF
--- a/docs/reference/config.md
+++ b/docs/reference/config.md
@@ -457,6 +457,7 @@ packages:
     queries: "./sql/query/"
     schema: "./sql/schema/"
     engine: "postgresql"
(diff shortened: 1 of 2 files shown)


daison12006013:
@andrewmbenton is there a documentation as well if we can re-use a certain override block in sqlc.yaml?

```yaml
overrides:
  go:
    overrides:
      - column: "profiles.payment_flag"
      - column: "profileviews.is_active_member"
      - column: "profileviews.is_active_profile"
        nullable: true
        go_type:
          import: "gopkg.in/guregu/null.v4"
          package: "null"
          type: "Bool"
```

As an example above, where in these 3 columns will have a `null.Bool`

PR #2779, opened 2023-09-26
feat: wildcard in queries config

Dionid:
We've got two databases and the same code need to use both of them. So the best way will be:

```
/authenticate
  /authenticate.go
  /query.firstdb.sql
  /query.seconddb.sql
```

And in config we can write:

```
version: "2"
sql:
  - engine: "postgresql"
    queries:
      - "./firstdb/"
      - "./features/*.firstdb.sql"
    schema: "./firstdb/schema.sql"
    gen:
      go:
        package: "firstdb"
        out: "firstdb"
  - engine: "postgresql"
    queries:
      - "./seconddb/schema.sql"
      - "./features/*.seconddb.sql"
    schema: "./seconddb/schema.sql"
    gen:
      go:
        package: "seconddb"
        out: "seconddb"
```

So wildcard in queries helps us to distinguish between sql to first or second DB.

### DIFF
--- a/internal/sql/sqlpath/read.go
+++ b/internal/sql/sqlpath/read.go
@@ -13,21 +13,27 @@ import (
 // in .sql. Omits hidden files, directories, and migrations.
 func Glob(paths []string) ([]string, error) {
 	var files []string
-	for _, path := range paths {
-		f, err := os.Stat(path)
+	for _, pathForGlob := range paths {
+		globPaths, err := filepath.Glob(pathForGlob)
 		if err != nil {
-			return nil, fmt.Errorf("path %s does not exist", path)
+			return nil, err
 		}
-		if f.IsDir() {
-			listing, err := os.ReadDir(path)
+		for _, path := range globPaths {
+			f, err := os.Stat(path)
 			if err != nil {
-				return nil, err
+				return nil, fmt.Errorf("path %s does not exist", path)
 			}
-			for _, f := range listing {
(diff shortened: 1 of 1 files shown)


kyleconroy:
Thanks for the PR, but this isn't a feature we're going to support at this time. I suggest using a separate subfolder for each of the queries like so:

```yaml
version: "2"
sql:
  - engine: "postgresql"
    queries:
      - "./firstdb/"
      - "./features/firstdb"
    schema: "./firstdb/schema.sql"
    gen:
      go:
        package: "firstdb"
        out: "firstdb"
  - engine: "postgresql"
    queries:
      - "./seconddb/schema.sql"
      - "./features/seconddb"
    schema: "./seconddb/schema.sql"
    gen:
      go:
        package: "seconddb"
        out: "seconddb"
```

Dionid:
@kyleconroy  The problem is i want my queries to be near my business logic, but in one use case can be used both DBs.

IMHO separating queries from business logic is a bad practice (sorry, Uncle Bob) and current version forces me to do this.

There must be a way to store queries near code.

Dionid:
Oh, i see that there are also a lot of my fork changes... The only file to support this feature is [this](https://github.com/sqlc-dev/sqlc/pull/2779/files#diff-e60925a8bc3e7452522c00a14f6291010d27ede00fbc9e4be800d20b812b41a5)

PR #2805, opened 2023-10-05
feat(analyzer): Analyze queries using a running PostgreSQL database

kyleconroy:
[94 of the open issues](https://github.com/sqlc-dev/sqlc/issues?q=is%3Aissue+is%3Aopen+label%3Aanalyzer) on sqlc are related to the analyzer. There are many cases where the current analyzer produces false positives or false negatives. 

[sqlx](https://github.com/launchbadge/sqlx) and some other projects have proven that it's possible to extract query metadata from a running database.

This approach is a bit different, in that the database analysis is layered on top of the existing query analyzer. We use the existing analysis for better name and NULL inference. 

This PR is an obvious work in progress, but I'm happy enough with the result to share an early peek.


internal/engine/postgresql/pg_catalog.go
```diff
@@ -29072,6 +29072,19 @@ var funcsgenPGCatalog = []*catalog.Function{
...
+			{
+				Type: &ast.TypeName{Name: "anyarray"},
+			},
+			{
+				Type: &ast.TypeName{Name: "anyarray"},
+				Mode: ast.FuncParamVariadic,
+			},
+		},
+		ReturnType: &ast.TypeName{Name: "anyelement"},
+	},
```
kyleconroy, line 29087: I'm going to remove this, as we shouldn't be updating this file.

internal/sql/validate/func_call.go
```diff
@@ -31,34 +30,8 @@ func (v *funcCallVisitor) Visit(node ast.Node) astutils.Visitor {
 		return v
 	}
 
-	// Custom validation for sqlc.arg, sqlc.narg and sqlc.slice
-	// TODO: Replace this once type-checking is implemented
+	// Maybe not necessary now?
```
andrewmbenton, line 33: Can remove

internal/compiler/resolve.go
```diff
@@ -10,6 +10,7 @@ import (
 	"github.com/sqlc-dev/sqlc/internal/sql/named"
 	"github.com/sqlc-dev/sqlc/internal/sql/rewrite"
 	"github.com/sqlc-dev/sqlc/internal/sql/sqlerr"
+	"golang.org/x/exp/slog"
```
andrewmbenton, line 13: We should be able to use stdlib `slog` right? We're building with go 1.21?

internal/config/config.go
```diff
@@ -64,8 +64,9 @@ type Config struct {
 }
 
 type Database struct {
-	URI     string `json:"uri" yaml:"uri"`
-	Managed bool   `json:"managed" yaml:"managed"`
+	URI      string `json:"uri" yaml:"uri"`
+	Managed  bool   `json:"managed" yaml:"managed"`
+	Analyzer *bool  `json:"analyzer" yaml:"analyzer"`
```
andrewmbenton, line 69: We may regret putting this here? I think we should discuss for a min.

internal/endtoend/endtoend_test.go
```diff
@@ -95,52 +110,103 @@ func TestReplay(t *testing.T) {
...
+
+				var diff string
+				for _, expectedErr := range strings.Split(expected, "---\n") {
+					diff = cmp.Diff(strings.TrimSpace(expectedErr), strings.TrimSpace(stderr.String()))
+					if diff == "" {
+						break
+					}
+				}
+				if diff != "" {
+					t.Log(stderr.String()) // TODO: Remove
```
andrewmbenton, line 205: remove

internal/endtoend/endtoend_test.go
```diff
@@ -12,13 +12,15 @@ import (
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
+	"golang.org/x/exp/slices"
```
andrewmbenton, line 15: `slices` is in stdlib now

internal/endtoend/endtoend_test.go
```diff
@@ -95,52 +110,103 @@ func TestReplay(t *testing.T) {
...
+				return len(os.Getenv("SQLC_AUTH_TOKEN")) > 0
+			},
+		},
+	}
+
 	for _, replay := range dirs {
 		tc := replay
-		t.Run(tc, func(t *testing.T) {
-			t.Parallel()
+		for name, testctx := range contexts {
```
andrewmbenton, line 139: I think this might be simpler if we set the default contexts explicitly in `parseExec` when none are given, like we do with command. Then we could move this for loop inside `t.Run` and just iterate over the contexts we have in `args.Contexts`, and we wouldn't need to `Skipf` any contexts.

But I guess if you explicitly want an independent test for each context then that doesn't work.

internal/endtoend/endtoend_test.go
```diff
@@ -95,52 +110,103 @@ func TestReplay(t *testing.T) {
...
+					t.Fatalf("sqlc %s failed: %s", args.Command, stderr.String())
+				}
+
+				var diff string
+				for _, expectedErr := range strings.Split(expected, "---\n") {
+					diff = cmp.Diff(strings.TrimSpace(expectedErr), strings.TrimSpace(stderr.String()))
+					if diff == "" {
+						break
+					}
+				}
```
andrewmbenton, line 203: This isn't ideal, but might not be worth the effort to clean up. Just matching **some** stderr output is worse than matching the actual expected output right? There are a couple of ways I can think of to have the expected output depend on the context explicitly (separate files with magic names, or demarcating with `"--- <context-name>\n"`), but I'm wondering what you think.

internal/config/config.go
kyleconroy:
```yaml
analyzer:
  database: false
```

Instad

Issue #2012, opened 2023-01-01
Being able to mark generated files with go:build tags


eladitzhakian:
### What do you want to change?

It could be useful to be able to emit `// go:build some-tag` into the golang generated files

This way I can, for example, exclude them when running coverage tests 😇 

### What programming language backends need to be changed?

Go

andrewmbenton:
This seems like a pretty easy option to add to the [sqlc v2 gen/go config](https://docs.sqlc.dev/en/latest/reference/config.html#go), if someone wants to take a crack at it.

Hyuga-Tsukui:
I find this feature very useful, and I'd like to try contributing. Would that be okay?

Hyuga-Tsukui:
@kyleconroy 

I find this feature very useful, and I'd like to try contributing. Would that be okay?

PR #2807, opened 2023-10-05
feat(codegen): add support for build tags (#2012)

Hyuga-Tsukui:
Attempt at adding the feature for #2012.

I am not a native English speaker, so some variable names might not be appropriate.

I've added a new `build_tags` option, which gets inserted into the template when enabled.

Fixes https://github.com/sqlc-dev/sqlc/issues/2012


docs/reference/config.md
```diff
@@ -146,6 +146,8 @@ The `gen` mapping supports the following keys:
 - `emit_all_enum_values`:
   - If true, emit a function per enum type
     that returns all valid enum values.
+- `build_tags`:
+  - build tags to add to the generated code. if you do not want to add it, leave it unset.
```
kyleconroy, line 150:
```suggestion
  - If set, add a `//go:build <build_tags>` directive at the beginning of each generated Go file.
```

docs/reference/config.md
```diff
@@ -443,6 +446,8 @@ Each mapping in the `packages` collection has the following keys:
...
```
kyleconroy, line 450:
```suggestion
  - If set, add a `//go:build <build_tags>` directive at the beginning of each generated Go file.
```

internal/codegen/golang/gen.go
```diff
@@ -38,6 +38,8 @@ type tmplCtx struct {
 	EmitAllEnumValues         bool
 	UsesCopyFrom              bool
 	UsesBatch                 bool
+
+	BuildTags string
```
kyleconroy, line 42:
```suggestion
	BuildTags string
```

internal/codegen/golang/templates/template.tmpl
```diff
@@ -1,4 +1,9 @@
-{{define "dbFile"}}// Code generated by sqlc. DO NOT EDIT.
+{{define "dbFile"}}
+{{if .BuildTags}}
+//go:build {{.BuildTags}}
+// +build {{.BuildTags}}
```
kyleconroy, line 4: The `// +build` syntax has been deprecated since Go 1.17. Let's remove it and only support `//go:build`.

protos/plugin/codegen.proto
```diff
@@ -78,6 +78,7 @@ message GoCode {
   bool emit_result_struct_pointers = 8;
   bool emit_params_struct_pointers = 9;
   bool emit_methods_with_db_argument = 10;
+  string build_tags = 29;
```
kyleconroy, line 81: Can you add this to the end of the message instead of in the middle?

kyleconroy (changes requested):
Thanks! Only one real change before we can merge this in. Appreciate you adding a test case.

Hyuga-Tsukui:
Thank you for the review. I've made the corrections, so please check again.

kyleconroy (approved):

PR #2822, opened 2023-10-11
refactor(codegen): remove golang and json settings from plugin proto

andrewmbenton:
In order to simplify the codegen plugin proto, I've removed the Go and JSON plugin settings and instead pass those settings encoded as JSON to their respective plugins.

I introduce a new struct type, `opts` in each codegen package, for deserialization of the passed settings. These are essentially the Go types formerly generated from the proto, but now isolated.


internal/cmd/generate.go
```diff
@@ -399,10 +400,20 @@ func codegen(ctx context.Context, combo config.CombinedSettings, sql outPair, re
...
 
 	case sql.Gen.JSON != nil:
 		out = combo.JSON.Out
-		handler = ext.HandleFunc(json.Generate)
+		handler = ext.HandleFunc(genjson.Generate)
+		opts, err := json.Marshal(sql.Gen.JSON)
+		if err != nil {
+			return "", nil, fmt.Errorf("opts marshal failed: %w", err)
+		}
+		req.Settings.Codegen.Options = opts
```
andrewmbenton, line 416: I would've stuck the opts in `req.PluginOptions` like I did with the Go case, but there was already some parsing code within the json codegen package that looks for opts in this other spot.

internal/codegen/golang/gen.go
```diff
@@ -103,55 +103,60 @@ func (t *tmplCtx) codegenQueryRetval(q Query) (string, error) {
...
-	queries, err := buildQueries(req, structs)
+	options, err := parseOpts(req)
 	if err != nil {
 		return nil, err
 	}
 
-	if req.Settings.Go.OmitUnusedStructs {
+	enums := buildEnums(req, options)
+	structs := buildStructs(req, options)
+	queries, err := buildQueries(req, options, structs)
```
andrewmbenton, line 113: I think some of these funcs won't need to take the `req` as a param at all, once we figure out how to handle overrides and renames.

internal/cmd/generate.go
kyleconroy: Can we update the JSON codegen package to look in PluginOptions? I think updating the Options instead the settings isn't the right place for this.

internal/codegen/golang/opts.go
```diff
@@ -0,0 +1,61 @@
...
+	OutputCopyfromFileName      string   `json:"output_copyfrom_file_name,omitempty"`
+	OutputFilesSuffix           string   `json:"output_files_suffix,omitempty"`
+	InflectionExcludeTableNames []string `json:"inflection_exclude_table_names,omitempty"`
+	QueryParameterLimit         *int32   `json:"query_parameter_limit,omitempty"`
+	OmitUnusedStructs           bool     `json:"omit_unused_structs,omitempty"`
+	BuildTags                   string   `json:"build_tags,omitempty"`
+
+	// Unused but left in for parsing convenience
+	Overrides json.RawMessage `json:"overrides,omitempty"`
+	Rename    json.RawMessage `json:"rename,omitempty"`
```
kyleconroy, line 44: Overrides and renames can be package specific. Will these still work?

PR #2846, opened 2023-10-13
feat(codegen): Include plugin information

kyleconroy:
Include the configured plugin's env, name, sha256 and URL

Fixes https://github.com/sqlc-dev/sqlc/issues/2468


kyleconroy:
@orisano Is this what you had in mind?

orisano:
Yes, this is exactly what I had in mind.

internal/cmd/shim.go
```diff
@@ -58,6 +58,7 @@ func pluginSettings(r *compiler.Result, cs config.CombinedSettings) *plugin.Sett
 		Overrides: over,
 		Rename:    cs.Rename,
 		Codegen:   pluginCodegen(cs.Codegen),
+		Plugin:    pluginPlugin(cs),
```
andrewmbenton, line 61: As discussed offline, probably better to move the elements of this into the existing `Codegen` field.

PR #2849, opened 2023-10-13
fix(golang): Escape reserved keywords

kyleconroy:
Append `_` to variable names that are reserved

Fixes https://github.com/sqlc-dev/sqlc/issues/2267


andrewmbenton:
Honestly I don't think I would even bother checking for reserved words, since that list is something we'll have to keep up to date. I would just append or prepend `"_"` to every variable name since they are all internal to `sqlc`'s funcs anyway.

kyleconroy:
The list of reserved words in Go is very short and changes infrequently. I don't think they've added one for years.

PR #2850, opened 2023-10-13
feat(compiler): Parse query parameter metadata from comments

andrewmbenton:
Groundwork for https://github.com/sqlc-dev/sqlc/issues/2800 (explicit type annotations in query comments metadata).

There are lots of tests that failed while I was refactoring this, so I have a fairly high degree of confidence that this refactor doesn't break anything. The changed endtoend test output is due to parsing bugs I fixed.

I chose to add a new `Metadata` struct type to house the growing set of stuff we associate with a query, but I don't have to so let me know if an alternative approach is preferred.

### DIFF
--- a/internal/metadata/meta.go
+++ b/internal/metadata/meta.go
@@ -9,6 +9,8 @@ import (
 	"github.com/sqlc-dev/sqlc/internal/source"
 )
 
+type CommentSyntax source.CommentSyntax
+
 type Metadata struct {
 	Name     string
 	Cmd      string
@@ -18,8 +20,6 @@ type Metadata struct {
 
 	Filename string
 }
-
-type CommentSyntax source.CommentSyntax
 
 const (
 	CmdExec       = ":exec"
--- a/internal/metadata/meta.go
+++ b/internal/metadata/meta.go
@@ -58,7 +58,7 @@ func validateQueryName(name string) error {
 func ParseQueryMetadata(rawSql string, commentStyle CommentSyntax) (Metadata, error) {
 	md := Metadata{}
 	s := bufio.NewScanner(strings.NewReader(strings.TrimSpace(rawSql)))
-	var lines, comments []string
+	var comments []string
 	for s.Scan() {
 		line := s.Text()
 		var prefix string
@@ -81,7 +81,6 @@ func ParseQueryMetadata(rawSql string, commentStyle CommentSyntax) (Metadata, er
 			prefix = "#"
 		}
 		if prefix == "" {
-			lines = append(lines, line)
 			continue
 		}
 
--- a/internal/metadata/meta.go
+++ b/internal/metadata/meta.go
@@ -1,10 +1,21 @@
 package metadata
 
 import (
+	"bufio"
 	"fmt"
 	"strings"
 	"unicode"
 )
+
+type Metadata struct {
+	Name     string
+	Cmd      string
+	Comments []string
+	Params   map[string]string
+	Flags    map[string]bool
+
+	Filename string
+}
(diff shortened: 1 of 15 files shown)


internal/metadata/meta_test.go
```diff
@@ -16,8 +16,11 @@ func TestParseQueryNameAndType(t *testing.T) {
 		`--name: CreateFoo :two`,
```
andrewmbenton, line 16: I could write a much nicer implementation of the parse function (using a word scanner) if we treat `--name: ...` as either (1) valid name and cmd metadata or (2) not invalid, but just ignored and treated like a regular comment.

I will also point out that we don't special case the treatment of a missing space between comment start and `name:` with the other two supported comment syntaxes.

internal/metadata/meta.go
```diff
@@ -83,7 +92,7 @@ func ParseQueryNameAndType(t string, commentStyle CommentSyntax) (string, string
 		if prefix == "/*" {
 			part = part[:len(part)-1] // removes the trailing "*/" element
 		}
-		if len(part) == 2 {
+		if len(part) == 3 {
```
andrewmbenton, line 95: This was a bug, fixing resulted in some updated test output.

internal/source/code.go
```diff
@@ -111,7 +120,46 @@ func StripComments(sql string) (string, []string, error) {
 			comments = append(comments, t)
 			continue
 		}
+		if strings.HasPrefix(t, "#") {
+			comments = append(comments, strings.TrimPrefix(t, "#"))
+			continue
+		}
```
andrewmbenton, line 126: This (along with similar change above) was a bug, fixing resulted in one updated mysql test output.

internal/source/code.go
```diff
@@ -111,7 +120,46 @@ func StripComments(sql string) (string, []string, error) {
...
 		lines = append(lines, t)
 	}
 	return strings.Join(lines, "\n"), comments, s.Err()
 }
+
+func CleanedComments(rawSQL string, cs CommentSyntax) ([]string, error) {
```
andrewmbenton, line 132: We've now implemented comment parsing/cleaning in three places. I was hoping to consolidate but walked it back in favor of a separate code path.

It's some unfortunate technical debt though, since at some point someone is going to complain that we don't support multi-line slash-star comment syntax and we'll have to refactor all of this...

internal/source/code.go
```diff
@@ -15,6 +15,12 @@ type Edit struct {
 	OldFunc  func(string) int
 }
 
+type CommentSyntax struct {
```
andrewmbenton, line 18: Handling comments seems like a source code thing, so I popped this in the `source` package. The new `CleanedComments` func I wrote depends on comment syntax, and we already have a func `StripComments` (which doesn't depend on comment syntax but I think that's just an oversight/bug).

internal/source/code.go
kyleconroy: So a comment like

```
/*
name: Foo :one
@param bar? text
*/
```

won't work? If that's the case, we'll want to have a fast follow to enable that.

kyleconroy: Looking over the implementation of CleanedComments, I don't think it's be too hard to add support for multiline comments. But let's address that in a future PR.

kyleconroy (approved):

internal/source/code.go
andrewmbenton: Yeah the multi-line comment syntax has never worked but there wasn't really a reason for anyone to ask before this change.

Agree we can implement later

Issue #2800, opened 2023-10-03
Type annotations for query parameters


kyleconroy:
I propose adding type annotations for query parameters. Users will no longer need to cast parameters to the desired type. This proposal builds on Andrew's query annotation
work for `vet`.

## Unified syntax

We added the `@sqlc-vet-disable` annotation to disable vet rules on a per-query
basis. We can extend this syntax to support other per-query configuration
options.

The current syntax for query name and command are different, so we'll
standardize on the `@` prefix. This will be the new, preferred syntax for name
and command.

```sql
-- name: GetAuthor :one
SELECT * FROM authors
WHERE id = $1 LIMIT 1;

-- becomes

-- @name GetAuthor :one
SELECT * FROM authors
WHERE id = $1 LIMIT 1;
```

The existing syntax will continue to work, but it will be an error to use both
annotations on a single query.

```sql
-- INVALID!
-- name: GetAuthor :one
-- @name GetAuthor :one
SELECT * FROM authors
WHERE id = $1 LIMIT 1;
```

## Query command

Today, queries must have a name and a command. With the new syntax, the
command option will default to `exec`.

```sql
-- These two annotations are the same

-- @name DeleteAuthor :exec
DELETE FROM authors
WHERE id = $1;

-- @name DeleteAuthor
DELETE FROM authors
WHERE id = $1;
```

### Validation

sqlc will delegate command validation to codegen plugins, allowing plugins to implement new commands without having
to merge anything into sqlc.

If you want to still validate those, you can simulate the current behavior by
using this vet rule.

```yaml
rules:
  - name: validate-command
    rule: |
      !(query.cmd in ["exec", "one", "many", "execrows", "execlastid", "execresult", "batchone", "batchmany", "batchexec"])
```

### Type annotations for query parameters

The `@param` annotation supports passing type information without having to use
a cast. 

```sql
-- @param name type
```

Casts are required today when sqlc infers an incorrect type, but these
casts are passed down to the engine itself, possibly hurting performance. 
For example, this cast is required to get sqlc working correctly, but isn't needed at runtime.

```sql
-- @name ListAuthorsByIDs :many
SELECT * FROM authors
WHERE id = ANY(@ids::bigint[]);
```
Here's what it looks like with the new syntax. The type annotation is
engine-specific and is the same that you'd pass to CAST or CREATE TABLE.

```sql
-- @name ListAuthorsByIDs :many
-- @param ids BIGINT[] NOT NULL
[...]

PR #2852, opened 2023-10-13
fix(mysql): Handle simplified CASE statements

kyleconroy:
MySQL and PostgreSQL now have the same behavior, generating a nullable string parameter. That behavior isn't right, as we'd expect the parameter to be a boolean. This can be fixed by using a type case or the new annotations in #2800.

Fixes https://github.com/sqlc-dev/sqlc/issues/2847


internal/endtoend/testdata/case_value_param/mysql/go/query.sql.go
```diff
@@ -0,0 +1,21 @@
...
+	"context"
+	"database/sql"
+)
+
+const update = `-- name: Update :exec
+UPDATE testing
+SET value = CASE ? WHEN true THEN 'Hello' WHEN false THEN 'Goodbye' ELSE value END
+`
+
+func (q *Queries) Update(ctx context.Context, value sql.NullString) error {
```
andrewmbenton, line 18: Shouldn't this generated func parameter be a boolean to match the cases and not a string which matches the output?

internal/engine/dolphin/convert.go
```diff
@@ -745,6 +745,7 @@ func (c *cc) convertCaseExpr(n *pcast.CaseExpr) ast.Node {
 		list.Items = append(list.Items, c.convertWhenClause(n))
 	}
 	return &ast.CaseExpr{
+		Arg:       c.convert(n.Value),
```
andrewmbenton, line 748: After looking at the test output, I think `n.Value` isn't correct here.

PR #2853, opened 2023-10-13
fix(mysql): Add, drop, rename and change enum values

kyleconroy:
Extends the work done in #2680 to support adding and dropping columns and tables.

Fixes https://github.com/sqlc-dev/sqlc/issues/2475


examples/kotlin/src/main/kotlin/com/example/ondeck/mysql/Models.kt
```diff
@@ -7,12 +7,12 @@ package com.example.ondeck.mysql
 import java.sql.Timestamp
 import java.time.Instant
 
-enum class VenuesStatus(val value: String) {
+enum class VenueStatus(val value: String) {
```
kyleconroy, line 10: This name changed because the table is called `venue` not `venues`. The changed name wasn't taken into account before.

examples/ondeck/mysql/models.go
```diff
@@ -11,46 +11,46 @@ import (
 	"time"
 )
 
-type VenuesStatus string
+type VenueStatus string
```
kyleconroy, line 14: This name changed because the table is called `venue` not `venues`. The changed name wasn't taken into account before.

PR #2858, opened 2023-10-15
Improve nullability evaluation

lorenz:
This now supports evaluating nullability if subqueries, function-defined tables or sqlc.embed macros are involved.
If nullable embeds are generated, this now errors out instead of emitting code which breaks at runtime.

Generating correct type definitions for that case would be easy, but scanning data into embed pointers is hard as it would require "peeking" at row nullability to decide if the type should be allocated and scanned into. Otherwise the null columns need to be directed into some kind of sink. Because this is nontrivial to do this is not part of this PR.


jaredweinfurtner:
@lorenz FYI you still have conflicts if you wanted to follow up on it.

kyleconroy:
**🤖 Automated message**

---

PR #2869, opened 2023-10-17
refactor(codegen/go): Remove `rename` field from codegen proto

andrewmbenton:
Prefer instead to pass the rename map to the internal golang codegen plugin via `options`.

This diff is mostly large because I also pull golang options into a subpackage. There are only three or four meaningful changes, all in `protos`, `internal/cmd`, and `internal/codegen/golang/options/options.go`.


internal/cmd/generate.go
```diff
@@ -418,6 +418,7 @@ func codegen(ctx context.Context, combo config.CombinedSettings, sql outPair, re
 	case sql.Gen.Go != nil:
 		out = combo.Go.Out
 		handler = ext.HandleFunc(golang.Generate)
+		sql.Gen.Go.Rename = combo.Rename
```
andrewmbenton, line 421: I'm not certain this assignment is the best idea but it was the simplest change that worked. Combined settings should have the merged rename map from global and package levels.

andrewmbenton:
After discussion, I'm withdrawing this PR since `rename` is fine where it is in the plugin proto for now.

PR #2870, opened 2023-10-17
patch: use docker compose v2 and update MYSQL_DATABASE env var

juanjuanzero:
small changes to development.md file and makefile
- updates to use [docker compose v2](https://docs.docker.com/compose/)
- fixes environment variable for MYSQL_DATABASE was causing failing tests when running `make test-examples`

I'm new to this repo and to contributing, please let me know if i had missed anything.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,7 @@ sqlc-gen-json:
 	go build -o ~/bin/sqlc-gen-json ./cmd/sqlc-gen-json
 
 start:
-	docker-compose up -d
+	docker compose up -d
 
 fmt:
 	go fmt ./...
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,7 +6,7 @@ services:
       - "3306:3306"
     restart: always
     environment:
-      MYSQL_DATABASE: mysql
+      MYSQL_DATABASE: dinotest
       MYSQL_ROOT_PASSWORD: mysecretpassword
       MYSQL_ROOT_HOST: '%'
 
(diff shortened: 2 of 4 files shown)


andrewmbenton:
Thanks for the contribution! The docker compose change is good.

The docs regarding MySQL connection string variables represent the defaults as encoded in https://github.com/sqlc-dev/sqlc/blob/main/internal/sqltest/mysql.go so we shouldn't change those (although the formatting of that "table" could be improved). I think it would be appropriate to add a note at the bottom of that section explaining that you will need to create the `dinotest` database or set the environment variable to a non-default value in order for the tests to run.

juanjuanzero:
> Thanks for the contribution! The docker compose change is good.
> 
> The docs regarding MySQL connection string variables represent the defaults as encoded in https://github.com/sqlc-dev/sqlc/blob/main/internal/sqltest/mysql.go so we shouldn't change those (although the formatting of that "table" could be improved). I think it would be appropriate to add a note at the bottom of that section explaining that you will need to create the `dinotest` database or set the environment variable to a non-default value in order for the tests to run.

Thanks for the warm welcome and the feedback. I see, we do use `dinotest` as a default. I'll update the docker-compose file instead to set the `MYSQL_DATABASE` as `dinotest` instead of `mysql` unless that is intended for something else.

andrewmbenton:
> I'll update the docker-compose file instead to set the `MYSQL_DATABASE` as `dinotest` instead of `mysql` unless that is intended for something else.

Hmm this brings up a good point. I'm not sure why we don't have the containers start up with the `dinotest` database created by default. Looks like @kyleconroy just merged this so I guess he thinks it makes sense :)

Although then I think we should do it for the postgres containers too. And while we're at it, we could change the name of the default db from `dinotest` to `sqlctest` (`dinosql` was the original name of `sqlc`).

PR #2871, opened 2023-10-17
feat(postgresql): Support system columns on tables

kyleconroy:
Fixes https://github.com/sqlc-dev/sqlc/issues/1745


internal/endtoend/testdata/select_system/pgx/sqlc.yaml
```diff
@@ -0,0 +1,14 @@
+version: "2"
+cloud:
+  project: "01HAQMMECEYQYKFJN8MP16QC41"
```
andrewmbenton, line 3: We're still replacing this and the `database` config with test context in exec.json no?

kyleconroy: D'oh, removed

PR #2873, opened 2023-10-17
feat(compiler): Support LEFT JOIN on aliased table

kyleconroy:
Fixes https://github.com/sqlc-dev/sqlc/issues/1897


internal/compiler/output_columns.go
```diff
@@ -423,10 +424,12 @@ const (
...
-			return prior
+		tableMatch := *n.Relname == col.Table.Name
+		aliasMatch := true
+		if n.Alias != nil && col.TableAlias != "" {
+			aliasMatch = *n.Alias.Aliasname == col.TableAlias
 		}
-		if n.Alias != nil && *n.Alias.Aliasname == col.TableAlias && *n.Relname == col.Table.Name {
+		if aliasMatch && tableMatch {
 			return prior
 		}
```
kyleconroy, line 434: If the column doesn't use a table alias, don't check that the aliases match.

andrewmbenton: Seems correct logically, although `aliasMatch` defaulting to true is potentially confusing to future readers. I might suggest just putting this comment you wrote in the actual source.

PR #2882, opened 2023-10-18
fix(analyzer): Error on unexpanded star expression

kyleconroy:
If a query includes a star reference, error if we're unable to expand that reference. The new analyzer will happily anaylyze that code. If executed at runtime, these queries could change values.

Fixes https://github.com/sqlc-dev/sqlc/issues/2880


prog8:
Hi @kyleconroy,

I'm experiencing an issue with sqlc when trying to infer the schema solely from my database, without relying on schema files. The reason for this approach is that Postgres supports a wide range of features, including function overloading, which can be challenging for sqlc to fully support.

When I attempt to fetch data using a query like `SELECT * FROM table`, I encounter an error stating `star expansion failed for query`. However, if I comment out the line that triggers this error, everything works as expected.

I'm wondering if there's a workaround or a nice way to handle the star expansion feature in this scenario. Any guidance would be greatly appreciated.

Thanks in advance.

PR #2885, opened 2023-10-19
feat: Improve messaging for common cloud config and rpc errors

andrewmbenton:
This turned out to have more changes than I expected. A high-level list follows. There are definitely some style choices I made here that I'm not at all attached to, so feel free to recommend changes.

1. Add a UnaryClientInterceptor for all gRPC client connections to catch and rewrite unauthenticated error messages

2. Put the sqlc auth token in the config struct, and add simple validation

3. Add more explanatory error messages in cases where users try to use cloud features without the proper configuration

4. Prevent sending the SQLC_AUTH_TOKEN env var to plugins

Resolves https://github.com/sqlc-dev/sqlc/issues/2881

### DIFF
--- a/internal/config/env.go
+++ b/internal/config/env.go
@@ -1,0 +1,17 @@
+package config
+
+import (
+	"fmt"
+	"os"
+	"strings"
+)
+
+func (c *Config) addEnvVars() error {
+	authToken := os.Getenv("SQLC_AUTH_TOKEN")
+	if authToken != "" && !strings.HasPrefix(authToken, "sqlc_") {
+		return fmt.Errorf("$SQLC_AUTH_TOKEN doesn't start with \"sqlc_\"")
+	}
+	c.Cloud.AuthToken = authToken
+
+	return nil
+}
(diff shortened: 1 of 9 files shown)


internal/config/env.go
```diff
@@ -0,0 +1,17 @@
...
+
+func (c *Config) addEnvVars() error {
+	authToken := os.Getenv("SQLC_AUTH_TOKEN")
+	if authToken != "" && !strings.HasPrefix(authToken, "sqlc_") {
+		return fmt.Errorf("$SQLC_AUTH_TOKEN doesn't start with \"sqlc_\"")
+	}
+	c.Cloud.AuthToken = authToken
+
+	return nil
+}
```
andrewmbenton, line 17: I wanted to get away from calling `os.Getenv()` all over the place, but I'm not sure this (adding fields to `Config` struct and then populating after parse) is the best way to do it.

internal/ext/process/gen.go
```diff
@@ -37,6 +37,9 @@ func (r Runner) Generate(ctx context.Context, req *plugin.CodeGenRequest) (*plug
 		fmt.Sprintf("SQLC_VERSION=%s", req.SqlcVersion),
 	}
 	for _, key := range r.Env {
+		if key == "SQLC_AUTH_TOKEN" {
+			continue
+		}
```
andrewmbenton, line 42: We don't really need this, it just seemed like a reasonable thing to do (prevent users from sending their auth tokens to plugins).

internal/config/env.go
kyleconroy: I think as long as we only do this in one place, here in the config package, it's okay.

kyleconroy (approved):

PR #2887, opened 2023-10-19
feat: Abort compiler when rpc fails as unauthenticated

andrewmbenton:
I took the approach of adding an exported error from the `rpc` package to describe what happened, rather than adding an error like `ErrAbort` which describes an expected outcome.


internal/compiler/compile.go
```diff
@@ -88,6 +90,10 @@ func (c *Compiler) parseQueries(o opts.Parser) (*Result, error) {
 					loc = e.Location
 				}
 				merr.Add(filename, src, loc, err)
+				// If this rpc unauthenticated error bubbles up, then all future parsing/analysis will fail
+				if errors.Is(err, rpc.ErrUnauthenticated) {
+					break fileLoop
+				}
```
kyleconroy, line 96: I think we can just return early here `return nil, merr`

PR #2889, opened 2023-10-19
feat(analyzer): Cache query analysis

kyleconroy:
When using managed databases, cache the query analysis if the query, schema and configuration file hasn't change. Also take into account the version of sqlc.

Analysis can only be cached for managed databases as we can't know if a connected database has been changed.


internal/analyzer/analyzer.go
```diff
@@ -2,45 +2,95 @@ package analyzer
...
+	if err != nil {
+		return nil, err
+	}
+
+	if c.config == nil {
+		c.config, err = os.ReadFile(c.configPath)
+		if err != nil {
+			return nil, err
+		}
+	}
```
andrewmbenton, line 52: What about just serializing the `config.Config` that we've already constructed from the parsed config file bytes? You could store that on the struct instead of just `config.Database`.

That would have the advantage of creating consistent keys based on our internal config representation rather than the input. But maybe that's a disadvantage? I guess I'm thinking that if someone just adds a newline to their config that shouldn't really break the cached analysis.

internal/analyzer/analyzer.go
```diff
@@ -2,45 +2,95 @@ package analyzer
...
+	result, err := c.a.Analyze(ctx, n, q, schema, np)
+
+	if err == nil {
+		contents, err := proto.Marshal(result)
+		if err != nil {
+			return nil, err
+		}
+		if err := os.WriteFile(path, contents, 0644); err != nil {
+			return nil, err
+		}
```
andrewmbenton, line 83: I haven't thought too thoroughly about it, but it strikes me that some of these errors (both here and before the call to `c.a.Analyze()`) shouldn't cause the analysis to "fail" (especially in cases where it has in fact already succeeded). If we had a `warning` type of output I'd say this is probably a good case for it, but I don't think we do?

andrewmbenton (changes requested):

protos/analyzer/pb/pb.proto
```diff
@@ -0,0 +1,41 @@
+syntax = "proto3";
+
+package analyzer.pb;
```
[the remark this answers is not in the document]
kyleconroy: Sure

internal/analyzer/analyzer.go
kyleconroy: Yeah, that's a good idea. I wasn't trying to figure out how to serialize that struct and forgot that we serialize it to JSON / YAML already.

internal/analyzer/analyzer.go
kyleconroy: Yep, good call here.

andrewmbenton (changes requested):

internal/analyzer/analyzer.go
```diff
@@ -2,45 +2,115 @@ package analyzer
...
-	TableAlias string
-	Type       *ast.TypeName
-	EmbedTable *ast.TableName
-
-	IsSqlcSlice bool // is this sqlc.slice()
+type CachedAnalyzer struct {
+	a           Analyzer
+	config      config.Config
+	configBytes []byte
+	db          config.Database
```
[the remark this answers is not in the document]
kyleconroy: We need it because the config may have multiple managed databases in it. We need the one for this project.

andrewmbenton (approved):

PR #2894, opened 2023-10-20
feat(createdb): Create ephemeral databases

kyleconroy:
Add a new `createdb` subcommand to sqlc. This creates an ephemeral database and prints the database URL for that database.

### DIFF
--- a/internal/cmd/createdb.go
+++ b/internal/cmd/createdb.go
@@ -1,0 +1,122 @@
+package cmd
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"os"
+	"os/exec"
+	"runtime/trace"
+	"strings"
+
+	"github.com/spf13/cobra"
+	"github.com/sqlc-dev/sqlc/internal/config"
+	"github.com/sqlc-dev/sqlc/internal/opts"
+	"github.com/sqlc-dev/sqlc/internal/quickdb"
+	pb "github.com/sqlc-dev/sqlc/internal/quickdb/v1"
+	"github.com/sqlc-dev/sqlc/internal/sql/sqlpath"
+)
+
+var createDBCmd = &cobra.Command{
+	Use:   "createdb",
+	Short: "Create an ephemeral database",
+	Args:  cobra.MinimumNArgs(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		defer trace.StartRegion(cmd.Context(), "createdb").End()
+		stderr := cmd.ErrOrStderr()
+		dir, name := getConfigPath(stderr, cmd.Flag("file"))
+		env, err := cmd.Flags().GetString("env")
+		if err != nil {
+			return err
+		}
+		code, err := CreateDB(cmd.Context(), dir, name, args, env, &Options{
+			Env:    ParseEnv(cmd),
+			Stderr: stderr,
+		})
+		if err != nil {
+			fmt.Fprintln(stderr, err.Error())
+			os.Exit(code)
+		}
+		return nil
+	},
+}
+
+func CreateDB(ctx context.Context, dir, filename string, args []string, env string, o *Options) (int, error) {
+	dbg := opts.DebugFromEnv()
+	if !dbg.ProcessPlugins {
+		return 1, fmt.Errorf("process-plugins disabled")
+	}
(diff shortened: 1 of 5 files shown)


internal/cmd/createdb.go
```diff
@@ -0,0 +1,122 @@
...
+
+var createDBCmd = &cobra.Command{
+	Use:   "createdb",
+	Short: "Create an ephemeral database",
+	Args:  cobra.MinimumNArgs(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		defer trace.StartRegion(cmd.Context(), "createdb").End()
+		stderr := cmd.ErrOrStderr()
+		dir, name := getConfigPath(stderr, cmd.Flag("file"))
+		env, err := cmd.Flags().GetString("env")
```
andrewmbenton, line 28: I would name this flag something like `env-var` and then name the variable `envVar` to avoid confusion with our other `Env` command context

internal/cmd/createdb.go
```diff
@@ -0,0 +1,122 @@
...
+func CreateDB(ctx context.Context, dir, filename string, args []string, env string, o *Options) (int, error) {
+	dbg := opts.DebugFromEnv()
+	if !dbg.ProcessPlugins {
+		return 1, fmt.Errorf("process-plugins disabled")
+	}
+	_, conf, err := o.ReadConfig(dir, filename)
+	if err != nil {
+		return 1, err
+	}
+	// Find the first SQL with a managed database
```
andrewmbenton, line 53: I guess we don't have any obvious name we can use as an identifier, but we should probably introduce one.

Or use some combination of `engine` and `out` or whatever required parameters we have in the config?

internal/cmd/createdb.go
```diff
@@ -0,0 +1,122 @@
...
+		Engine:     "postgresql",
+		Region:     quickdb.GetClosestRegion(),
+		Migrations: migrations,
+	})
+	if err != nil {
+		return 1, fmt.Errorf("managed: create database: %w", err)
+	}
+
+	defer func() {
+		client.DropEphemeralDatabase(ctx, &pb.DropEphemeralDatabaseRequest{
```
andrewmbenton, line 96: I think you can just defer this directly, without wrapping in an anonymous `func()`

internal/cmd/createdb.go
```diff
@@ -0,0 +1,122 @@
...
+			DatabaseId: resp.DatabaseId,
+		})
+	}()
+
+	cmd := exec.Command(args[0], args[1:]...)
+	cmd.Env = append(os.Environ(), fmt.Sprintf("%s=%s", env, resp.Uri))
```
andrewmbenton, line 102: You overwrite this later

andrewmbenton (changes requested):

internal/cmd/createdb.go
```diff
@@ -0,0 +1,101 @@
...
+		})
+		if err != nil {
+			fmt.Fprintln(stderr, err.Error())
+			os.Exit(1)
+		}
+		return nil
+	},
+}
+
+func CreateDB(ctx context.Context, dir, filename, name string, o *Options) error {
```
andrewmbenton, line 41: `name` -> `qs`

internal/cmd/createdb.go
```diff
@@ -0,0 +1,101 @@
...
+
+	resp, err := client.CreateEphemeralDatabase(ctx, &pb.CreateEphemeralDatabaseRequest{
+		Engine:     "postgresql",
+		Region:     quickdb.GetClosestRegion(),
+		Migrations: ddl,
+	})
+	if err != nil {
+		return fmt.Errorf("managed: create database: %w", err)
+	}
+	fmt.Fprintln(os.Stderr, "WARNING: This database will be removed in two minutes")
```
andrewmbenton, line 98: I'm pretty sure `log` prints to `stderr`

internal/config/config.go
```diff
@@ -103,6 +103,7 @@ type GenGo struct {
 }
 
 type SQL struct {
+	Name                 string    `json:"name" yaml:"name"`
```
andrewmbenton, line 106: gotta update that `v_two.json`

andrewmbenton (changes requested):

andrewmbenton (approved):

PR #2898, opened 2023-10-20
feat(debug): Add databases=managed debug option

kyleconroy:
- Remove the --no-database flag
- Refactor the shfmt package

### DIFF
--- a/internal/cmd/vet.go
+++ b/internal/cmd/vet.go
@@ -482,7 +482,7 @@ func (c *checker) checkSQL(ctx context.Context, s config.SQL) error {
 	var expl explainer
 	if s.Database != nil { // TODO only set up a database connection if a rule evaluation requires it
 		if s.Database.URI != "" && c.OnlyManagedDB {
-			return fmt.Errorf("database: connections disabled via command line flag")
+			return fmt.Errorf("database: connections disabled via SQLCDEBUG=databases=managed")
 		}
 		dburl, cleanup, err := c.fetchDatabaseUri(ctx, s)
 		if err != nil {
--- a/internal/cmd/vet.go
+++ b/internal/cmd/vet.go
@@ -144,13 +144,13 @@ func Vet(ctx context.Context, dir, filename string, opts *Options) error {
 	}
 
 	c := checker{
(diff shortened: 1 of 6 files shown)


internal/engine/postgresql/analyzer/analyze.go
```diff
@@ -204,8 +209,10 @@ func (a *Analyzer) Analyze(ctx context.Context, n ast.Node, query string, migrat
 				return nil, err
 			}
 			uri = edb.Uri
+		} else if a.dbg.OnlyManagedDatabases {
+			return nil, fmt.Errorf("database: connections disabled via command line flag")
```
andrewmbenton, line 213:
```suggestion
			return nil, fmt.Errorf("database: connections disabled via SQLCDEBUG=databases=managed")
```

internal/cmd/vet.go
```diff
@@ -488,7 +481,7 @@ func (c *checker) checkSQL(ctx context.Context, s config.SQL) error {
 	var prep preparer
 	var expl explainer
 	if s.Database != nil { // TODO only set up a database connection if a rule evaluation requires it
-		if c.NoDatabase {
+		if s.Database.URI != "" && c.OnlyManagedDB {
 			return fmt.Errorf("database: connections disabled via command line flag")
```
andrewmbenton, line 485:
```suggestion
			return fmt.Errorf("database: connections disabled via SQLCDEBUG=databases=managed")
```

internal/cmd/vet.go
```diff
@@ -448,14 +448,7 @@ func (c *checker) fetchDatabaseUri(ctx context.Context, s config.SQL) (string, f
 }
 
 func (c *checker) DSN(dsn string) (string, error) {
```
andrewmbenton, line 450: Don't really need this func anymore

andrewmbenton (changes requested):

internal/cmd/vet.go
kyleconroy: Agreed, but it was easier to keep it around

andrewmbenton (approved):

PR #2918, opened 2023-10-26
refactor: Remove `Overrides` from `Settings` in codegen.proto, pass in JSON opts

andrewmbenton:
The `Overrides` configuration only applies to Go, so we remove it from the codegen plugin proto and instead pass the configuration as JSON to be unmarshaled by the codegen/golang package like other options. Also pull codegen/golang opts into its own package since its about to get a lot more complex.

Eventually we will need to push all of the validation of overrides from internal/config into the codegen/golang package, but that's a fair bit of work so I didn't push it here.

A lot of the changes in this diff are just from pushing the `opts` type within codegen/golang into a package, which is necessary eventually but pretty distracting right now so sorry about that.

Aside from the proto change, which is obviously important, the other meaningful change is to generate.go on line 421, and the function implementation in shim.go. The types in opts/override.go are just recreations of what the plugin proto types were.


internal/codegen/golang/opts/options.go
```diff
@@ -39,9 +39,10 @@ type Options struct {
 	OmitUnusedStructs           bool     `json:"omit_unused_structs,omitempty"`
 	BuildTags                   string   `json:"build_tags,omitempty"`
 
-	// Unused but included in marshaled json we receive
-	Overrides json.RawMessage `json:"overrides,omitempty"`
-	Rename    json.RawMessage `json:"rename,omitempty"`
+	QuerySetOverrides []Override      `json:"overrides,omitempty"`
+	QuerySetRename    json.RawMessage `json:"rename,omitempty"` // Unused, TODO merge with req.Settings.Rename
+
+	Overrides []GoOverride `json:"-"`
```
andrewmbenton, line 45: We consolidate all overrides here and only use overrides parsed from JSON options and overrides coming in from `req.Settings.Overrides` within `ParseOpts()` below to build this slice.

internal/codegen/golang/imports.go
```diff
@@ -59,11 +58,10 @@ func mergeImports(imps ...fileImports) [][]ImportSpec {
 }
 
 type importer struct {
-	Settings *plugin.Settings
```
andrewmbenton, line 62: having all overrides in `Options` lets us drop this field

internal/cmd/shim.go
```diff
@@ -101,12 +108,12 @@ func pluginWASM(p config.Plugin) *plugin.Codegen_WASM {
 	return nil
 }
 
-func pluginGoType(o config.Override) *plugin.ParsedGoType {
+func pluginGoType(o config.Override) *goopts.ParsedGoType {
```
andrewmbenton, line 111: It's obviously not great to use the plugin-defined type here, since in a world where there's actual separation between sqlc and golang codegen we wouldn't be able to import it. But it seemed like the best choice for now.

Although maybe we can delegate all parsing of the go type to the plugin and not do any at sqlc config parse time?

internal/endtoend/testdata/process_plugin_disabled/gen/codegen.json
```diff
@@ -14,41 +14,6 @@
...
-      "emit_pointers_for_null_types": false,
-      "query_parameter_limit": 1,
-      "output_batch_file_name": "",
-      "json_tags_id_uppercase": false,
-      "omit_unused_structs": false
-    },
-    "json": {
-      "out": "",
-      "indent": "",
-      "filename": ""
```
andrewmbenton, line 51: This is cleanup from previous change in https://github.com/sqlc-dev/sqlc/pull/2822

internal/endtoend/testdata/overrides_go_struct_tags/invalid_tags/stderr.txt
```diff
@@ -1 +1,2 @@
-error parsing sqlc.json: bad syntax for struct tag pair
+# package override
+error generating code: bad syntax for struct tag pair
```
andrewmbenton, line 2: Because struct tags get parsed in the codegen plugin now

internal/codegen/golang/opts/options.go
```diff
@@ -52,6 +53,30 @@ func ParseOpts(req *plugin.CodeGenRequest) (*Options, error) {
...
+	for _, override := range req.Settings.Overrides {
+		var goType ParsedGoType
+		if err := json.Unmarshal(override.CodeType, &goType); err != nil {
+			return options, err
+		}
+		options.Overrides = append(options.Overrides, GoOverride{
+			override,
+			&goType,
+		})
+	}
```
kyleconroy: This feels backwards to me. We shouldn't need to serialize the query set overrides back to JSON only to parse them again a few lines later. Can't we iterate over req.Settings.Overrides and then iterate over the QuerySetOverrides?

andrewmbenton: Yeah we can, I think this is a remnant of a different approach I abandoned. I will re-work it

internal/cmd/shim.go
andrewmbenton: Ended up dropping this altogether after removing parse logic for go type in sqlc config

PR #2927, opened 2023-10-30
fix(engine/sqlite): support -> and ->> operators

### LINKED ISSUE #2919: sqlite/json: support -> and ->> operators

> Beginning with SQLite version 3.38.0 (2022-02-22), the -> and ->> operators are available for extracting subcomponents of JSON.[^1]

[^1]: 4.6. The -> and ->> operators
https://www.sqlite.org/json1.html#jptr

seems the latest sqlc v1.23 release does not support it:

https://play.sqlc.dev/p/25c82bb68f7a141175f4e51fd58047e1dc1466f61e02d059f5dcd6edf06a04d3

SQLite

Go

orisano:
close #2919

There are still cases where the type inference part does not work properly, but this can be worked around with explicit CAST.

The E2E test included in this PR also yields incorrect type inference results.
The type of the result of the ->> operator is any, but it is inferred to be text.


internal/compiler/output_columns.go
```diff
@@ -158,12 +158,14 @@ func (c *Compiler) outputColumns(qc *QueryCatalog, node ast.Node) ([]*Column, er
...
-			case lang.IsComparisonOperator(astutils.Join(n.Name, "")):
+			switch op := astutils.Join(n.Name, ""); {
+			case lang.IsComparisonOperator(op):
 				// TODO: Generate a name for these operations
 				cols = append(cols, &Column{Name: name, DataType: "bool", NotNull: true})
-			case lang.IsMathematicalOperator(astutils.Join(n.Name, "")):
+			case lang.IsMathematicalOperator(op):
 				cols = append(cols, &Column{Name: name, DataType: "int", NotNull: true})
+			case op == "||" || op == "->":
+				cols = append(cols, &Column{Name: name, DataType: "text"})
```
kyleconroy, line 168: Can we just remove this line so that we get "any" type back?

orisano: removed

kyleconroy (approved):

PR #2929, opened 2023-10-30
feat(codegen): Remove Go-specific overrides from codegen proto

kyleconroy:
As part of our long-term plans for additional languages, we're cleaning up the interface between sqlc and codegen plugins. This change removes the rename and overrides struct, as it had Go-specific information. In it's place is a plugin-agnostic global options block. The current Go-specific configuration looks like this:

```yaml
version: "2"
overrides:
  go:
    rename:
      one: "ONE"
    overrides:
      - db_type: "text"
        go_type: "github.com/sqlc-dev/sqlc-testdata/pkg.CustomType"
      - column: "overrides.three"
        go_type:
          import: "github.com/sqlc-dev/sqlc-testdata/pkg"
          type: "CustomType"
sql:
- schema: "schema.sql"
  queries: "query.sql"
  engine: "postgresql"
  gen:
    go: 
      out: "db"
      rename:
        one: "ONE"
      overrides:
        - column: "overrides.three"
          go_type:
            import: "invalid/import"
            type: "ShouldNotSeeThis"
```

A two line change to use this with a theoretical Go plugin

```yaml
version: "2"
plugins:
- name: go
  wasm:
    url: https://downloads.sqlc.dev/plugin/sqlc-gen-go_X.X.X.wasm
    sha256: ""
options:
  go:
    rename:
      one: "ONE"
    overrides:
      - db_type: "text"
        go_type: "github.com/sqlc-dev/sqlc-testdata/pkg.CustomType"
      - column: "overrides.three"
        go_type:
          import: "github.com/sqlc-dev/sqlc-testdata/pkg"
          type: "CustomType"
sql:
- schema: "schema.sql"
  queries: "query.sql"
  engine: "postgresql"
  codegen:
  - out: "db"
    plugin: go 
    options:
      rename:
        one: "ONE"
      overrides:
        - column: "overrides.three"
          go_type:
            import: "invalid/import"
            type: "ShouldNotSeeThis"
```

Supersedes https://github.com/sqlc-dev/sqlc/pull/2918


protos/plugin/codegen.proto
```diff
@@ -156,6 +122,7 @@ message CodeGenRequest {
   repeated Query queries = 3 [json_name = "queries"];
   string sqlc_version = 4 [json_name = "sqlc_version"];
   bytes plugin_options = 5 [json_name = "plugin_options"];
```
andrewmbenton, line 124: sadly this should probably be `queryset_options` but not worth changing.

internal/codegen/golang/opts/override.go
```diff
@@ -1,13 +1,79 @@
...
+	return &PluginGoType{
+		ImportPath: o.GoImportPath,
+		Package:    o.GoPackage,
+		TypeName:   o.GoTypeName,
+		BasicType:  o.GoBasicType,
+		StructTags: o.GoStructTags,
+	}
+}
+
+type PluginOverride struct {
```
andrewmbenton, line 34: I would rename this and the `PluginGoType` based on whatever you think a good name for the `Override` field is.

internal/config/config.go
```diff
@@ -241,12 +205,11 @@ func ParseConfig(rd io.Reader) (Config, error) {
...
-	Package   SQL
-	Go        SQLGo
-	JSON      SQLJSON
-	Rename    map[string]string
-	Overrides []Override
+	Global  Config
+	Package SQL
+	Go      golang.Options
+	JSON    SQLJSON
+	Rename  map[string]string
```
andrewmbenton, line 212: I don't think this Rename map is used anymore? Because in `codegen()` we pull everything from `combo.Global.Overrides.Go`.

internal/config/v_two.go
```diff
@@ -84,11 +77,6 @@ func v2ParseConfig(rd io.Reader) (Config, error) {
...
 			}
 
-			for i := range conf.SQL[j].Gen.Go.Overrides {
-				if err := conf.SQL[j].Gen.Go.Overrides[i].Parse(); err != nil {
-					return conf, err
-				}
-			}
 			for k, v := range conf.SQL[j].Gen.Go.Rename {
 				conf.SQL[j].Gen.Go.Rename[k] = v
 			}
```
andrewmbenton, line 82: Not related to this PR, but am I correct in reading that this loop is just an expensive noop?

andrewmbenton (changes requested):

internal/codegen/golang/opts/options.go
```diff
@@ -4,60 +4,110 @@ import (
...
 	dec := json.NewDecoder(bytes.NewReader(req.PluginOptions))
-	dec.DisallowUnknownFields()
 	if err := dec.Decode(&options); err != nil {
-		return options, fmt.Errorf("unmarshalling options: %w", err)
+		return &options, fmt.Errorf("unmarshalling options: %w", err)
+	}
+
+	for i := range options.Overrides {
+		if err := options.Overrides[i].parse(req); err != nil {
+			return nil, err
```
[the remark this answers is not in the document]
kyleconroy: We should return nil here and not a valid options struct

internal/codegen/golang/opts/override.go
kyleconroy: I'm going to move both of these into shim.go

internal/cmd/generate.go
```diff
@@ -411,10 +411,19 @@ func codegen(ctx context.Context, combo config.CombinedSettings, sql outPair, re
...
-			return "", nil, fmt.Errorf("invalid plugin options")
+			return "", nil, fmt.Errorf("invalid plugin options: %w", err)
 		}
 		req.PluginOptions = opts
 
+		global, found := combo.Global.Options[plug.Name]
+		if found {
+			opts, err := convert.YAMLtoJSON(global)
+			if err != nil {
+				return "", nil, fmt.Errorf("invalid global options; %w", err)
```
andrewmbenton, line 422:
```suggestion
				return "", nil, fmt.Errorf("invalid global options: %w", err)
```

andrewmbenton (approved):

PR #2930, opened 2023-10-31
feat(plugin): Use gRPC interface for codegen plugin communication

andrewmbenton:
I propose to introduce a `service` block at the top of codegen.proto which currently defines a single rpc named `Generate` which takes and returns the existing codegen proto messages. Establishing a way for codegen plugins to interact with sqlc in an extensible way is the goal.

I use the generated grpc client stub as a proof-of-concept, implementing a basic client conn interface for each of our three plugin types (built-in, process and wasm). The implementation doesn't yet expect a plugin to handle anything other than the existing codegen call, though I do pass the `method` parameter as an argument to process and wasm plugins with the expectation that we will update our plugin-building sdk to use generated server stubs and implement a handler "router" of some kind.


kyleconroy (approved):
Remove the vtproto code

PR #2955, opened 2023-11-07
Enhance `sqlpath.Glob` to actually support glob wildcard

iamwavecut:
This PR enhances the `Glob` function in the `internal/sqlpath` package to support glob patterns in the input paths. It also adds comprehensive tests for the `Glob` function, ensuring its correct behavior and in various scenarios, as well as protects it from regressions.

Problem
====
I want `sqlc` to be able to collect queries from different modules of my applications:
```sqlc.yaml
...
sql:
  - engine: "postgresql"
    queries: "modules/*/sql"
...
```
Epected:
====
Normal execution.

Actual:
====
```
error parsing queries: path /home/masked/project/internal/modules/*/sql does not exist
```

- `internal/sql/sqlpath/read_test.go`: Introduced a new test file with multiple test cases for the Glob function. These tests cover various scenarios like handling of SQL files, non-SQL files, hidden files, non-existent paths, unreadable directories, and symlinks. 

- `internal/sql/sqlpath/read.go`: Added a new function `expandGlob` s to expand glob patterns in the input paths. The `Glob` function now calls `expandGlob` s before processing the paths. Also, added error handling for cases where the path does not exist or is not a file. 

- `internal/sql/sqlpath/testdata/`: Dummy directory of files and symlinks to test various scenarios.

### DIFF
--- a/internal/sql/sqlpath/read.go
+++ b/internal/sql/sqlpath/read.go
@@ -49,7 +49,7 @@ func Glob(patterns []string) ([]string, error) {
 			files = append(files, path)
 		}
 	}
-	var sqlFiles []string //nolint:prealloc // can be empty
+	var sqlFiles []string
 	for _, file := range files {
 		if !strings.HasSuffix(file, ".sql") {
 			continue


iamwavecut:
@sqlc-dev please take a look

kyleconroy:
@iamwavecut I made a few changes to the PR. I remove the nolint comment, as we don't use those anywhere in the codebase. I only call filepath.Glob if the path contains `?*[]`. It's easier to preserve existing behavior this way.

I also changed the error messages to include the info from os.Stat. This information isn't as useful as the results from lstat, but it is more consistent across platforms.

iamwavecut:
@kyleconroy Sure! Thanks for the improvements.

PR #2962, opened 2023-11-09
fix: Return an error instead of generating duplicate Go names

andrewmbenton:
Resolves https://github.com/sqlc-dev/sqlc/issues/588

### DIFF
--- a/internal/codegen/golang/gen.go
+++ b/internal/codegen/golang/gen.go
@@ -132,22 +132,26 @@ func Generate(ctx context.Context, req *plugin.GenerateRequest) (*plugin.Generat
 }
 
 func validate(options *opts.Options, enums []Enum, structs []Struct, queries []Query) error {
-	usedNames := make(map[string]struct{})
+	enumNames := make(map[string]struct{})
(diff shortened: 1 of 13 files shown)


internal/codegen/golang/gen.go
```diff
@@ -124,9 +124,36 @@ func Generate(ctx context.Context, req *plugin.GenerateRequest) (*plugin.Generat
...
+	if !options.EmitExportedQueries {
+		return nil
+	}
+	for _, query := range queries {
+		if _, ok := usedNames[query.ConstantName]; ok {
+			return fmt.Errorf("query constant name conflicts with struct name: %s", query.ConstantName)
+		}
+	}
+	return nil
+}
```
kyleconroy, line 155: Let's keep two separate maps for used names for providing better error messages.

```
struct name conflicts with enum name: %s

query constant name conflicts with struct name: %s

query constant name conflicts with enum name: %s
```

kyleconroy (changes requested):
A few suggestions on error messages.

kyleconroy (approved):

PR #2966, opened 2023-11-11
Set the default schema name from configs.

debugger84:
The codebase already has the "Default schema" term. But unfortunately, it is hardcoded in the codebase as "public" for the PostgreSQL. It cannot be changed from configs. And this is the problem in such a situation: 

I have several services with separate schemas in a database. For example, the "auth microservice" has its own schema "auth" in the DB. All parts of this microservice work only with this schema and no one else. In terms of codebase, I have entities like AccessToken, Account, etc.  But SQLc generates me structs like AuthAccessToken, AuthAccount, etc. This Auth prefix is redundant and I don't have the ability to switch off it.

This request solved this issue it added the changing of the default schema, which asked SQLc to generate Entities without the schema name as a prefix.


debugger84:
If someone is interested in this topic. 
I've solved this issue in another way. I've created a fork from the sqlc-go-gen plugin and added this functionality there. 

Use it https://github.com/debugger84/sqlc-gen-go instead of these changes

kyleconroy:
**🤖 Automated message**

---

PR #2976, opened 2023-11-14
fix(codegen): Never generate return structs for :exec

kyleconroy:
Also validate `:batchone` commands to make sure they return values. Currently this causes a template error as seen here: https://play.sqlc.dev/p/bf5cff5c903377ae7d92efaeb063c8446b93fef7dac67188eb233f5f65095d41

Fixes https://github.com/sqlc-dev/sqlc/issues/2970


DavidS-om (approved):
Thanks for the quick fix! I've verified that this works for my use-case.

PR #2982, opened 2023-11-15
feat(examples): Use hosted MySQL databases for tests

kyleconroy:


examples/authors/sqlc.yaml
```diff
@@ -20,8 +20,9 @@ sql:
 - schema: mysql/schema.sql
   queries: mysql/query.sql
   engine: mysql
+  name: mysql
```
andrewmbenton, line 23: Set name on postgres queryset too?

andrewmbenton (approved):

internal/cmd/vet.go
```diff
@@ -404,10 +404,6 @@ func (c *checker) fetchDatabaseUri(ctx context.Context, s config.SQL) (string, f
 		return uri, cleanup, err
 	}
 
-	if s.Engine != config.EnginePostgreSQL {
-		return "", cleanup, fmt.Errorf("managed: only PostgreSQL currently")
-	}
```
andrewmbenton, line 409: We might still want to bail early if the engine is sqlite?

kyleconroy: I wanted to let the client make the request and fail so that it's easier to add engine support at the server without a client change.

PR #2986, opened 2023-11-17
feat(verify): Add new command to verify queries and migrations

kyleconroy:
### push

We've renamed the upload sub-command to push. Using `upload` is supported for now, but please switch over to using `push`. We changed the data sent along in a push request. Upload used to include the configuration file, migrations, queries, and all generated code. Push drops the generated code in favor of including the [plugin.GenerateRequest], which is the protobuf message we pass to codegen plugins.

We now pass along annotations for each push. By default, we include these environment variables if they are present:

```
GITHUB_REPOSITORY
GITHUB_REF
GITHUB_REF_NAME
GITHUB_REF_TYPE
GITHUB_SHA
```

Like upload, push should be run when you tag a release of your application. We run it on every push to main, as we continuously deploy those commits.

### verify

The verify command, building on top of the push command,  ensures migrations are safe to deploy. Verify sends your current schema and queries to sqlc cloud. There, we run the queries for your latest push against your new schema changes. This check catches backwards incompatible schema changes for existing queries.


internal/bundler/upload.go
```diff
@@ -58,32 +67,79 @@ func (up *Uploader) Validate() error {
...
+
+func annotate() map[string]string {
+	labels := map[string]string{}
+	for _, ev := range envvars {
+		key := strings.ReplaceAll(strings.ToLower(ev), "_", ".")
+		labels[key] = os.Getenv(ev)
 	}
-	outs, err := readOutputs(up.dir, result)
+	return labels
+}
```
andrewmbenton, line 85: I guess it's frowned upon these days but this seems like a case for `init()`

internal/cmd/push.go
```diff
@@ -0,0 +1,73 @@
...
+
+	"github.com/sqlc-dev/sqlc/internal/bundler"
+	"github.com/sqlc-dev/sqlc/internal/compiler"
+	"github.com/sqlc-dev/sqlc/internal/config"
+)
+
+var pushCmd = &cobra.Command{
+	Use:     "push",
+	Aliases: []string{"upload"},
+	Short:   "Push the schema, queries, and configuration to your project",
```
andrewmbenton, line 19: This used to be "...for this project" which is maybe not ideal but I think is clearer than "...to your project"

internal/cmd/push.go
```diff
@@ -0,0 +1,73 @@
...
+	if err != nil {
+		return err
+	}
+	up := bundler.NewUploader(configPath, dir, conf)
+	if err := up.Validate(); err != nil {
+		return err
+	}
+	p := &pusher{}
+	if err := Process(ctx, p, dir, filename, opts); err != nil {
+		os.Exit(1)
```
andrewmbenton, line 66: Should just be `return err` no?

internal/cmd/verify.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/spf13/cobra"
+
+	"github.com/sqlc-dev/sqlc/internal/bundler"
+	"github.com/sqlc-dev/sqlc/internal/quickdb"
+	quickdbv1 "github.com/sqlc-dev/sqlc/internal/quickdb/v1"
+)
+
+var verifyCmd = &cobra.Command{
+	Use:   "verify",
+	Short: "Verify schema, queries, and configuration against main",
```
andrewmbenton, line 17: I don't think "against main" means anything here. I don't have a better idea at the moment, but I think just saying `"Verify schema, queries, and configuration for this project"` would be fine for now.

internal/cmd/verify.go
```diff
@@ -0,0 +1,63 @@
...
+		return err
+	}
+	client, err := quickdb.NewClientFromConfig(conf.Cloud)
+	if err != nil {
+		return fmt.Errorf("client init failed: %w", err)
+	}
+	p := &pusher{}
+	if err := Process(ctx, p, dir, filename, opts); err != nil {
+		// hmmm
+		os.Exit(1)
```
andrewmbenton, line 46: Is there a reason we can't just return err here?

andrewmbenton (changes requested):

PR #3032, opened 2023-12-04
fix: Use derived ArrayDims instead of deprecated attndims

hendrikhofstadt:
Postgres sometimes returns incorrect values for attndims which explains the TODO in the code. Use the type-derived dimensions instead.
This fixes a bug where in rare cases arrays were not correctly detected in one of our applications.

According to https://postgrespro.com/list/thread-id/2304332
> We don't really care.  attndims is a historical leftover and should never
be relied on for any semantically significant purpose.
TBH, my inclination would be to remove the column altogether rather than
"fix" this.


kyleconroy:
Thanks!

PR #3042, opened 2023-12-05
feat(plugins): Use wazero instead of wasmtime

kyleconroy:
After merging #3027 and #3040, wasmtime is our only cgo dependency left. This change replaces with wazero, a pure-Go WASM runtime. I've played around with it locally and everything appears to work, but I'm going to need to understand how the caching works before this is merged.


internal/ext/wasm/wasm.go
```diff
@@ -245,72 +195,56 @@ func (r *Runner) Invoke(ctx context.Context, method string, args any, reply any,
...
+	rt := wazero.NewRuntimeWithConfig(ctx, config)
+	defer rt.Close(ctx)
 
-	// Configure WASI imports to write stdout into a file.
-	wasiConfig := wasmtime.NewWasiConfig()
-	wasiConfig.SetArgv([]string{"plugin.wasm", method})
-	wasiConfig.SetStdinFile(stdinPath)
-	wasiConfig.SetStdoutFile(stdoutPath)
-	wasiConfig.SetStderrFile(stderrPath)
+	// TODO: Handle error
```
anuraaga, line 217: There is `Instantiate` if you'd like to return the error. Though I think any failure here would be a programming bug, not non-determinstic

internal/ext/wasm/wasm.go
```diff
@@ -245,72 +195,56 @@ func (r *Runner) Invoke(ctx context.Context, method string, args any, reply any,
...
+	wasi_snapshot_preview1.MustInstantiate(ctx, rt)
 
-	keys := []string{"SQLC_VERSION"}
-	vals := []string{info.Version}
-	for _, key := range r.Env {
-		keys = append(keys, key)
-		vals = append(vals, os.Getenv(key))
+	// Compile the Wasm binary once so that we can skip the entire compilation time during instantiation.
+	mod, err := rt.CompileModule(ctx, wasmBytes)
```
anuraaga, line 221: If it's possible, it would be nice to rejigger to scope this to `Runner`, possibly with some `map[/* wasm url */ string]wazero.CompiledModule`. The compilation cache is good to reuse across executions of the sqlc process itself, but it's also good to only compile once per wasm within a process if possible, the cache key computation isn't trivial. Though if the latter doesn't happen that much maybe it doesn't matter

internal/ext/wasm/wasm.go
```diff
@@ -245,72 +195,56 @@ func (r *Runner) Invoke(ctx context.Context, method string, args any, reply any,
...
+	if cerr := checkError(err, &stderr); cerr != nil {
 		return cerr
 	}
 
 	// Print WASM stdout
-	stdoutBlob, err := os.ReadFile(stdoutPath)
+	stdoutBlob, err := io.ReadAll(&stdout)
 	if err != nil {
 		return fmt.Errorf("read file: %w", err)
 	}
```
anuraaga, line 250:
```suggestion
	stdoutBlob := stdout.Bytes()
```

internal/ext/wasm/wasm.go
```diff
@@ -331,21 +265,19 @@ func (r *Runner) NewStream(ctx context.Context, desc *grpc.StreamDesc, method st
...
+			return nil
 		}
 	}
+
 	// Print WASM stdout
-	stderrBlob, rferr := os.ReadFile(stderrPath)
+	stderrBlob, rferr := io.ReadAll(stderr)
 	if rferr == nil && len(stderrBlob) > 0 {
 		return errors.New(string(stderrBlob))
 	}
```
anuraaga, line 283:
```suggestion
	stderrBlob := stderr.String()
	if len(stderrBlob) > 0 {
		return errors.New(stderrBlob)
	}
```

anuraaga (commented):
I was thinking of sending this, great to see it :D

anuraaga:
Oops sorry looks like need to update the type argument to `checkError` method

anuraaga:
Hi @kyleconroy - sorry if this steps on any toes and feel free to ignore it, being interested in this PR I was hoping to help continue it. I sent #3082 as a PR to this branch with the suggestions. Thanks.

PR #3045, opened 2023-12-07
Use wazero based SQLite bindings.

ncruces:
Since you're using wazero to avoid cgo in #3027 and #3042 I though you might want to consider using my wazero based SQLite bindings: [`github.com/ncruces/go-sqlite3`](https://github.com/ncruces/go-sqlite3).

They're no where near as battle tested as #3040, so feel free to close this.


kyleconroy:
@ncruces Appreciate the pull request. I actually switched from mattn/go-sqlite3 to modernc.org/sqlite a few days in https://github.com/sqlc-dev/sqlc/pull/3040. That removes our cgo dependency for now.

ncruces:
Sure, makes sense.
If you ever feel like it's worth it to consolidate dependencies, now you know it exists.

PR #3048, opened 2023-12-08
build(sqlite): Disable modernc for WASM

kyleconroy:
I've been experimenting with compiling the sqlc CLI to WASI. To get this to work, the `modernc.org/sqlite` package needs to be excluded from the build, as it doesn't support the WASI target.

```
CGO_ENABLED=0 GOOS=wasip1 GOARCH=wasm go build -o sqlc.wasm ./cmd/sqlc
```


ncruces:
Just FYI, #3045 would take care of this [^1] (with a caveat: WASI doesn't support file locking, so concurrent access is problematic).

[^1]: as well as other exotic targets, like DragonFlyBSD, illumos, Solaris…

PR #3060, opened 2023-12-14
override match column nullable

trevatk:
MR to fix override types not being applied. 

Issue: https://github.com/sqlc-dev/sqlc/issues/3005


kyleconroy:
@trevatk I believe that the current code is correct, based on the test failures. Closing this for now.

PR #3069, opened 2023-12-19
feat(analyzer): Return zero values when encountering unexpected ast nodes

andrewmbenton:
This allows users to run generate on any parsable query, even when sqlc doesn't understand it.

Closes #2377


andrewmbenton:
I think we need to think this through a little more. The vet_failures test adjustment hints at the fact that we are now adding queries to the compiler result (and codegenrequest) that we weren't previously, because they were "unsupported". We are doing this even though the queries don't have any annotation, which is perhaps unexpected.

andrewmbenton:
> What do you think about ignoring queries that don't have an annotation? I think that feels like the expected behavior.

Yeah I had implemented that initially, and I think it's correct. It will change a decent amount of test output though so I would do it in a separate PR.

I could also argue that this isn't that big of a change, since we must have been adding un-annotated queries to codegen requests all along, but only if they were "supported."

kyleconroy (approved):

PR #3073, opened 2023-12-20
fix(codegen/golang): Don't import `"strings"` for `sqlc.slice()` with pgx

andrewmbenton:
Also fix the invalid test that should have caught this.

Resolves https://github.com/sqlc-dev/sqlc/issues/3065

### DIFF
--- a/internal/codegen/golang/imports.go
+++ b/internal/codegen/golang/imports.go
@@ -390,7 +390,7 @@ func (i *importer) queryImports(filename string) fileImports {
 	}
 
 	sqlpkg := parseDriver(i.Options.SqlPackage)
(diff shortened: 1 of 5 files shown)


internal/codegen/golang/imports.go
```diff
@@ -390,7 +390,7 @@ func (i *importer) queryImports(filename string) fileImports {
 	}
 
 	sqlpkg := parseDriver(i.Options.SqlPackage)
-	if sqlcSliceScan() {
+	if sqlcSliceScan() && !sqlpkg.IsPGX() {
```
kyleconroy, line 393: :( so gross

kyleconroy (approved):

PR #3074, opened 2023-12-20
feat(push): Add tag support

kyleconroy:
When pushing queries to sqlc Cloud, you can now specify tags in the push command.

```
sqlc push --tag one --tag two
```

These tags operate similar to Docker and Git tags. They can be overwritten and will always point to a single push.

The verify command gets the new `--against` flag. If supplied, we'll use the tagged push to look up queries to verify against the current schema.

```
sqlc verify --against one
```


internal/cmd/options.go
```diff
@@ -10,6 +10,9 @@ type Options struct {
 	Env          Env
 	Stderr       io.Writer
 	MutateConfig func(*config.Config)
+	// TODO: Move these to a command-specific struct
+	Tags    []string
+	Against string
```
kyleconroy, line 15: This is gross, right now we expect the syntax of the `Verify` and `Push` command to be the same. Will clean this up in another pull request.

andrewmbenton (approved):

PR #3093, opened 2024-01-02
feat(sqlc): support  emit_pointers_for_null_types for enums

godwhoa:


andrewmbenton:
I think this might not be enough. Do we not also need to modify the scanning code for the non-nullable enum types so that they properly return a nil pointer when they scan a `NULL`? Or does this somehow "just work"?

Have you tried manually modifying the sqlc output to use pointers for nullable enums and confirmed that this proposed change alone does actually work as expected?

godwhoa:
@andrewmbenton I did try with a basic `INSERT RETURNING` locally which worked, but didn't try any more elaborate cases that you mention.
Will try to pick it up over the week.

kyleconroy:
**🤖 Automated message**

---

PR #3105, opened 2024-01-04
fix(scripts): Remove test output regeneration script

andrewmbenton:
This script no longer works, since test output now depends on a "context" and a significant amount of output relies on using a sqlc cloud managed database. So we should just remove the script.

Resolves https://github.com/sqlc-dev/sqlc/issues/3097


kyleconroy (approved):

Jille:
What's the expected workflow for testing changes? How do I verify a codegen change doesn't affect more than I expect?

andrewmbenton:
The existing tests will still fail when run if you make a change that modifies their output.

The regen script was mostly used to bump the sqlc version string in code comments in test output. We have another way to do that now though.

To create expected output for a new test, you can `make sqlc-dev` and then manually run `sqlc-dev generate` in the test directory.

PR #3125, opened 2024-01-11
feat: add "use_interface_return_types" option

### LINKED ISSUE #383: WithTx is not part of the method set of interface generated

Jesse0Michael:
- Add the option "**use_interface_return_types**" to direct sqlc generation to use the querier interface for the return type of the WithTx method on the Queries struct.

- Include `WithTx` in the Querier interface (when "**emit_methods_with_db_argument**" is false)

---

Looks to resolve https://github.com/sqlc-dev/sqlc/issues/383

Where the lack of `WithTx` in the interface causes makes it hard to use the `Querier` interface 
and the `*Queries` return type makes it hard to mock the sqlc `Querier` for testing

Please let me know if I should have approached this differently or if there's something else I need to do! ✌️


Jesse0Michael:
It would be nice if we could just update the **WithTx** signature to use "Querier" instead of "*Queries" _if_ `emit_interface` is enabled.. but, as @kyleconroy mentions in the issue https://github.com/sqlc-dev/sqlc/issues/383#issuecomment-596053061, that would be a breaking change. 

My hope is that we could put the change behind an explicit option.

I imagine `use_interface_return_types` isn't a clear enough name for an option that would fix the function signature of **WithTx**, but instead of guessing what would be a clearer name, I'd rather be told would be an acceptable option name

kyleconroy:
**🤖 Automated message**

---

PR #3182, opened 2024-02-05
docs: Add instructions for PostGIS/GEOS

twpayne:
This PR adds instructions on how to use sqlc with PostGIS via [GEOS](https://libgeos.org) (the standard library for geometry operations, as used by PostGIS itself) and [go-geos](https://github.com/twpayne/go-geos) (the official Go bindings for GEOS).


kyleconroy (approved):

kyleconroy:
Thank you!

PR #3191, opened 2024-02-11
Remove --experimental flag from the persistent FlagSet

jakoguta:
The `--experimental` flag was deprecated and later removed from the projects code. However, it seems that the command was not removed from the persistent `FlagSet`. Now this is confusing as it appears on the help list but it is not implemented.


kyleconroy:
We didn't want to remove it at that time as it might break some workflows. Happy to finally remove it now.

PR #3195, opened 2024-02-12
fix(generate): First pass at fixing path traversal vuln

kyleconroy:
Fixes https://github.com/sqlc-dev/sqlc/issues/3194

### DIFF
--- a/internal/cmd/generate.go
+++ b/internal/cmd/generate.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime/trace"
+	"strings"
 	"sync"
 
 	"google.golang.org/grpc"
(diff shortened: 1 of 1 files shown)


cbiffle:
This is _slightly_ overly conservative, technically speaking, in that it will prevent the creation of filenames like `out/foobar..txt`. I'm not terribly concerned about that, personally. However, I think you could fix it by testing to see if the path _starts with_ `..` -- any path returned from `Clean` (and thus `Join`) should only contain `..` iff it starts with `..`.

This still permits following symlinks in the out directory to arbitrary locations, if such symlinks exist. It looks like Go has a standard library routine for this in `EvalSymlinks` if you're concerned; you could apply it to the result of your `Join`. Given that WASM plugins have no way to create such a symlink atm, this is probably not a serious issue.

internal/cmd/generate.go
```diff
@@ -208,8 +209,21 @@ func (g *generator) ProcessResult(ctx context.Context, combo config.CombinedSett
...
+
 	for n, source := range files {
 		filename := filepath.Join(g.dir, out, n)
+		// filepath.Join calls filepath.Clean which should remove all "..", but
+		// double check to make sure
+		if strings.Contains(filename, "..") {
+			return fmt.Errorf("invalid file output path: %s", filename)
+		}
+		// The output file must be contained inside the output directory
+		if !strings.HasPrefix(filename, absout) {
```
cbiffle, line 224: Has `absout` also been run through `Clean`? If so, there's a possibility that `absout != filepath.Clean(absout)` which would cause this to fail.

kyleconroy: It has, since the result of `Join` is cleaned.

cbiffle: Ah, so it has. Thanks!

PR #3202, opened 2024-02-14
feat(sqlc): Support custom generic nullable types (i.e. `sql.Null[string]`)

gregoryjjb:
Support wrapping any nullable field with a generic type provided by the user. The config looks like this:

```yaml
sql:
  - gen:
      go:
        ...
        generic_null_type:
          import: "database/sql"
          type: "Null"
```

and the generated models look like this:

```go
type Foo struct {
	ID                    int64
	SometimesBlankText    sql.Null[string]
	SometimesBlankNumeric sql.Null[int64]
}
```

I only use sqlite but I added support & tests for Postgres and MySQL.

~The test cases use `sql.Null` even though the project isn't on Go 1.22 yet, but that doesn't seem to be an issue since the generated test package doesn't seem to be built/run anywhere?~

~nvm it has indeed been upgraded already.~

nvm it was only upgraded in the Dockerfile. I'll make a dummy type for the tests.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,7 @@ require (
 	github.com/davecgh/go-spew v1.1.1
 	github.com/fatih/structtag v1.2.0
 	github.com/go-sql-driver/mysql v1.7.1
-	github.com/google/cel-go v0.19.0
+	github.com/google/cel-go v0.20.0
 	github.com/google/go-cmp v0.6.0
 	github.com/jackc/pgx/v4 v4.18.1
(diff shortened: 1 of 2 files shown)


gregoryjjb:
I'm not supper happy about how this works (or rather doesn't work) with overrides. Ideally I would specify a custom override type without specifying its nullability, and then with this new option set it would wrap with the nullable type automatically (i.e. `sql.Null[MyCustomOverride]`).

gregoryjjb:
Superseded by #3279

PR #3223, opened 2024-02-26
dolphin: Fix the return type of SUM()

Jille:


kyleconroy:
Is this accurate? In PostgreSQL, `SUM` is polymorphic and returns the type of its input parameters.

Jille:
Good point. I'll write it out for all types, just like the PostgreSQL catalog does.

However, picking the right function from the catalog doesn't seem to work yet (https://play.sqlc.dev/p/a09737c4aa272d364f56e1a9b5d24d1424a5cc6b5b29ad8da220990129db6c92).

jozuenoon:
Hi! That would be very useful any chances to merge this?

PR #3247, opened 2024-03-05
docs(datatypes): improve details on TEXT

josethz00:
This is a small PR on the docs to help the developers to understand how the TEXT datatype works. I struggled a lot trying to override  a TEXT column to a `string`

### DIFF
--- a/docs/reference/datatypes.md
+++ b/docs/reference/datatypes.md
@@ -204,6 +204,10 @@ type Book struct {
 }
 ```
 
+## TEXT
+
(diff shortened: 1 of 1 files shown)


josethz00:
> This change isn't correct. By default, `TEXT` columns are mapped to strings. Nullable string are mapped to `pgtype.Text` when using the pgx/v5 driver. If you want nullable strings to map to `*string`, you can set `emit_pointers_for_null_types`.
> 
> Full example here https://play.sqlc.dev/p/3f99e092ec2143c1a7b62ef2ea2b037ff31c7c1d11e406d521037367238d8fa9

You are correct, my bad. I will change the PR

kyleconroy (approved):

PR #3287, opened 2024-03-22
enable go_struct_tag on column types

emanuelturis:
Hi,

So, first, this is my first contribution on open source, which means the quality of the pull request isn't the greatest, but I hope at least it leads to the right solution.

I noticed that when specifying the go_struct_tags on overrides, it would only work with column and not db_type.

I found something which to me seems like a workaround, but it now works on my end.

I ran the tests and it seems to all be passing just fine (5 failing tests before the change and after, so no difference there).

I'm looking forward to some feedback and see if this can be merged into the project.

Thanks to everyone!


kyleconroy:
**🤖 Automated message**

---

PR #3288, opened 2024-03-22
feat: trim output extension

stevenh:
Add an option to trim the extension of the source filename when writing outputs. This allows consumers to eliminate unnecessary .sql within the names of the generated files.


stevenh:
Oh, how do people ask for features like this?

PR #3298, opened 2024-03-28
fix(codegen/golang): Use int8 for MySQL TINYINT

kyleconroy:
Fixes https://github.com/sqlc-dev/sqlc/issues/3201


ariesdevil:
What if `sqlc.narg('A_TINYINT_COL')`? cause golang sql package has not `NullInt8`, just a `NullByte` but `byte` is `uint8`.

PR #3309, opened 2024-04-03
fix(docs): correct go install instructions

jdahm:
Go 1.21+ is required for recent versions of sqlc. This is because it uses components that require the 1.21 toolchain. Coincidentally 1.21 is also when forward compatibility was introduced.

You can test this
```
$ docker run -it --rm golang:1.19
% go install github.com/sqlc-dev/sqlc/cmd/sqlc@latest
...
pkg/mod/github.com/sqlc-dev/sqlc@latest/internal/bundler/upload.go:7:2: package log/slog is not in GOROOT (/usr/local/go/src/log/slog)
pkg/mod/github.com/sqlc-dev/sqlc@latest/internal/codegen/golang/opts/options.go:6:2: package maps is not in GOROOT (/usr/local/go/src/maps)
```

### DIFF
--- a/docs/overview/install.md
+++ b/docs/overview/install.md
@@ -14,18 +14,12 @@ brew install sqlc
 sudo snap install sqlc
 ```
 
-## go install 
+## go install
 
-### Go >= 1.17:
+Installing recent versions of sqlc requires Go 1.21+.
 
 ```
 go install github.com/sqlc-dev/sqlc/cmd/sqlc@latest
-```
-
-### Go < 1.17:
-
-```
-go get github.com/sqlc-dev/sqlc/cmd/sqlc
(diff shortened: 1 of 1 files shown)


docs/overview/install.md
```diff
@@ -14,20 +14,14 @@ brew install sqlc
...
 ```
 go install github.com/sqlc-dev/sqlc/cmd/sqlc@latest
 ```
 
-### Go < 1.17:
-
-```
-go get github.com/sqlc-dev/sqlc/cmd/sqlc
-```
-
```
jdahm, line 30: I removed this since it's irrelevant now that Go 1.21 is required for maps and slog.

PR #3354, opened 2024-04-25
Fix table inheritance

### LINKED ISSUE #3024: Inherited tables do not get updating after adding column to parent table
### Version

1.24.0

### What happened?

I created an inherited table in Postgres and everything worked as expected.

Then I decided I needed to add a column to the parent table, but when I did this the models for the inherited tables did not get updated with the new column.

### Relevant log output

_No response_

### Database schema

```sql
CREATE TABLE traffic_sensors(
    id uuid PRIMARY KEY DEFAULT gen_random_uuid(),
    serial_number varchar(16) NOT NULL,
    sampling_rate interval NOT NULL,
    base_station_serial varchar NOT NULL
);

CREATE TABLE traffic_sensors_random(
    max integer NOT NULL,
    min integer NOT NULL
)
INHERITS (
    traffic_sensors
);

ALTER TABLE traffic_sensors
    ADD sensor_type varchar;
```

### SQL queries

```sql
SELECT
    *
FROM
    traffic_sensors_random;
```

### Configuration

```yaml
version: "2"

sql:
- engine: "postgresql"
  queries: "sql/queries"
  schema: "sql/migrations"
  gen:
    go:
      package: "persist"
      out: "persist"
      emit_interface: true

overrides:
  go:
    overrides:
      - db_type: "pg_catalog.interval"
        engine: "postgresql"
        go_type:
          import: "github.com/jackc/pgtype"
          type: "Interval"
```

### Playground URL

_No response_

macOS

PostgreSQL

Go

dbhoot:
# Previously
Fixes #3024 

There are two main fixes in this PR

- Any alterations to a parent table never propagated to the child. This is because the columns of the parent table were only added on table create.
- Not null overrides on child tables were not applied correctly. This is because there was a shared column reference between the parent and child tables. When any override for any child table existed, it was applied to the shared reference, which actually affected (incorrectly) other tables.

# Now
- Changes to a parent table are propagated to the child tables
- Child tables have their own copy of the column definition. Overriding on a single table has no effect on other tables.


dbhoot:
hi @kyleconroy -- is there any hope of this change getting reviewed / merged? Are there concerns about it the implementation or anything? I know the core maintainers are busy so I'd like to do what I can to help.

ShadowBr0ther:
Hello @kyleconroy , It would be great if this could get reviewed because we really need this functionality.

PR #3370, opened 2024-05-09
hacky support for Postgres schemas w/ runtime interpolation

bgentry:
Hi @kyleconroy and @andrewmbenton 👋 @brandur and I have recently been looking more into making [River](https://github.com/riverqueue/river) fully support isolation at the Postgres schema level (that is, allowing different River clients to be configured to use different [Postgres schemas](https://www.postgresql.org/docs/current/ddl-schemas.html)). Our initial hope was to be able to rely on the `search_path` connection param as [recommended by pgx](https://github.com/jackc/pgx/issues/1013) and https://github.com/sqlc-dev/sqlc/discussions/2635 for this purpose.

Unfortunately, one of our users [quickly discovered](https://github.com/riverqueue/river/issues/345) that [this approach is basically incompatible with pgbouncer](https://github.com/pgbouncer/pgbouncer/issues/89). As far as I know the only ways to make this work with pgbouncer and sqlc would be:

1. Explicitly `SET search_path` prior to any query, _within a transaction_. This is error prone and has perf penalties in the form of added round trips.
2. Configure pgbouncer to isolate client pools for specific schemas at the user level. This is a pain to configure and precludes mixing and matching multiple schemas within a single conn pool. Not great.
3. Hypothetically, configure the database queries to explicitly specify the schema prefix on every query where it's required. I say hypothetically because this is of course not supported by sqlc today due to the fact that the schema can't be parameterized (it must be explicitly inserted into the SQL string).

I decided to spend chunk of time last night seeing if I could hack a way to achieve option (3) in sqlc. It actually wasn't too difficult to achieve in a hacky way, though doing so did cement my dislike of Go's `text/template` 😆 This draft PR shows one potential path for achieving this using a new `use_schema_placeholder` option, which when enabled causes all queries to be passed through a runtime string replacement on the known prefix placeholder `sqlc_schema_placeholder.`. It's definitely ugly and probably shouldn't be shipped in any form that resembles this PR, but I did want to at least raise it as a possible proof-of-concept.

Here's [a corresponding River commit](https://github.com/riverqueue/river/compare/bg-sqlc-hacky-schema-support) showing the queries that I updated to make use of the placeholder prefix. This sqlc PR does actually build and generate fully functional code for being able to use arbitrary schemas specified at runtime as shown in that commit.

I have lots of ideas for how this could be improved or done differently from the way I've done here (caching, maybe make this customizable per-query [ala Ecto](https://hexdocs.pm/ecto/Ecto.Query.html#module-query-prefix) instead of being statically wired into `Queries`, etc.), but rather than spin my wheels on them I thought I'd use this draft PR to start a discussion around if and how this might be achieved in sqlc given the fact that the previously recommended approach is not actually viable for those using pgbouncer in the most common way.

I also want to highlight that this is fairly distinct from allowing arbitrary runtime SQL substitution, because a schema prefix can be strictly validated to avoid i.e. injection concerns.

Related issues:

* https://github.com/sqlc-dev/sqlc/issues/2633
* https://github.com/sqlc-dev/sqlc/discussions/2635

### DIFF
--- a/internal/codegen/golang/gen.go
+++ b/internal/codegen/golang/gen.go
@@ -37,6 +37,7 @@ type tmplCtx struct {
 	EmitMethodsWithDBArgument bool
 	EmitEnumValidMethod       bool
 	EmitAllEnumValues         bool
+	UseSchemaPlaceholder      bool
 	UsesCopyFrom              bool
 	UsesBatch                 bool
 	OmitSqlcVersion           bool
@@ -177,6 +178,7 @@ func generate(req *plugin.GenerateRequest, options *opts.Options, enums []Enum, 
 		EmitMethodsWithDBArgument: options.EmitMethodsWithDbArgument,
 		EmitEnumValidMethod:       options.EmitEnumValidMethod,
 		EmitAllEnumValues:         options.EmitAllEnumValues,
+		UseSchemaPlaceholder:      options.UseSchemaPlaceholder,
 		UsesCopyFrom:              usesCopyFrom(queries),
 		UsesBatch:                 usesBatch(queries),
 		SQLDriver:                 parseDriver(options.SqlPackage),
--- a/internal/codegen/golang/imports.go
+++ b/internal/codegen/golang/imports.go
@@ -124,6 +124,10 @@ func (i *importer) dbImports() fileImports {
 		{Path: "context"},
 	}
 
+	if i.Options.UseSchemaPlaceholder {
+		std = append(std, ImportSpec{Path: "strings"})
+	}
+
 	sqlpkg := parseDriver(i.Options.SqlPackage)
 	switch sqlpkg {
 	case opts.SQLDriverPGXV4:
--- a/internal/codegen/golang/opts/options.go
+++ b/internal/codegen/golang/opts/options.go
@@ -44,6 +44,7 @@ type Options struct {
 	OmitUnusedStructs           bool              `json:"omit_unused_structs,omitempty" yaml:"omit_unused_structs"`
 	BuildTags                   string            `json:"build_tags,omitempty" yaml:"build_tags"`
 	Initialisms                 *[]string         `json:"initialisms,omitempty" yaml:"initialisms"`
+	UseSchemaPlaceholder        bool              `json:"use_schema_placeholder,omitempty" yaml:"use_schema_placeholder"`
 
 	InitialismsMap map[string]struct{} `json:"-" yaml:"-"`
 }
--- a/internal/codegen/golang/query.go
+++ b/internal/codegen/golang/query.go
@@ -278,7 +278,7 @@ func (q Query) hasRetType() bool {
 func (q Query) TableIdentifierAsGoSlice() string {
 	escapedNames := make([]string, 0, 3)
 	for _, p := range []string{q.Table.Catalog, q.Table.Schema, q.Table.Name} {
-		if p != "" {
+		if p != "" && p != "sqlc_schema_placeholder" {
 			escapedNames = append(escapedNames, fmt.Sprintf("%q", p))
 		}
 	}
--- a/internal/codegen/golang/struct.go
+++ b/internal/codegen/golang/struct.go
@@ -17,6 +17,7 @@ type Struct struct {
 }
 
 func StructName(name string, options *opts.Options) string {
+	name = strings.TrimPrefix(name, "sqlc_schema_placeholder")
 	if rename := options.Rename[name]; rename != "" {
 		return rename
 	}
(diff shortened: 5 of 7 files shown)


kyleconroy:
@bgentry Thanks for kicking off the conversation. I agree with you that this is a hacky solution, but I think we can figure out a solution. To confirm, when writing your queries, the SQL looks like the following:

```sql
-- name: GetAuthor :one
SELECT * FROM sqlc_schema_placeholder.authors
WHERE id = $1 LIMIT 1;
```

Instead of this, I was thinking we could add support for dynamic identifiers. The queries would look something like this

```sql
-- name: GetAuthor :one
SELECT * FROM "{schema}.authors"
WHERE id = $1 LIMIT 1;
```

This would add another parameter to the queries. The generated code would look like:

```gp
func GetAuthor(ctx context.Context, schema string, id int) {
    ....
}
```

elee1766:
> Do you have any thoughts on the query cache side of it? 

im not him but i do. i dont think a cache here is as bad as you think

> Obviously you wouldn't want to perform the same string substitution repeatedly hundreds of times per second.

i dont think this is obvious either. a string replacement is simple O(n) copy operation that can be optimized in advance since the we know how many to replace and actually we already know where the things we need to replace are. 

on the other hand, a lookup in cache/LRU is not only a O(n) operation since it needs to validate equality, but it also requires hashing and extra time to shuffle buckets, and it won't necessarily save you any string copies either.

A simple benchmark with naive strings.ReplaceAll shows this- I bet that some simple preprocessing and hits will make it faster and more memory efficient than the mapping. the speed is the same for small queries, and hash table is slower than replace for a large table, im assuming because of the hash. 

code:
```
package subbench_test

import (
	"strings"
	"testing"
)

func genQuery(replacement string, prefix, suffix int) string {

	return strings.Repeat("a", prefix) + replacement + strings.Repeat("b", suffix)
}

func BenchmarkReplaceSmallString(b *testing.B) {
	q := genQuery("{schema}", 30, 250)
	b.StartTimer()
	for i := 0; i < b.N; i++ {
		strings.ReplaceAll(q, "{schema}", "public")
	}
}

func BenchmarkReplaceLargeString(b *testing.B) {
	q := genQuery("{schema}", 1000000, 1000000)

	b.StartTimer()
	for i := 0; i < b.N; i++ {
		strings.ReplaceAll(q, "{schema}", "public")
	}
}

func BenchmarkCacheSmallString(b *testing.B) {
	q := genQuery("{schema}", 30, 250)
	ans := strings.ReplaceAll(q, "{schema}", "public")
	cache := map[string]string{}
	cache[q+"public"] = ans
	b.StartTimer()
	for i := 0; i < b.N; i++ {
		_ = cache[q+"public"]
	}
}

func BenchmarkCacheLargeString(b *testing.B) {
	q := genQuery("{schema}", 1000000, 1000000)
	ans := strings.ReplaceAll(q, "{schema}", "public")
	cache := map[string]string{}
	cache[q+"public"] = ans
	b.StartTimer()
	for i := 0; i < b.N; i++ {
		_ = cache[q+"public"]
	}
}
```


results:
```
goos: linux
goarch: amd64
pkg: benchmarks
cpu: 13th Gen Intel(R) Core(TM) i9-13900K
BenchmarkReplaceSmallString-32           9861722               194.3 ns/op           288 B/op          1 allocs/op
BenchmarkReplaceLargeString-32              3223            497422 ns/op         2008308 B/op          1 allocs/op
BenchmarkCacheSmallString-32             9496440               183.6 ns/op           320 B/op          1 allocs/op
BenchmarkCacheLargeString-32                3402            588169 ns/op         2009413 B/op          1 allocs/op
PASS
```

bgentry:
Hey @kyleconroy, wanted to see if you've thought any more about how this might be implemented or where it fits on your roadmap. As of now it's the only thing blocking River from having full support for multiple Postgres schemas, and while we really don't want to abandon sqlc it's a pretty important feature that we'll need to support soon one way or another.

elee1766:
> Instead of this, I was thinking we could add support for dynamic identifiers. The queries would look something like this
> 
> ```sql
> -- name: GetAuthor :one
> SELECT * FROM "{schema}.authors"
> WHERE id = $1 LIMIT 1;
> ```
> 
> This would add another parameter to the queries. The generated code would look like:
> 
> ```gnuplot
> func GetAuthor(ctx context.Context, schema string, id int) {
>     ....
> }
> ```

@kyleconroy at this point freeing up these sessions (thousands) that river requires because of this limitation of sqlc would be big difference for us. i'm glad to do all the work to implement this feature, but I want to confirm that this is still something you are interested in and would be okay with merging.

kyleconroy:
**🤖 Automated message**

---

kyleconroy:
**🤖 Automated message**

---

Issue #1171, opened 2021-09-09
Add support for PostgreSQL system catalogs


kyleconroy:
We currently don't have any support for the PostgreSQL [system catalogs](https://www.postgresql.org/docs/current/catalogs-overview.html).

To list all of the catalogs, we can use the following pg_dump command:

```
pg_dump -s -t pg_*.*
```

This output will need to be cleaned up before it can be parsed, as it includes some complex views that might not be parsed correctly.

See #1170 for an example of how this breaks queries.

PR #3442, opened 2024-06-13
Added MySQL System Catalogs

terricain:
I started out looking at #1171, realising that `sqlc` already contains all of the Postgres catalogs, then realised we didn't have any MySQL ones. Made a copy of the `sqlc-pg-gen` tool for MySQL, then tweaked the codegen to exclude the right schemas from being dumped into the output.

Adds:
* Tables, Views, Functions and Procedures from the `mysql`, `information_schema`, `performance_schema` and `sys` databases.
* `sqlc-dolphin-gen` tool to dump the above from a MySQL cluster.

Changes:
* Schema exclusion logic to exclude specific sets of schema depending on the engine in use.

Fixes:
* Typo in `sqlc-pg-gen`


terricain:
Ah looks like the schema exclusion bit in `internal/codegen/golang/result.go` needs to be propagated to kotlin plugin

terricain:
Have updated the kotlin plugin here - https://github.com/sqlc-dev/sqlc-gen-kotlin/pull/24

PR #3455, opened 2024-06-21
update: Set current year in docs and license

gitslav:
Updates the year in docs and LICENSE from 2023 to 2024


kyleconroy:
Thanks!

Issue #3444, opened 2024-06-13
"Relation does not exist" error when using SQLC with goose migrations 


rob2244:
### Version

1.26.0

### What happened?

I'm integrating sqlc into an existing golang project using Goose. When I run sqlc generate I'm getting the following error:

migrations/0036_source_table_embeddings.sql:1:1: relation "global_metric" does not exist

The table does exist and is in an earlier migration file. Strangely if remove the lines referencing that table in the migration file referenced in the error, sqlc generate runs successfully. 

The table in question is created in migration 5 and if I use it in any migration before migration 9 sqlc generate runs successfully, if I include it in any migration after migration 9 I get the above error of the reference not being found. The table is not deleted or modified in any way in migration 9.

### Relevant log output

```shell
migrations/0036_source_table_embeddings.sql:1:1: relation "global_metric" does not exist
```

### Database schema

```sql
-- Migration 5 (table is introduced)
-- +goose Up
CREATE TABLE IF NOT EXISTS client_portal.public.global_metric (
    id SERIAL primary key,
    metric varchar NOT NULL,
    user_id uuid NOT NULL,
    org_id uuid NOT NULL,
    formula varchar NOT NULL,
    required_metrics varchar,
    "description" varchar,
    tags varchar,
    tags_query tsquery,
    created_date timestamp DEFAULT (now() at time zone 'utc'),
    updated_date timestamp DEFAULT (now() at time zone 'utc'),
    UNIQUE(org_id, user_id, metric)
);

-- +goose Down
DROP TABLE IF EXISTS client_portal.public.global_metric;

-- Migration 9 (Last migration where I can include the table and it runs successfully)
-- +goose Up
CREATE TABLE IF NOT EXISTS client_portal.public.global_metric (
    id SERIAL primary key,
    metric varchar NOT NULL,
    user_id uuid NOT NULL,
    org_id uuid NOT NULL,
    formula varchar NOT NULL,
    required_metrics varchar,
    "description" varchar,
    tags varchar,
    tags_query tsquery,
    created_date timestamp DEFAULT (now() at time zone 'utc'),
    updated_date timestamp DEFAULT (now() at time zone 'utc'),
    UNIQUE(org_id, user_id, metric)
);

-- +goose Down
DROP TABLE IF EXISTS client_portal.public.global_metric;

-- Migration 10 (starts failing here if I include the table)
-- +goose Up
CREATE TABLE IF NOT EXISTS client_portal.public.global_concept (
[...]

PR #3481, opened 2024-07-07
fix(docs): Add ordering warning for numeric goose migration

rob2244:
When using Goose with numeric migrations, if the migration files aren't in lexographic order SQLC parses them in the wrong order. This is similar to the issue that occurs with golang-migrate. This pr documents this behaviour.

### DIFF
--- a/docs/howto/ddl.md
+++ b/docs/howto/ddl.md
@@ -145,6 +145,31 @@ type Post struct {
 
 ### goose
 
+**Warning:**
+sqlc parses migration files in lexicographic order. **If you are using numeric filenames for migrations in Goose and you choose to have sqlc enumerate your migration files**,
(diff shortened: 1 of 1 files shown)


rob2244:
Closes #3444

kyleconroy:
Thanks Rob. Longer term I do want to make sure we fix this. It may involve embedding goose inside sqlc itself. We'll see!

noahjamison:
@rob2244 @kyleconroy

I found a quick fix (see link below) that might be worth adding to the docs until the root problem gets solved. I'd be happy to spin up a quick PR if you think it'd be helpful for others!

https://github.com/sqlc-dev/sqlc/issues/3444#issuecomment-2606640205

PR #3493, opened 2024-07-19
fix(dbmanager): Add leading slash to db uri path rewrite

andrewmbenton:
Lack of leading slash was causing database connections to fail whenever dbmanager tried to access a newly-created database.


kyleconroy (approved):

andrewmbenton:
In general this path rewrite isn't safe I don't think, since the database name in the original URI could exist in a query parameter rather than the path and then there'd be a potential conflict. We should modify the database name after parsing I think.

Issue #3552, opened 2024-08-18
Table "product_metadata" incorrectly generates model name.


qanghaa:
### Version

1.27.0

### What happened?

I have 2 tables with name like `bonuses, product_metadata` when I using `sqlc` to generate models. The struct name generated seems unexpected
- table `bonuses` => struct name `Bonuse`
- table `product_metadata` => struct name `ProductMetaDatum`

### Playground URL

https://play.sqlc.dev/p/c459554e53b478b864dd0728b83783f6e7e22bc10c737cd3834f9df7435cec67

### What operating system are you using?

Linux

### What database engines are you using?

PostgreSQL

### What type of code are you generating?

Go

PR #3553, opened 2024-08-18
Fix incorrect generate model names for "metadata" suffix

qanghaa:
Fix [Table "bonuses" and "product_metadata" incorrectly generates model name. #3552](https://github.com/sqlc-dev/sqlc/issues/3552)

### DIFF
--- a/internal/endtoend/testdata/inflection/postgresql/pgx/v4/go/models.go
+++ b/internal/endtoend/testdata/inflection/postgresql/pgx/v4/go/models.go
@@ -4,11 +4,19 @@
 
 package querytest
 
+type Bonuse struct {
+	ID string
+}
+
 type Calorie struct {
 	ID string
 }
 
 type Campus struct {
+	ID string
+}
+
+type ProductMetadatum struct {
 	ID string
 }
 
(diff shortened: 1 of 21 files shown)


internal/endtoend/testdata/inflection/postgresql/pgx/v4/go/models.go
```diff
@@ -4,6 +4,10 @@
 
 package querytest
 
+type Bonuse struct {
+	ID string
+}
```
kyleconroy, line 9: This appears to still be wrong here?

qanghaa: yep, I expected "bonuses" table become "Bonus" after generated instead of "Bonuse"

qanghaa: Yep, I want the "bonuses" table to become "Bonus" instead of "Bonuse" after generate

qanghaa: I just deleted this one from PR 👍

Issue #3192, opened 2024-02-11
Improve "Using transactions" documentation


bbonagura9:
## Overview

I was trying to implement a simple transaction following the [documentation](https://docs.sqlc.dev/en/stable/howto/transactions.html) but it's not clear and the code return errors.

## Details to reproduce

Files:
```
├── dbsqlc
│   ├── db.go
│   ├── models.go
│   └── query.sql.go
├── go.mod
├── go.sum
├── main.go
├── query.sql
├── schema.sql
└── sqlc.yaml
```

schema.sql
```sql
CREATE TABLE records (
  id SERIAL PRIMARY KEY,
  counter INT NOT NULL
);
```

query.sql
```sql
-- name: GetRecord :one
SELECT * FROM records
WHERE id = $1;

-- name: UpdateRecord :exec
UPDATE records SET counter = $2
WHERE id = $1;
```

sqlc.yaml
```yaml
version: "2"
sql:
  - engine: "postgresql"
    queries: "query.sql"
    schema: "schema.sql"
    gen:
      go:
        package: "dbsqlc"
        out: "dbsqlc"
        sql_package: "pgx/v5"
```

main.go
```go
package foo

import (
	"context"
	"database/sql"
	"foo/dbsqlc"

	"github.com/jackc/pgx/v5"
)

func bumpCounter(ctx context.Context, db *sql.DB, queries *dbsqlc.Queries, id int32) error {
	tx, err := db.Begin()
	if err != nil {
		return err
	}
	defer tx.Rollback()
	qtx := queries.WithTx(tx)  // ERROR 1
	r, err := qtx.GetRecord(ctx, id)
	if err != nil {
		return err
	}
	if err := qtx.UpdateRecord(ctx, dbsqlc.UpdateRecordParams{
		ID:      r.ID,
		Counter: r.Counter + 1,
	}); err != nil {
		return err
	}
	return tx.Commit()
}

func main() {
  ctx := context.Background()
  conn, err := pgx.Connect(ctx, "host=localhost user=*** password=*** sslmode=disable")
  if err != nil {
    panic(err)
  }
  defer conn.Close(ctx)

  queries := dbsqlc.New(conn)

  bumpCounter(ctx, conn, queries, 1)  // ERROR 2
}
``` 

## Issues encontered

Just like the documentation indicates. But I came to two problems:

[...]

PR #3557, opened 2024-08-23
feat(docs): how-to use transactions with pgx

pedro-tiple:
Update the documentation as described in issue #3192.

### DIFF
--- a/docs/howto/transactions.md
+++ b/docs/howto/transactions.md
@@ -57,6 +57,7 @@ func (q *Queries) WithTx(tx *sql.Tx) *Queries {
 You'd use it like this:
 
 ```go
+// Using `github/lib/pq` as the driver.
 func bumpCounter(ctx context.Context, db *sql.DB, queries *tutorial.Queries, id int32) error {
 	tx, err := db.Begin()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


kyleconroy (approved):

fr11nik:
I prefer to create into db package ReadCommited function that start transaction 

```
// any sqlc function
type Handler func(*Queries) error

func (store *SQLStore) ReadCommited(ctx context.Context, fn Handler) error {
	tx, err := store.db.Begin(ctx)
	if err != nil {
		return err
	}
	defer func() {
		rbErr := tx.Rollback(ctx)
		if rbErr != nil {
			// Log the rollback error
			log.Printf("rollback error: %v", rbErr)
		}
	}()
	err = fn(store.WithTx(tx))
	if err != nil {
		// Return the original error along with the rollback error if any
		return fmt.Errorf("transaction error: %v", err)
	}

	return tx.Commit(ctx)
}
```

And use into **store.go** 

```
type Store interface {
	Querier
	ReadCommited(ctx context.Context, fn Handler) error
}
```

Use it via 
```
func bumpCounter(ctx context.Context,store tutorial.Store, id int32) error {
  err := store.ReadCommited(ctx,func(qtx *tutorial.Queries) error {
    r, err := qtx.GetRecord(ctx, id)
    if err != nil {
      return err
    }
    return qtx.UpdateRecord(ctx,tutorial.UpdateRecordParams{
      ID:      r.ID,
      Counter: r.Counter + 1,
    })
  })
  return err
}
```

PR #3601, opened 2024-09-12
docs: add sqlc-gen-from-template

fdietze:
I hope this helps with adoption.


kyleconroy:
Could you please add this to the new list of community projects? I think that's a better place for it.

fdietze:
Sure. You mean the table above? Or is there a newer list?

fdietze:
I put `Stable` in the database support section, because stability depends on your own template.

ghandic:
Is there a place we can add a link to language templates using this? How should it be structured?

PR #3620, opened 2024-09-29
Allow selective disabling of rules per query

### LINKED ISSUE #3619: Allow selective disabling of rules per query

Currently, the only way to [disable the vetting of rules](https://docs.sqlc.dev/en/stable/howto/vet.html#opting-out-of-lint-rules) for a query is with the `@sqlc-vet-disable` flag.
The problem with this is that it is an "all-or-nothing" approach. When provided, that flag disables the vetting of all rules for the query including some which might be needed.
It would be good to be able to provide a list of rules to skip, like: 
``` sqlc
/* @sqlc-vet-disable sqlc/db-prepare no-exec */
```

In this case, only the 2 listed rules in the comment would be skipped, but all others would run.
When no rule is provided, the flag should disable all rules in the query, just like the current behaviour, for backwards-compatibility.

#### Use Case
I have a basic `no-delete-without-where` rule which needs to be disabled for a few queries, but the current way to do that involves skipping all rules for the affected queries.

PostgreSQL, MySQL, SQLite

_No response_

rhodeon:
This resolves #3619 and has some minor refactor in regards to reusable constants.

As mentioned there, the `@sqlc-vet-disable` query annotation now accepts a list of rules to skip:
```sql
/* @sqlc-vet-disable sqlc/db-prepare no-pg */
```
Providing just `/* @sqlc-vet-disable */` without any parameters retains the previous behaviour of skipping all rules for a query.

### DIFF
--- a/docs/howto/vet.md
+++ b/docs/howto/vet.md
@@ -276,7 +276,7 @@ SELECT * FROM authors
 WHERE id = ? LIMIT 1;
 ```
 
-To skip all rules for a query, you can provide the `@sqlc-vet-disable` annotation without any options. 
+To skip all rules for a query, you can provide the `@sqlc-vet-disable` annotation without any parameters. 
 
 ```sql
 /* name: GetAuthor :one */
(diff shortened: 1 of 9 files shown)


kyleconroy:
@rhodeon thanks for the updated push. The docs and tests are great! This will go out with the next release.

PR #3640, opened 2024-10-08
fix(dbmanager): use correct SQL to drop databases

scottt:
Before this patch, `ManagedClient.CreateDatabase` could issue the following SQL:

```
DROP DATABASE mydb IF EXISTS WITH (FORCE)
```

which is not standard SQL and does not work with PostgreSQL. See https://www.postgresql.org/docs/current/sql-dropdatabase.html

This patch fixes to code to issue:

```
DROP DATABASE IF EXISTS mydb WITH (FORCE)
```

@kyleconroy , this patch fixes the DROP DATABASE syntax error introduced in https://github.com/sqlc-dev/sqlc/pull/3525 and https://github.com/sqlc-dev/sqlc/pull/3421


kyleconroy:
Thank you!

PR #3647, opened 2024-10-09
fix(golang): escape q field name

### LINKED ISSUE #3627: Invalid go generated code when there's only one field and it's named "q"
### Version

1.27.0

### What happened?

The generated function `func (q *Queries) GetAuthorWithQ(ctx context.Context, q sql.NullString) ([]Author, error)` has both a `q` struct and `q` parameter, which is invalid go syntax.

### Relevant log output

_No response_

### Database schema

```sql
CREATE TABLE authors (
  id   BIGSERIAL PRIMARY KEY,
  name text      NOT NULL,
  bio  text,
  q    text
);
```

### SQL queries

```sql
-- name: GetAuthorWithQ :many
SELECT * FROM authors
WHERE q = $1
ORDER BY name;
```

### Configuration

_No response_

### Playground URL

https://play.sqlc.dev/p/ba6e22e495b02c910c219f86bc6b9b145bbf2af16fff7913059fdb26889b4ec3

macOS

PostgreSQL

Go

vitords:
Adds `q` to the reserve keywords.

Although not a golang keyword per se, it works as one since `q` is reserved as a shorthand for `Queries` in the generated code so there's no need to rename the reserved keywords list.

Fixes #3627.

### DIFF
--- a/internal/codegen/golang/reserved.go
+++ b/internal/codegen/golang/reserved.go
@@ -59,6 +59,8 @@ func IsReserved(s string) bool {
 		return true
 	case "var":
 		return true
+	case "q":
+		return true
 	default:
 		return false
 	}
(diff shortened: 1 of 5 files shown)


kyleconroy (approved):
Thanks! I've been meaning to fix this for a while.

PR #3657, opened 2024-10-16
fix (engine/sqlite): normalize table names (#3162)

### LINKED ISSUE #3162: warn when using quotes or parse the SQL

I have the following setup:

queries.sql
```sql
-- name: CheckDatabase :exec
SELECT
	COUNT(ID),
	COUNT(Destination),
	COUNT(Hits),
	COUNT(Owner)
FROM redirections;
```

schema.sql
```sql
CREATE TABLE "redirections" (
	"ID"	TEXT NOT NULL,
	"Destination"	TEXT NOT NULL,
	"Hits"	INTEGER NOT NULL DEFAULT 0,
	"Owner"	TEXT NOT NULL,
	PRIMARY KEY("ID")
);
```

sqlc.json
```json
{
	"version": "2",
	"sql": [{
		"schema": "./internal/datastore/db/schema.sql",
		"queries": "./internal/datastore/db/queries.sql",
		"engine": "sqlite",
		"gen": {
			"go": {
				"out": "db"
			}
		}
	}]
}
```

I think this should work, and in fact play.sqlc.dev agrees with me there. BUT
```console
$ go run github.com/sqlc-dev/sqlc/cmd/sqlc@latest generate
# package 
internal/datastore/db/queries.sql:1:1: relation "redirections" does not exist
exit status 1
```

Thanks to @qbit, I learned about the use of quotations in the `CREATE` statement leading to all uses of the identifier having to be quoted. Leaving aside how … unideal … this limitation is, I would consider this unexpected behaviour, so I propose one of the following solutions:

- actually parse the name of the identifier (**preferred**)
- warn the user about this pitfall (if not in the CLI output, then in the docs)

Since I don't think this affects any specific database engine, or language backend, I have not selected any of them. If that's incorrect, I apologise.

_No response_

_No response_

jvns:
This aims to fix #3162, where in SQLite if `schema.sql` contained  `CREATE TABLE "users"` and `query.sql` contained `SELECT * from users`, it would complain `relation "users" does not exist`, because the quotes don't match.

I think this was happening because the `identifier` function (which removes quotes from table names) wasn't consistently being called everywhere in the parser. I went through and replaced all instances of `n.Table_name().GetText()` with `identifier(n.Table_name().GetText())` and that seems to have fixed the issue.

I've also added a test which previously failed and now passes. Here's what the output was previously:

```
$ cd internal/endtoend/testdata/quoted_tablename/sqlite/
$ sqlc generate
# package querytest
query.sql:1:1: relation "users" does not exist
```


kyleconroy:
hey! long time

I've been a bit underwater with pull requests. Appreciate the added tests.

Do you mind rebasing on main? The tests didn't run in CI

kyleconroy (approved):

Issue #3514, opened 2024-08-02
Go panics when sub query is missing an alias with sqlite


leetrout:
### Version

1.26.0

### What happened?

I accidentally left an alias off my subquery and I got a panic instead of a useful error message.

Failing playground without alias (included below as well): https://play.sqlc.dev/p/9bd88a936c60738a57503c6314f5805316327b3dbf51a03c49518ff729382b44

Passing playground with alias: https://play.sqlc.dev/p/47e54486a69100bf2e6e61436518f642835859e73a7ba42ab89f0f74234be50c

Example with postgres that gives a useful error message instead of the panic: https://play.sqlc.dev/p/dd2d6eaf25d9bef2b07e1cd7dc5a4b949694ba6d200b419724d13a513c1eaf1a

### Relevant log output

```shell
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x10d0247]

goroutine 6 [running]:
github.com/sqlc-dev/sqlc/internal/compiler.(*Compiler).sourceTables(0xc00049c008, 0xc00048ed50, {0x1d84f80?, 0xc0009d9ad0?})
        /home/user/.asdf/installs/golang/1.22.2/packages/pkg/mod/github.com/sqlc-dev/sqlc@v1.26.0/internal/compiler/output_columns.go:601 +0xaa7
github.com/sqlc-dev/sqlc/internal/compiler.(*Compiler).outputColumns(0xc00049c008, 0xc00048ed50, {0x1d84f80, 0xc0009d9ad0})
        /home/user/.asdf/installs/golang/1.22.2/packages/pkg/mod/github.com/sqlc-dev/sqlc@v1.26.0/internal/compiler/output_columns.go:55 +0x45
github.com/sqlc-dev/sqlc/internal/compiler.(*Compiler)._analyzeQuery(0xc00049c008, 0xc000f8fde0, {0xc000062fcc, 0x83}, 0x1)
        /home/user/.asdf/installs/golang/1.22.2/packages/pkg/mod/github.com/sqlc-dev/sqlc@v1.26.0/internal/compiler/analyze.go:180 +0xced
github.com/sqlc-dev/sqlc/internal/compiler.(*Compiler).analyzeQuery(...)
[...]

Issue #3337, opened 2024-04-15
1.26.0 crashes on subquery without alias


mishak87:
### Version

1.26.0

### What happened?

1.26.0 crashes while
1.25.0 reports `query.sql:5:5: subquery in FROM must have an alias`

### Relevant log output

```shell
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x10d0027]

goroutine 13 [running]:
github.com/sqlc-dev/sqlc/internal/compiler.(*Compiler).sourceTables(0xc00035f208, 0xc000736f60, {0x1d84e80?, 0xc0006e7b80?})
        /home/mishak/go/pkg/mod/github.com/sqlc-dev/sqlc@v1.26.0/internal/compiler/output_columns.go:601 +0xaa7
github.com/sqlc-dev/sqlc/internal/compiler.(*Compiler).outputColumns(0xc00035f208, 0xc000736f60, {0x1d84e80, 0xc0006e7b80})
        /home/mishak/go/pkg/mod/github.com/sqlc-dev/sqlc@v1.26.0/internal/compiler/output_columns.go:55 +0x45
github.com/sqlc-dev/sqlc/internal/compiler.(*Compiler)._analyzeQuery(0xc00035f208, 0xc000052be0, {0xc0006e8be3, 0x3f}, 0x1)
        /home/mishak/go/pkg/mod/github.com/sqlc-dev/sqlc@v1.26.0/internal/compiler/analyze.go:180 +0xced
github.com/sqlc-dev/sqlc/internal/compiler.(*Compiler).analyzeQuery(...)
        /home/mishak/go/pkg/mod/github.com/sqlc-dev/sqlc@v1.26.0/internal/compiler/analyze.go:110
github.com/sqlc-dev/sqlc/internal/compiler.(*Compiler).parseQuery(0xc00035f208, {0x1d85040, 0xc000052be0}, {0xc0006e8800, 0x73e}, {{0x0, 0x0, {0x0, 0x0}, 0x0, ...}})
        /home/mishak/go/pkg/mod/github.com/sqlc-dev/sqlc@v1.26.0/internal/compiler/parse.go:103 +0x5d5
github.com/sqlc-dev/sqlc/internal/compiler.(*Compiler).parseQueries(0xc00035f208, {{0x0, 0x0, {0x0, 0x0}, 0x0, 0x0, 0x0, 0x0}})
[...]

Issue #2453, opened 2023-07-13
sqlc panics on union all with subqueries using mysql


andrewmbenton:
### Version

Other

### What happened?

From https://github.com/kyleconroy/sqlc/issues/2416#issuecomment-1633767537, sqlc panics on the below query going back to at least version v1.14.0.

### Relevant log output

```shell
panic: interface conversion: ast.Node is *ast.SetOprSelectList, not *ast.SelectStmt

goroutine 6 [running]:
github.com/kyleconroy/sqlc/internal/engine/dolphin.(*cc).convertSetOprSelectList(0xc000a7ebc0?, 0xc0001e4930)
        /go/pkg/mod/github.com/kyleconroy/sqlc@v1.19.0/internal/engine/dolphin/convert.go:1212 +0x505
github.com/kyleconroy/sqlc/internal/engine/dolphin.(*cc).convertSetOprStmt(0x20?, 0xc000100000?)
        /go/pkg/mod/github.com/kyleconroy/sqlc@v1.19.0/internal/engine/dolphin/convert.go:1255 +0x7b
github.com/kyleconroy/sqlc/internal/engine/dolphin.(*cc).convert(0xc000810000?, {0x2804eb0?, 0xc0001e48c0?})
        /go/pkg/mod/github.com/kyleconroy/sqlc@v1.19.0/internal/engine/dolphin/convert.go:1735 +0x20f6
github.com/kyleconroy/sqlc/internal/engine/dolphin.(*Parser).Parse(0xc000015130, {0x27f0380?, 0xc000a4e0a0?})
        /go/pkg/mod/github.com/kyleconroy/sqlc@v1.19.0/internal/engine/dolphin/parse.go:63 +0x205
github.com/kyleconroy/sqlc/internal/compiler.(*Compiler).parseQueries(0xc00021dc00, {{0x40?, 0xf1?, {0x0?, 0x1?}, 0x1?}})
        /go/pkg/mod/github.com/kyleconroy/sqlc@v1.19.0/internal/compiler/compile.go:74 +0x26d
github.com/kyleconroy/sqlc/internal/compiler.(*Compiler).ParseQueries(...)
        /go/pkg/mod/github.com/kyleconroy/sqlc@v1.19.0/internal/compiler/engine.go:49
[...]

PR #3711, opened 2024-11-16
fix: range subselect alias pointer dereference

ahme-dev:
Made to solve dereference panics reported by a list of open issues (I've also seen it in my codebase):

1. #3337 
2. #3514
3. #2453

Queries and schemas combined below, none of which panic anymore on this branch.
https://play.sqlc.dev/p/f1541063c62ed9302d43b7d00eec0d487764e5ee577a43e4114ff6c654585768


andrewmbenton:
The third item in your list is unrelated, but the other two are valid. I will attempt to add some tests to your branch so that we can evaluate and merge.

kyleconroy (approved):

PR #3728, opened 2024-11-28
fix(postgresql): Support renaming and dropping materialized views

swallowstalker:
solving the issue here https://github.com/sqlc-dev/sqlc/issues/3371

basically if there are create mv - rename mv - drop mv, then on drop mv statement, the sqlc generate will fail because mv involved in rename statement is not registered to the list of statements, in short make them invisible to drop mv operations.

before
```
$ sqlc-dev generate -f internal/endtoend/testdata/ddl_rename_drop_materialized_views/postgresql/pgx/v5/sqlc.yaml
# package querytest
schema.sql:1:1: relation "authors_mv_old" does not exist
```

after (no error)
```
$ sqlc-dev generate -f internal/endtoend/testdata/ddl_rename_drop_materialized_views/postgresql/pgx/v5/sqlc.yaml

```


swallowstalker:
hi @kyleconroy, as this is my first PR in this repo, let me know if i needed to add some additional testing for this (i'm unable to assign you as reviewer in this PR, maybe its maintainer only)

kyleconroy:
Thanks for the excellent first pull request. For future changes, this is the correct amount of testing for a change like this.

PR #3739, opened 2024-12-10
feat: adding nullable text if selected column is json operator and casted as text

swallowstalker:
Trying to solve https://github.com/sqlc-dev/sqlc/issues/3710

Adding `IsJSONOperator` to detect if given `ast.A_Expr` contains operator specific to postgresql JSON. If it is detected within type cast `ast.TypeCast`, then the query column should be set to nullable (by default it is not null).

`(mt.myjson->'thing1'->'thing2')::text` should be converted to `Column3 pgtype.Text`

but for non-type cast select column, I think that it should still be `interface{}` because we don't know user intentions and how the data structure under that field.


veqryn:
Can we add this test case as well please?
```
-- name: GetNullable2B :many
SELECT CASE WHEN true THEN 'hello'::text ELSE null::text END 
FROM "mytable";
```

veqryn:
Where in your tests are we setting the expectation of the generated results?
Like, I would have expected your `GetNullable2` to return an `any`/`interface{}`, but I don't see the expectation set anywhere.  Right now, the generated code appears to return a `pgtype.Text`, which is odd for something returning integers, floats, and strings.

Or is the committed code the expectation?

swallowstalker:
> Can we add this test case as well please?

GetNullable3B added

> Like, I would have expected your GetNullable2 to return an any/interface{}

In my PR description I proposed type precedence, for multiple type inside case expression. For example

- 1st condition result is int, 2nd condition result is float, 3rd condition result is text, and default condition is null, then it should be nullable text, not an interface
- 1st condition result is int, 2nd condition result is float, and default condition is int, then it should be non-nullable float, not an interface
- 1st condition result is int, and default condition is null, then it should be nullable int, not an interface

and so on. but on the other side, I think that the user should be given freedom if the return value of case expr has several type variance (in short, just return `any` or `interface{}`). @kyleconroy maybe you have feedback on this particular case? 👀

swallowstalker:
> Or is the committed code the expectation?

I have generated the result using code from this PR, and then make it as expectation (commited on repo), but to be honest I have a problem in github workflow tests, it seems that the binary that the workflow use is the previous version before I made these changes, so the generated go file would always be different than the one I commited on the repo. Especially on `TestReplay` endtoend tests. Maybe somebody can help giving me some pointers?

veqryn:
So after thinking about it, I realized we really only care about what Postgres thinks the returned values are, not what we think they should be.
I tried various combinations of your GetNullable2, and found that postgres doesn't really care about the order of which case/when statement comes first, but instead has some kind of preference, which you mentioned.

For example, if you insert at least 4 rows into `mytable`, then running this query:
```sql
SELECT CASE
    WHEN id = 1 THEN '1'
    WHEN id = 2 THEN null
    WHEN id = 3 THEN id
    WHEN id = 4 THEN 4.5
    ELSE '10'
END
FROM "mytable";
```

Returns:
```
1
[NULL]
3
4.5
10
10
```

The strings for `'1'` and `'10'` got cast into numbers.  `id` is still an integer/number.  The decimal or float of `4.5` got to stay as is.  I suppose the correct type for this would be a nullable decimal?

But if you change the query to this:
```sql
SELECT CASE
    WHEN id = 1 THEN '1'::text
    WHEN id = 2 THEN null
    WHEN id = 3 THEN id
    WHEN id = 4 THEN 4.5
    ELSE '10'
END
FROM "mytable";
```
You get an error: `SQL Error [42804]: ERROR: CASE types text and bigint cannot be matched`

or this:
```sql
SELECT CASE
    WHEN id = 1 THEN 'foobar'
    WHEN id = 2 THEN null
    WHEN id = 3 THEN id
    WHEN id = 4 THEN 4.5
    ELSE '10'
END
FROM "mytable";
```
You get a similar error: `SQL Error [22P02]: ERROR: invalid input syntax for type numeric: "foobar"`


So to sum up: I think you are on the right track, with forcing these types of queries into a type rather than an any/interface{}

I don't know how realistic these are in the real world.  Perhaps we should try to find queries that hit an actual table of data, rather than case statements of literals.

swallowstalker:
I agree that it's not quite realistic, I better left it to sqlc user to decide which data type that will be used in case of multiple data type exist on case-when-then-else. In short, I'll make it interface then

internal/compiler/output_columns.go
```diff
@@ -764,3 +779,74 @@ func findColumnForRef(ref *ast.ColumnRef, tables []*Table, targetList *ast.List)
...
+	switch aconst.Val.(type) {
+	case *ast.String:
+		col = &Column{Name: name, DataType: "text", NotNull: true}
+	case *ast.Integer:
+		col = &Column{Name: name, DataType: "int", NotNull: true}
+	case *ast.Float:
+		col = &Column{Name: name, DataType: "float", NotNull: true}
+	case *ast.Boolean:
+		col = &Column{Name: name, DataType: "bool", NotNull: true}
+	case *ast.Null:
```
swallowstalker, line 846: this is new addition to handle null const value. I need to distinguish between `any` and `null`, because `null` in `CaseExpr` would determine query column result nullability, but does not make it included in column type decision.

for `ast.A_Const` condition, i have converted DataType `null` back to `any` (line 136, `outputColumns` function) because we don't know its type and to preserve backward compatibility

internal/compiler/output_columns.go
```diff
@@ -183,44 +168,70 @@ func (c *Compiler) outputColumns(qc *QueryCatalog, node ast.Node) ([]*Column, er
...
+
+			var defaultCol *Column
+			if n.Defresult.Pos() != 0 {
+				defaultCol, err = convertCaseExprCondToColumn(n.Defresult, &name)
+				if err != nil {
+					return nil, err
 				}
 			} else {
-				cols = append(cols, &Column{Name: name, DataType: "any", NotNull: false})
+				defaultCol = &Column{Name: name, DataType: "null", NotNull: false}
```
swallowstalker, line 209: missing `ELSE` clause on CaseExpr is considered to return null value

veqryn:
Based on the tests, I believe this fixes the issue for me.
How do we get this reviewed by sqlc devs?

swallowstalker:
from my previous PR, I mentioned @kyleconroy to ask for review. would you mind checking it @kyleconroy? sorry i still can't assign reviewer for this PR

swallowstalker:
bumping this PR, is there anything more i can do for this PR?

PR #3755, opened 2024-12-20
fix: cte and subquery relation name and columns should be registered on query catalog

swallowstalker:
Trying to solve https://github.com/sqlc-dev/sqlc/issues/3720

### CTE columns cannot be referred on main query

This is because if table name is not found on catalog, sqlc will attempt to find it on query catalog, but once it is found, it is skipped and not registered on QueryCatalog, making it invisible for it to be referred for the next steps. https://github.com/sqlc-dev/sqlc/blob/996f7d1dab74a913afd6739a55ae034a8ed9cac5/internal/compiler/resolve.go#L70-L73

### Subquery relation name and columns cannot be referred on `WHERE` clause

Similar to above case, but this needs to be detected on `*ast.RangeSubselect` to register alias of subquery and its output column into QueryCatalog


swallowstalker:
just realized that there is open PR about this issue after I finished create this MR 👀 for now maybe I will wait for PR https://github.com/sqlc-dev/sqlc/pull/3220 to be merged as it has more complete tests

PR #3759, opened 2024-12-22
fix: Do not close wazero module on error (#3758)

### LINKED ISSUE #3758: WASM plugin error causes SIGSEGV
### Version

1.27.0

### What happened?

sqlc throws a SIGSEGV when a WASM plugin fails to load.

That's caused by https://github.com/sqlc-dev/sqlc/blob/main/internal/ext/wasm/wasm.go#L226 where sqlc calls `result.Close()` on a wazero module, but wazero already closes modules on error.

### Relevant log output

```shell
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0xb0 pc=0x1140027]

goroutine 8 [running]:
github.com/tetratelabs/wazero/internal/wasm.(*ModuleInstance).setExitCode(...)
        github.com/tetratelabs/wazero@v1.7.3/internal/wasm/module_instance.go:142
github.com/tetratelabs/wazero/internal/wasm.(*ModuleInstance).CloseWithExitCode(0x8?, {0x1dce8a0?, 0xc00036dc20?}, 0x1?)
        github.com/tetratelabs/wazero@v1.7.3/internal/wasm/module_instance.go:101 +0x27
github.com/tetratelabs/wazero/internal/wasm.(*ModuleInstance).Close(0xc0007bc090?, {0x1dce8a0?, 0xc00036dc20?})
        github.com/tetratelabs/wazero@v1.7.3/internal/wasm/module_instance.go:96 +0x1f
github.com/sqlc-dev/sqlc/internal/ext/wasm.(*Runner).Invoke(0xc0009cf3c0, {0x1dce8a0, 0xc00036dc20}, {0x1aa58a8, 0x1f}, {0x19991e0?, 0xc0007fa630?}, {0x18e1f80, 0xc0009cf400}, {0x0, ...})
        github.com/sqlc-dev/sqlc/internal/ext/wasm/wasm.go:229 +0x1039
github.com/sqlc-dev/sqlc/internal/plugin.(*codegenServiceClient).Generate(0xc0005f8ad8, {0x1dce8a0, 0xc00036dc20}, 0xc0007fa630, {0x0, 0x0, 0x0})
        github.com/sqlc-dev/sqlc/internal/plugin/codegen_grpc.pb.go:42 +0xc8
github.com/sqlc-dev/sqlc/internal/cmd.codegen({_, _}, {{{0x33eb9d0, 0x1}, {{0x0, 0x0}, {0x0, 0x0}, {0x0, 0x0}, ...}, ...}, ...}, ...)
        github.com/sqlc-dev/sqlc/internal/cmd/generate.go:410 +0x552
github.com/sqlc-dev/sqlc/internal/cmd.(*generator).ProcessResult(_, {_, _}, {{{0x33eb9d0, 0x1}, {{0x0, 0x0}, {0x0, 0x0}, {0x0, ...}, ...}, ...}, ...}, ...)
        github.com/sqlc-dev/sqlc/internal/cmd/generate.go:203 +0xa5
github.com/sqlc-dev/sqlc/internal/cmd.processQuerySets.func1()
        github.com/sqlc-dev/sqlc/internal/cmd/process.go:113 +0x8f0
golang.org/x/sync/errgroup.(*Group).Go.func1()
        golang.org/x/sync@v0.8.0/errgroup/errgroup.go:78 +0x50
created by golang.org/x/sync/errgroup.(*Group).Go in goroutine 1
        golang.org/x/sync@v0.8.0/errgroup/errgroup.go:75 +0x96
```

### Database schema

```sql
NA
```

### SQL queries

```sql
NA
```

### Configuration

```yaml
NA
```

### Playground URL

_No response_

_No response_

_No response_

_No response_

edman:
Fixes #3758

wazero takes care not to leak modules when `InstantiateModule` returns an error.

~This means that sqlc will call `result.Close` a second time if a non nil module was returned, which causes a segfault.~

Edit: this statement is wrong. See my comment below for the correct explanation of the segfault.

### DIFF
--- a/internal/ext/wasm/wasm.go
+++ b/internal/ext/wasm/wasm.go
@@ -222,7 +222,7 @@ func (r *Runner) Invoke(ctx context.Context, method string, args any, reply any,
 	}
 
 	result, err := runtimeAndCode.rt.InstantiateModule(ctx, runtimeAndCode.code, conf)
-	if result != nil {
+	if err == nil {
 		defer result.Close(ctx)
 	}
 	if cerr := checkError(err, stderr); cerr != nil {


kyleconroy:
> wazero takes care not to leak modules when InstantiateModule returns an error.

So does wazero return a valid module and an error message in these cases? `checkError` also looks at stderr, which make be written to if `InstantiateModule` does not return an error. Merged as is we wouldn't call `result.Close` in all cases.

edman:
> So does wazero return a valid module and an error message in these cases?

Ok so.. Your comment made me go back and track more closely what happened. And I found sqlc actually tripped this https://go.dev/doc/faq#nil_error/ interesting behavior of go with `nil` values and interfaces.

See this example https://go.dev/play/p/66QShXidalq

Basically we can't always compare `result != nil`, where `result` is an interface type. That might work or not depending on wazero's code. (it gets awkward when you remember that `error` is an interface).

> checkError also looks at stderr, which make be written to if InstantiateModule does not return an error. Merged as is we wouldn't call result.Close in all cases.

`checkError` doesn't look at `stderr` when `err` is `nil` to begin with, so it should close in all relevant cases? But I agree the indirection makes things less clear.

I'm thinking we can just defer close when `err == nil`? Please see the new commit.

Issue #3430, opened 2024-06-07
Allow for passing comments through the table schema to the generated Go Struct


PaulSonOfLars:
### What do you want to change?

Currently, generating a table from the basic SQL will strip the field comments completely:
```
CREATE TABLE test (
  id text PRIMARY KEY,
-- Some helpful comment to pass on to the go code
  some_field text NOT NULL
)
```

It would be great if we could write comments in our table description, and have that comment show up in our go code as it gets run. Some fields need more than just a name, and having these would be a really helpful way of reading the code. Most people will read the go struct, not go hunt for the SQL file with the comments themselves.

Happy to open a PR for it if theres interest :) 

### What programming language backends need to be changed?

Go

ChocoChipset:
Migrated recently to sqlc. I'm missing the comments in my old classes. :(

ChocoChipset:
For anyone interested: this can be achieved by using `comment on`


```
CREATE TABLE authors (
  id   BIGSERIAL PRIMARY KEY,
  name text      NOT NULL,
  bio  text
);
comment on table authors is 'contains authors in the library';
comment on column authors.name is 'name of the author starting with first name';
```

oxisto:
> For anyone interested: this can be achieved by using `comment on`
> 
> ```
> CREATE TABLE authors (
>   id   BIGSERIAL PRIMARY KEY,
>   name text      NOT NULL,
>   bio  text
> );
> comment on table authors is 'contains authors in the library';
> comment on column authors.name is 'name of the author starting with first name';
> ```

Unfortunately this does not work for sqlite :(

PR #3766, opened 2024-12-27
Passing SQLite comments to the generated struct

oxisto:
This is a partial solution for #3430 for sqlite, which may or may not be adaptable for other engines. It basically looks for single-line comments before the CREATE command and in-between column definitions.


oxisto:
The test fails are „intentional“. The generated structs are now different because they contain comments. I can either adjust the tests or provide a configuration flag to enable/disable this feature.

PR #3768, opened 2024-12-31
perf: preallocate slice with known length to optimize Mem Allocs

anazcodes:


kyleconroy:
I'm skeptical that this one change is causing much of a performance problem, so going to close this for now.

anazcodes:
> I'm skeptical that this one change is causing much of a performance problem, so going to close this for now.

It may not provide much observable effect in a few iterations, but it significantly improves performance with frequent iterations. Pre-allocating memory for the slice with a known size reduces the processes that occur internally when calling append in a loop.

This PR was intended to write the better version of the noticed line of code.

PR #3772, opened 2025-01-04
Add support for getting db session form context

rodionstolyarov:
sqlc already have an option to emit the methods with db argument `emit_methods_with_db_argument`, but sometimes it is more convenient to keep the current db session in the context. It would be nice if sqlc could extract db from the context itself. 

To do this, we need to pass a function that knows how to get db form the context to the `New` method of `Querier` and then use this function in all the methods generated.

This PR adds support for a new option `get_db_from_context` that implements this feature.


kyleconroy:
@rodionstolyarov Thank you very much for your first contribution to the project!

While I understand that you'd like this feature, it's not something that's been asked for often. Since it adds for configuration options and complicates the generated code. Instead, I'd suggest building this either as a plugin or a separate codegen package that wraps the existing `emit_methods_with_db_argument` code with methods that pull the database session out of the context and pass it to those methods.

rodionstolyarov:
I see, it's wise not to complicate the code unnecessarily. Thanks for your feedback!

PR #3824, opened 2025-02-05
docs: Add Zig plugin

tinyzimmer:
I've got a new plugin for Zig that seems mostly stable and passes some basic tests. Figured I'd add it to the list to get it some exposure and start uncovering edge cases.


kyleconroy (approved):

kyleconroy:
@tinyzimmer Excited to see how the Zig support works out

Issue #3837, opened 2025-02-12
Nullable datetimes are used for timestaps marked as not null in PGX / GO


benjaco:
### Version

1.28.0

### What happened?

For pgx, `pgtype.Timestamp` is used for `TIMESTAMP` (and other time related types) regardless if the column is nullable or not.

`time.Time` was expected to be used for timestamps marked with `NOT NULL`

"database/sql" is handling it correctly https://play.sqlc.dev/p/e1e23e0d41d3d1340a4a009099bb81b589385b49187b89fd40c9aee4c33097cf

### Relevant log output

```shell

```

### Database schema

```sql
CREATE TABLE authors (
  id   BIGSERIAL PRIMARY KEY,
  name text      NOT NULL,
  bio  text,
  
  born TIMESTAMP      NOT NULL,
  died  TIMESTAMP
);
```

### SQL queries

```sql
-- name: ListAuthors :many
SELECT * FROM authors
ORDER BY name;
```

### Configuration

```yaml
{
  "version": "2",
  "sql": [{
    "schema": "schema.sql",
    "queries": "query.sql",
    "engine": "postgresql",
    "gen": {
      "go": {
        "out": "db",
        "sql_package": "pgx/v5"
      }
    }
  }]
}
```

### Playground URL

https://play.sqlc.dev/p/7ac654a59b8d17aa0f7e9bd8dfd689bec9d035bc8e8ea7a5a1beb078be531769

### What database engines are you using?

PostgreSQL

### What type of code are you generating?

Go

benjaco:
I would guess that the not null checks here should be moved up as the first check
https://github.com/sqlc-dev/sqlc/blob/ec9d492c40168739f813a0dc75f4ccd601770573/internal/codegen/golang/postgresql_type.go#L236-L246

As it is with the string types
https://github.com/sqlc-dev/sqlc/blob/ec9d492c40168739f813a0dc75f4ccd601770573/internal/codegen/golang/postgresql_type.go#L260-L271

eashan-dirac:
Facing this same issue. This is a pretty urgent bug considering nullable timestamptz's are relatively common in Postgres and having them be represented as nil in Go is very idiomatic.

benjaco:
@mcdoker18 I can see you authored the original part of this code in #1874, do you have any comments about the desired intent here? Great job by the way!

[...]

PR #3838, opened 2025-02-12
Update postgresql_type.go

theAnuragMishra:
Check notNull and emitPointersForNull before checking postgres driver


benjaco:
Hi, it looked like you took a stab at #3837 thank you for giving it a shot

Why was this closed? I can see there is a lot of tests there fails, the question should then be if the new output is the desired or not? Is it to big of a breaking change? should it be behind a new flag?

theAnuragMishra:
Hey @benjaco !! I didn't want to just steal your idea. I had the same problem (time.Time instead of *time.Time being emitted). So I did a Google search that took me to your issue. I had done this kind of a pr before for sqlc. So I decided to do another. 
Closed this one for another that I opened with a patch branch.

PR #3854, opened 2025-02-20
update csharp support to stable

doron050:
Hi,
We have implemented all supported query annotations and macros for all three databases, and we have tests for most of the generated code features.

We think our plugin is stable now 💯  and would like to update the docs


kyleconroy:
Excellent!

PR #3876, opened 2025-03-09
feat(codegen/golang): add an option to wrap query errors that includes query name

andrewmbenton:
This is an attempt to add wrapped errors to query functions in order to get the query name, as described in https://github.com/sqlc-dev/sqlc/issues/1612.

If this approach looks reasonable I'll add more tests. But I'll also have to decide whether to wrap errors coming from the more esoteric function types like batch and copyfrom.


kyleconroy:
I'd change the error message to be shorter. Here's the error message from [os.ReadFile if a file doesn't exist](https://go.dev/play/p/I9GcD4drm_Y)

```
open hello: no such file or directory
```

so I'd change the error message to something like

```
query {{MethodName}}: %w
```

I also know that no matter what format we choose, people will want to change it.

kyleconroy:
> But I'll also have to decide whether to wrap errors coming from the more esoteric function types like batch and copyfrom.

Yeah, I'd be okay not wrapping errors from those functions simply because they have a larger surface area

andrewmbenton:
Re error messages, agree people will want to change it and we should go as short as possible. I was mimicking what we do in `dbCode` e.g.

```go
fmt.Errorf("error preparing query {{.MethodName}}: %w", err)
```

and

```go
fmt.Errorf("error closing {{.FieldName}}: %w", cerr)
```

arguably those could be shorter as well though.

andrewmbenton:
Let me know if you think we need more tests, e.g. for configurations that branch in the templates in ways I might have missed.

kyleconroy (approved):

PR #3878, opened 2025-03-11
allow renaming function Params structs

skandragon:
I have a case where I want to wrap some of the database calls in some internal layers, and would like to have a single params structure that is exposed between my hand-rolled wrapped code and the generated names.

Specifically, I am making a function called `UpsertFooUncached` using code gen and I want a cached version:

```go
func (q *Queries) UpsertFoo(ctx context.Context, args UpsertFooParams) (uuid.UUID, error) {
  // caching would go here
  id, err := q.UpsertFooUncached(ctx, args)
  if err != nil {
    return uuid.Nil, err
  }
  // cache the result
  return id, nil
}
```

In this case, I could use `type UpsertFooParams UpsertFooUncachedParams` but this doesn't work well for users as it doesn't always expand with VSCode, so the internals are a mystery.  I could recreate the structure, but then I need to keep it in sync.  I could embed the Uncached version, but then using it is a pain.

I'd like to rename, and this change seems to do the trick.

### DIFF
--- a/docs/howto/rename.md
+++ b/docs/howto/rename.md
@@ -21,7 +21,7 @@ sql:
   queries: "postgresql/query.sql"
   engine: "postgresql"
   gen:
-    go: 
+    go:
       package: "authors"
       out: "postgresql"
       rename:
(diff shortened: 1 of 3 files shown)


skandragon:
Any chance this can be merged, or feedback on what it would take?

skandragon:
bump.

skandragon:
@kyleconroy 

It has been over a year; can I get a rejection or a merge if this is acceptable?

PR #3919, opened 2025-04-03
fix(migrations): case-sensitive migration rollback detection

0xSherlokMo:
This PR fixes an issue where SQLC was failing to properly detect and remove rollback statements in migration files due to case sensitivity. Previously, SQLC only recognized uppercase `Down` in rollback statements (e.g., `-- +goose Down`), but migration tools like Goose have updated their [documentation](https://pressly.github.io/goose/documentation/annotations/) to use lowercase `down` (e.g., `-- +goose down`).

The change makes the rollback statement detection case-insensitive by converting the statement to lowercase before checking, ensuring compatibility with:
- Both `-- +goose Down` and `-- +goose down` (Goose migrations)
- Both `-- +migrate Down` and `-- +migrate down` (sql-migrate)
- Case variations in other migration tool formats

This maintains backward compatibility with existing migrations while supporting the current documentation standards of migration tools.

Fixes issue #3444

## Changes
- Modified `RemoveRollbackStatements` to perform case-insensitive comparison by converting statements to lowercase before checking
- Maintains support for all existing migration tool formats:
  - Goose
  - sql-migrate
  - tern
  - dbmate

## Testing
- Verified that both uppercase and lowercase rollback statements are properly detected and removed
- Ensures no breaking changes for existing migration files

### DIFF
--- a/internal/migrations/migrations.go
+++ b/internal/migrations/migrations.go
@@ -15,16 +15,17 @@ func RemoveRollbackStatements(contents string) string {
 	s := bufio.NewScanner(strings.NewReader(contents))
 	var lines []string
 	for s.Scan() {
-		if strings.HasPrefix(s.Text(), "-- +goose Down") {
+		statement := strings.ToLower(s.Text())
+		if strings.HasPrefix(statement, "-- +goose down") {
 			break
 		}
-		if strings.HasPrefix(s.Text(), "-- +migrate Down") {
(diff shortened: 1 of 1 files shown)


FadyGamilH (approved):

kyleconroy:
@0xSherlokMo appreciate the simple change and the excellent explanation.

PR #3962, opened 2025-05-14
Fix `ORDER BY` arg parsing in sqlite

### LINKED ISSUE #3788: Args in sqlite ORDER BY not respected
### Version

1.27.0

### What happened?

Similarly to https://github.com/sqlc-dev/sqlc/issues/3113
The postgres generator supports variables in the ORDER BY, but sqlite does not.

```sql
CREATE TABLE authors (
  id int PRIMARY KEY
);
```
```sql
-- name: GetAuthor :one
SELECT * FROM authors
WHERE id = @id
ORDER BY @order_by;
```

```
const getAuthor = `-- name: GetAuthor :one
SELECT id FROM authors
WHERE id = ?1
ORDER BY @order_by
`

func (q *Queries) GetAuthor(ctx context.Context, id int64) (int64, error) {
	row := q.db.QueryRowContext(ctx, getAuthor, id)
	err := row.Scan(&id)
	return id, err
}
```
Playground link is to 1.25.0, but this is happening locally on 1.27.0 as well. Changing only the engine value to `postgresql` results in the expected output for postgresql. But the same is not reflected in sqlite.

### Relevant log output

```shell

```

### Database schema

```sql
CREATE TABLE authors (
  id int PRIMARY KEY
);
```

### SQL queries

```sql
-- name: GetAuthor :one
SELECT * FROM authors
WHERE id = @id
ORDER BY @order_by;
```

### Configuration

```yaml
{
  "version": "2",
  "sql": [{
    "schema": "schema.sql",
    "queries": "query.sql",
    "engine": "sqlite",
    "gen": {
      "go": {
        "out": "db"
      }
    }
  }]
}
```

### Playground URL

https://play.sqlc.dev/p/a0947e011d5ec187db7c502d1044c9eb10d6a2f9f5c6d44f166b23a2fa6c5f07

macOS

SQLite

Go

andriygm:
`sqlc` currently doesn't handle bound `ORDER BY` clauses in sqlite, nor does it have tests for this behavior. This PR attempts for rectify this. Closes https://github.com/sqlc-dev/sqlc/issues/3788.


andriygm:
currently, base behavior works but mixing params like so:
```sql
-- name: ListAuthorsColumnSortDirection :many
SELECT * FROM authors
WHERE id > ?
ORDER BY
    CASE
        WHEN @order_by = 'asc' THEN name
    END ASC,
    CASE
        WHEN @order_by = 'desc' OR @order_by IS NULL THEN name
    END DESC;
```
produces this incorrect output (question marks in CASE clauses should instead be same variable)
```sql
-- name: ListAuthorsColumnSortDirection :many
SELECT id, name, bio FROM authors
WHERE id > ?
ORDER BY
    CASE
        WHEN ? = 'asc' THEN name
    END ASC,
    CASE
        WHEN ? = 'desc' OR ? IS NULL THEN name
    END DESC
```
this can still be worked around by using named params in this case.

ilikeorangutans:
How can we get this PR merged? I'm currently running into this issue :(

kyleconroy:
**🤖 Automated message**

---

PR #3968, opened 2025-05-16
SQLite: Coerce jsonb columns to json before returning to Go code

brandur:
This one follows up the discussion in #3953 to try and make the `jsonb`
data type in SQLite usable (see discussion there, but I believe that
it's currently not).

According the SQLite docs on JSONB [1], it's considered a format that's
internal to the database itself, and no attempt should be made to parse
it elsewhere:

> JSONB is not intended as an external format to be used by
> applications. JSONB is designed for internal use by SQLite only.
> Programmers do not need to understand the JSONB format in order to use
> it effectively. Applications should access JSONB only through the JSON
> SQL functions, not by looking at individual bytes of the BLOB.

Currently, when trying to use a `jsonb` column in SQLite, sqlc ends up
returning the internal binary data, which ends up being unparsable in Go:

    riverdrivertest.go:3030:
                Error Trace:    /Users/brandur/Documents/projects/river/internal/riverinternaltest/riverdrivertest/riverdrivertest.go:3030
                Error:          Not equal:
                                expected: []byte{0x7b, 0x22, 0x66, 0x6f, 0x6f, 0x22, 0x3a, 0x20, 0x22, 0x62, 0x61, 0x72, 0x22, 0x7d}
                                actual  : []byte{0x8c, 0x37, 0x66, 0x6f, 0x6f, 0x37, 0x62, 0x61, 0x72}

                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1,3 +1,3 @@
                                -([]uint8) (len=14) {
                                - 00000000  7b 22 66 6f 6f 22 3a 20  22 62 61 72 22 7d        |{"foo": "bar"}|
                                +([]uint8) (len=9) {
                                + 00000000  8c 37 66 6f 6f 37 62 61  72                       |.7foo7bar|
                                 }
                Test:           TestDriverRiverSQLite/QueueCreateOrSetUpdatedAt/InsertsANewQueueWithDefaultUpdatedAt

The fix is that we should make sure to coerce `jsonb` columns back to
`json` before returning. That's what this pull request does, intercepting
`SELECT *` and wrapping `jsonb` columns with a `json(...)` invocation.

I also assign `json` and `jsonb` a `[]byte` data type by default so they
don't end up as `any`, which isn't very useful. `[]byte` is consistent
with the default for `pgx/v5`.

[1] https://sqlite.org/jsonb.html


internal/codegen/golang/sqlite_type.go
```diff
@@ -56,6 +56,9 @@ func sqliteType(req *plugin.GenerateRequest, options *opts.Options, col *plugin.
 		}
 		return "sql.NullTime"
 
+	case "json", "jsonb":
+		return "[]byte"
```
brandur, line 60: Is `json.RawMessage` preferred over `[]byte`?

internal/compiler/expand.go
```diff
@@ -149,6 +150,9 @@ func (c *Compiler) expandStmt(qc *QueryCatalog, raw *ast.RawStmt, node ast.Node)
 				if counts[cname] > 1 {
 					cname = tableName + "." + cname
 				}
+				if _, ok := c.parser.(*sqlite.Parser); ok && column.DataType == "jsonb" {
```
brandur, line 153: So I left the `*sqlite.Parser` in here as a dummy implementation for detecting when we're running as SQLite, but I could use some help determining what the preferred convention should be. It feels a little heavy handed to make it part of one of the interfaces like `Parser` given it's such an unusual corner case.

brandur:
@kyleconroy I'm sure this could use some work, but I wanted to pitch something to prove out the concept. Thoughts?

internal/codegen/golang/sqlite_type.go
kyleconroy: Yeah, it's what we do for PostgreSQL.

internal/compiler/expand.go
kyleconroy: I looked through the code to see if there was a better place to put this. The ideal solution is to create a `SelectExpr` function on the compiler.Column which adds the `json()`. That said, we'd have to pipe it through the catalog and the ast, all of which are engine-agnostic.

Instead, let's create a new `Selector` interface with a single method `Expr(name, column) string`. In NewCompiler we already switch on the engine so we can pick the correct one to use. Unlike the parser and the catalog, the selector structs shouldn't be exported and should live in the compiler package.

What do you think?

internal/codegen/golang/sqlite_type.go
brandur: Cool, that makes sense. Changed!

kyleconroy (approved):

brandur:
@kyleconroy Thx! One tweak I made is to make the selector interface internal to the `compiler` package as requested in your original comment.

In terms of review: is there a particular process I should know about/follow? i.e. Do you go through all open PRs with lgtms periodically and merge them? Do I need to do anything?

PR #3980, opened 2025-06-03
Avoid updating target file mtime in 'generate'

n-oden:
If the target file exists on disk already, compare its contents to the generated source in memory: only update the target file on disk if the contents differ or if the target file is missing.

This should allow build tools downstream of sqlc to make conditional decisions based on whether the generated code files have been updated (e.g. mockgen).

### DIFF
--- a/internal/cmd/cmd.go
+++ b/internal/cmd/cmd.go
@@ -204,6 +204,15 @@ var genCmd = &cobra.Command{
 		}
 		defer trace.StartRegion(cmd.Context(), "writefiles").End()
 		for filename, source := range output {
+			different, err := diffFile(filename, []byte(source))
+			if err != nil {
+				fmt.Fprintf(stderr, "%s: %s\n", filename, err)
+				return err
+			}
+			if !different {
+				// if the file is the same, we can skip writing it
+				continue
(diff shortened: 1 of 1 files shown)


n-oden:
@kyleconroy small nudge here: is this a change that you're okay with?

andrewmbenton:
Seems like a good idea and pretty cheap to do. I'll hold off on a review until @kyleconroy weighs in.

n-oden:
@kyleconroy gentle ping? :)

n-oden:
@andrewmbenton @kyleconroy Sorry for being a nuisance, but I figured I'd check in once more before the end of the year: is there any chance we can get this merged? (Of course if you have feedback I'm all ears!)

jwarlander:
Also curious about this, as I'd like to have my "make build" target generate the sqlc output just like it does for Templ & our Tailwind CSS output..

Currently though, if I use that build target with `air` to rebuild on changes when I'm testing things locally, it just keeps looping as it thinks the sqlc output files were modified.

n-oden:
It's been a year; I am closing this since I assume it is not wanted.

PR #4005, opened 2025-06-29
feat: Introduce sqlc.optional for dynamic query generation

copito:
This commit introduces the `sqlc.optional` feature, allowing conditional inclusion of SQL query fragments at runtime.

Key changes:

1.  **Parser Enhancement**: The SQL parser now recognizes `sqlc.optional('ConditionKey', 'SQLFragment')` syntax within query files. This information is stored in the query's metadata.

2.  **Code Generation**:
    - Go code generation logic has been updated to process these `OptionalBlocks`.
    - Generated Go functions now include new parameters (typed as `interface{}`) corresponding to each `ConditionKey`.
    - Templates (`stdlib/queryCode.tmpl`, `pgx/queryCode.tmpl`) were modified to dynamically build the SQL query string and its arguments at runtime. If an optional Go parameter is non-nil, its associated SQL fragment is included in the final query, and its value is added to the list of database arguments.

3.  **Parameter Handling**: `$N` placeholders in all SQL fragments (base or optional) consistently refer to the Nth parameter in the generated Go function's signature.

4.  **Documentation**: Added comprehensive documentation for `sqlc.optional` in `docs/reference/query-annotations.md`, covering syntax, behavior, parameter numbering, and examples.

5.  **Examples**: A new runnable example has been added to `examples/dynamic_query/postgresql/` to demonstrate practical usage.

6.  **Tests**: New end-to-end tests were added in `internal/endtoend/testdata/dynamic_query/` for both `stdlib` and `pgx` drivers, ensuring the correctness of the generated code.

### DIFF
--- a/docs/reference/query-annotations.md
+++ b/docs/reference/query-annotations.md
@@ -113,6 +113,180 @@ func (q *Queries) GetAuthor(ctx context.Context, id int64) (Author, error) {
 }
 ```
 
+## Conditional SQL with `sqlc.optional`
+
+The `sqlc.optional` annotation allows for parts of a SQL query to be conditionally included at runtime. This is useful for building queries with optional filters or other dynamic components.
+
+### Purpose
+
+`sqlc.optional` provides a way to construct dynamic SQL queries where certain SQL fragments are only appended to the base query if a corresponding Go parameter is non-`nil`. This avoids the need for complex string manipulation or multiple similar queries for different filtering scenarios.
+
+### Syntax
+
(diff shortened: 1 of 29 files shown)


ororsatti:
Just a comment, but maybe the optional types can be something like NullString ? 

```go
type xxxOptional struct {
Xxx *<type>
Valid bool
}
```

This way it won't have to be interface{} when ever you wanna use a optional value.

kyleconroy:
Really appreciate you taking a crack at this! You're heading in the right direction, but I don't think this exact syntax is what we're going to adopt. Sorry I don't have more specific feedback, but hoping to get more time in the next few months to get to a design that I'm happy with.

In the future, please start a discussion first so that you don't implement a full feature I end up not merging.

PR #4006, opened 2025-06-30
fix(compiler): INSERT-WITH-SELECT statements

dennisgsmith:
Fixes https://github.com/sqlc-dev/sqlc/issues/3847
Fixes https://github.com/sqlc-dev/sqlc/issues/3996

I was bumping into the above issues (but for postgres) and this bug seemed to be engine agnostic. After inspecting the output with `SQLCDEBUG=dumpast=1` and `SQLCDEBUG=dumpcatalog=1` I found that:

1. The CTE relation name was not in the catalog when using insert-cte-select queries.
2. The AST `WithClause` was under the `SelectStmt`, not directly within the parent `InsertStmt` node.

Using the `SelectStmt` `WithClause` when it's not null resolves the issue. Please let me know if there are any changes you'd like to see. Thanks!


dennisgsmith:
@kyleconroy could you please review this PR?

Issue #1389, opened 2022-01-24
Support for CRDB and other Postgres features


danthegoodman1:
When trying to use sqlc for our project I noticed a few things were not supported that we require. We do use CockroachDB with the postgres engine.

1. The [`ON UPDATE`](https://www.cockroachlabs.com/docs/v21.2/add-column#add-a-column-with-an-on-update-expression) expression caused SQL validation errors. This is specific to CRDB.
2. [Hash sharded indexes](https://www.cockroachlabs.com/docs/stable/hash-sharded-indexes.html) caused SQL validation errors as well
3. Composite primary keys with ordering like `PRIMARY KEY(col1, col2 DESC)` said that `DESC` was an error

ajwerner:
I feel like you don't need to tell sqlc about the exact schema because it doesn't ultimately manage the schema. All of the features you just mentioned are schema features and not query features. Like, you can imagine telling sqlc about a table that doesn't use any of that syntax and then all the queries should still work, right?

danthegoodman1:
@ajwerner That is true, but now I need to have SQL generating from a file that is not the true SQL of our schema, which means a second place to manage SQL and manually ensuring that it is consistent with our true SQL schema (which could lead to errors, which sqlc is designed to avoid)

For example we have the following primary key:
```sql
PRIMARY KEY(player_id, id DESC) 
```

sqlc does not like the `DESC` part of it (considers invalid SQL, which is weird because this works in postgres too...)

Removing the `DESC` would now be an incorrect version of the table.

Another example is this column:

```sql
updated_at TIMESTAMPTZ NOT NULL DEFAULT NOW() ON UPDATE NOW(), -- auto update the timestamp
```

We would have to maintain another copy of that that that removes the `ON UPDATE`.

I think the risk of losing consistency between the SQL schema and the actual schema used is not worth the extra time writing the SQL wrapper functions manually in this case. Since we always have to update the code in 2 places: Either the schema and the sqlc compatible schema, or the schema and the functions that use the affected parts of the schema.

TL;DR I believe you make an excellent point, but I am still considering the balance of maintaining 2 schemas vs a schema and hand written functions.

danthegoodman1:
In reality this isn't something we should be constantly iterating on anyway... so maybe that leans more in favor of doing the second fake sqlc schema...

danthegoodman1:
[...]

PR #4009, opened 2025-07-01
Adding support for CockroachDB

Exar04:
Fixes : #1389 
- [x] Implement basic type support beyond text
- [x] Implement expansion of  *

will be adding more soon


kyleconroy:
@Exar04 I really appreciate the pull requests, but I'm not looking to add support for new engines right now. There are some internal changes I'd like to make first which should make adding new engines much easier.

PR #4012, opened 2025-07-07
[RFC] Added Gleam Community Project to docs - language-support.rst

LarsArtmann:
I thought https://github.com/daniellionel01/parrot/  would be a good addtion to the docs.
But, I am actually not sure if https://github.com/daniellionel01/parrot/ qualifies as a sqlc plugin.


daniellionel01:
Hi, author of Parrot here!

The referenced library is not strictly usable as a sqlc plugin. It makes use of a `sqlc-gen-json` plugin, which it parses and turns into gleam code, however not through the protobuf mechanisms that other plugins do.

If this disqualifies parrot from being listed as a plugin for the gleam programming language, no hard feelings 😃

kyleconroy:
I'd love to add it as a community project, but I don't think it fits into the current section. Can you add another table below that has the same structure but just replace Plugin with Project?

Issue #3852, opened 2025-02-20
Relation does not exist on update only


dimmerz92:
### Version

1.28.0

### What happened?

Expected `sqlc generate` to generate all queries.

I tried removing queries one by one and found that generating worked for everything except for `UPDATE` statements.

No issues in sqlc playground.

### Relevant log output

```shell
# package models
sql/queries/users.sql:45:1: relation "Users" does not exist
sql/queries/users.sql:53:1: relation "Users" does not exist
```

### Database schema

```sql
-- +goose Up
CREATE TABLE Users (
	ID INTEGER PRIMARY KEY,
	FullName TEXT NOT NULL,
	Email TEXT UNIQUE NOT NULL CHECK (Email LIKE '%_@%_.%__'),
	Password BLOB NOT NULL,
	CreatedAt INTEGER NOT NULL DEFAULT (unixepoch('now')),
	UpdatedAt INTEGER
);

-- +goose StatementBegin
CREATE TRIGGER UpdateUser
AFTER UPDATE ON Users
BEGIN
	UPDATE
		Users
	SET	
		UpdatedAt = unixepoch('now')
	WHERE
		ID = NEW.ID;
END;
-- +goose StatementEnd

-- +goose Down
DROP TABLE Users;
```

### SQL queries

```sql
-- name: AddUser :one
INSERT INTO Users (
	FullName,
	Email,
	Password
)
VALUES (
	@FullName,
	@Email,
	@Password
)
RETURNING
	*;

-- name: CheckEmailExists :one
SELECT
	1
FROM
	Users
WHERE
	Email = @email;

-- name: GetUser :one
SELECT
	*
FROM
	Users
WHERE
	ID = @userID OR
	Email = @email;

-- name: GetPaginatedUsers :many
SELECT
	*
FROM
	Users
ORDER BY
	ID
LIMIT
	@limit
OFFSET
	@offset;

-- name: UpdateEmail :exec
UPDATE
	Users
SET
	Email = @email
WHERE
	ID = @userID;

-- name: UpdatePassword :exec
UPDATE
	Users
SET
	Password = @password
WHERE
	ID = @userID;

-- name: DeleteUser :exec
DELETE FROM
	Users
WHERE
	ID = @userID;
```

### Configuration

```yaml
version: 2
sql:
  - engine: sqlite
    queries: sql/queries
    schema: sql/schema
    gen:
      go:
        package: models
        out: models
```

### Playground URL

[...]

PR #4028, opened 2025-07-23
Fix SQLite Update query generation

compiledpanda:
# Purpose
Resolve the issue described in #3852 where update statements fail when generating queries. Note that this regression was introduced in 1.28.0 and is currently affecting the current latest release (1.29.0).

# Background
In commit 58b08ae the table names in sqlite queries were normalized (i.e. lower cased). In that commit a number of changes were made in the file `internal/engine/sqlite/convert.go` to lowercase tables names in during query parsing so that the table names would always match the table names pulled from the schema even if the casing is different. The linked issue above exists because the table name in update statements was not included in the linked set of changes last year. In essence, if you used an uppercase table name in your update query (i.e. `UPDATE MY_TABLE ...`) it will not match the parsed table name which is currently being converted to lowercase (i.e. `my_table`).

# Changes
- Call the `identifier()` function in `convertUpdate_stmtContext()` so that tableName is properly cased before it is returned and used.

### DIFF
--- a/internal/engine/sqlite/convert.go
+++ b/internal/engine/sqlite/convert.go
@@ -1035,7 +1035,7 @@ func (c *cc) convertUpdate_stmtContext(n Update_stmt) ast.Node {
 	}
 
 	relations := &ast.List{}
-	tableName := n.Qualified_table_name().GetText()
+	tableName := identifier(n.Qualified_table_name().GetText())
 	rel := ast.RangeVar{
 		Relname:  &tableName,
 		Location: n.GetStart().GetStart(),


kyleconroy:
Could you please add a test for this? I want to make sure we catch this in the future.

kyleconroy (changes requested):
Can you add a test for this? It's helpful so that we don't break this in the future.

compiledpanda:
@kyleconroy I added an e2e test for uppercase tables/relations in sqlite. The tests fail when you revert the identifier change in `internal/engine/sqlite/convert.go` and succeed with the change present.

PR #4033, opened 2025-07-30
Enhance package error message for gen go

rodrigoherera:
### Description
I enhance the gen go error output to give a hint on what could lead to the error

### tests
tested manually


kyleconroy:
I'd like to keep the error messages the same for right now. I have larger plans for error messages from plugins in the future.

PR #4035, opened 2025-08-02
docs: add FOR UPDATE clause to MySQL transaction example

### LINKED ISSUE #3485: Error in the documentation for transactions?
### Version

1.26.0

### What happened?

https://docs.sqlc.dev/en/latest/howto/transactions.html

In the given example, the code tries to select a row from records and increase the counter by 1 (in a transaction). According to the official MySQL manual:

> If you query data and then insert or update related data within the same transaction, the regular SELECT statement does not give enough protection. Other transactions can update or delete the same rows you just queried. InnoDB supports two types of [locking reads](https://dev.mysql.com/doc/refman/8.4/en/glossary.html#glos_locking_read) that offer extra safety - https://dev.mysql.com/doc/refman/8.4/en/innodb-locking-reads.html

So, shouldn't

```
-- name: GetRecord :one
SELECT * FROM records
WHERE id = $1;
```

be

```
-- name: GetRecord :one
SELECT * FROM records
WHERE id = $1
FOR UPDATE;
```

in this case?

kotahorii:
Fixes #3485

## Summary

Adds the `FOR UPDATE` clause to the MySQL transaction example to prevent race conditions when reading and then updating data within the same transaction.

## Changes Made

1. **Modified GetRecord query** to include `FOR UPDATE` clause:
   ```sql
   -- Before
   SELECT * FROM records WHERE id = $1;
   
   -- After  
   SELECT * FROM records WHERE id = $1 FOR UPDATE;
   ```

2. **Added explanatory section** about locking reads in transactions to help users understand when and why to use `SELECT ... FOR UPDATE`

## Problem Addressed

The current transaction example shows a read-then-update pattern without proper row locking, which can lead to race conditions in MySQL where other transactions might modify the same rows between the SELECT and UPDATE operations.

According to the MySQL documentation:
> "If you query data and then insert or update related data within the same transaction, the regular SELECT statement does not give enough protection. Other transactions can update or delete the same rows you just queried."

## Solution

The `FOR UPDATE` clause provides exclusive row locking in MySQL/InnoDB, preventing other transactions from modifying the selected rows until the current transaction completes.

## References

- [MySQL InnoDB Locking Reads Documentation](https://dev.mysql.com/doc/refman/8.4/en/innodb-locking-reads.html)
- Issue #3485

🤖 Generated with [Claude Code](https://claude.ai/code)


kyleconroy:
The sqlc documentation isn't the correct place for database-specific transaction documentation. When and where to use `FOR UPDATE` is a large topic better left to MySQL-specific documentation.

Issue #3936, opened 2025-04-19
Casing issue with DROP TABLE statement ...


TorrinLynn:
### Version

1.29.0

### What happened?

If schema.sql contains a mixed case table name, you can only drop the table using lowercase name, otherwise, you get a relation `"<table name>" does not exist`. The following schema.sql illustrates the problem:

```
CREATE TABLE Authors (
  id   INTEGER PRIMARY KEY,
  name text    NOT NULL
);

DROP TABLE Authors; -- generates relation "Authors" does not exist error
```

Changing the drop statement to `DROP TABLE authors` fixes the reported error.

The bug first appeared in 1.28.0 and continues to exist in 1.29.0

### Relevant log output

```shell
# package tutorial
schema.sql:1:1: relation "Authors" does not exist
schema.sql:1:1: relation "authors" already exists
```

### Database schema

```sql
CREATE TABLE Authors (
  id   INTEGER PRIMARY KEY,
  name text    NOT NULL
);

DROP TABLE Authors;

CREATE TABLE Authors (
  id   INTEGER PRIMARY KEY,
  name text    NOT NULL,
  bio  text
);
```

### SQL queries

```sql

```

### Configuration

```yaml
version: "2"
sql:
  - engine: "sqlite"
    queries: "query.sql"
    schema: "schema.sql"
    gen:
      go:
        package: "tutorial"
        out: "tutorial"
```

### What operating system are you using?

macOS

### What database engines are you using?

SQLite

### What type of code are you generating?

Go

TorrinLynn:
Also found that capitalizing the table name for update queries causes an error:

```
-- name: UpdateAuthor :exec
UPDATE Authors
set name = ?,
bio = ?
WHERE id = ?;
```

Generates the following error:
`query.sql:18:1: relation "Authors" does not exist`

SELECT, INSERT, and DELETE statements don't seem to care about table name casing.

TorrinLynn:
Using Alter Table with rename is also case sensitive. 

```
CREATE TABLE authors (
  id   INTEGER PRIMARY KEY,
  name text    NOT NULL
);

DROP TABLE authors;

CREATE TABLE authors_new (
  id   INTEGER PRIMARY KEY,
  name text    NOT NULL,
  bio  text
);

ALTER TABLE authors_new
    RENAME TO Authors;
```

```
-- name: GetAuthor :one
SELECT * FROM authors
WHERE id = ? LIMIT 1;

-- name: ListAuthors :many
[...]

PR #4045, opened 2025-08-07
bugfix: Normalize identifier usage for table names

Mr-Destructive:
Addressing issue #3936 

- Use identifier to get the table names consistent across all insert, update and other statements

Do you want to add tests for these? Thought would be a very specific test, so didn't add.
Let me know about the thoughts for this
Thanks


andrewmbenton:
Tests would be ideal. @kyleconroy may feel differently but I think because we already have an endtoend test for drop table statements it would be reasonable to just add a couple of statements to that test to cover this case.

As for the other failure cases, those are a bit more complicated and would require their own endtoend tests.

PR #4067, opened 2025-08-19
docs: add multi-worded table example for renaming

lukeraphael:
This update improves documentation for struct renaming in sqlc.yaml. 

- Adds an example showing how to handle multi-word table names, where underscores must be included (e.g., book_publisher → Publisher).


kyleconroy:
Thanks for the docs update!

PR #4082, opened 2025-08-29
fix(compiler): prevent schema parse failures by ignoring psql meta commands

andrewmbenton:
This is kind of a lazy quick-and-dirty implementation that doesn't attempt to consolidate some duplicate line parsing that we do during migration removal and just sticks a func right in the package that handles file parsing for all engines, even though this line filter only applies to postgresql.

Happy to revise if there's a better place to put this.

Resolves https://github.com/sqlc-dev/sqlc/issues/4065

### DIFF
--- a/internal/compiler/compile.go
+++ b/internal/compiler/compile.go
@@ -1,6 +1,7 @@
 package compiler
 
 import (
+	"bufio"
 	"errors"
 	"fmt"
 	"io"
@@ -38,6 +39,7 @@ func (c *Compiler) parseCatalog(schemas []string) error {
 			continue
 		}
 		contents := migrations.RemoveRollbackStatements(string(blob))
+		contents = removePsqlMetaCommands(contents)
 		c.schema = append(c.schema, contents)
 		stmts, err := c.parser.Parse(strings.NewReader(contents))
 		if err != nil {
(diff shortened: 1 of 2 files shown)


atombender:
Can this please be merged? It's a blocking issue with Postgres 17. @kyleconroy

internal/compiler/compile.go
```diff
@@ -57,6 +59,19 @@ func (c *Compiler) parseCatalog(schemas []string) error {
 	return nil
 }
 
+func removePsqlMetaCommands(contents string) string {
+	s := bufio.NewScanner(strings.NewReader(contents))
+	var lines []string
+	for s.Scan() {
+		line := s.Text()
+		if strings.HasPrefix(line, `\`) {
```
gbarr, line 67: This seems a little aggressive to remove all lines that begin with `\`. The following is valid SQL, but would be broken by this fix

```
SELECT E'
\x61 b c
';
```

I don't think `pg_dump` would create such a statement but hand written SQL could.

I would suggest that the fix only exclude lines starting with `\restrict` and `\unrestrict`

j-houston:
Bump. Can we get this merged?

ignatremizov:
I added https://github.com/sqlc-dev/sqlc/pull/4177 to address @gbarr comment and not to break any other existing postgresql workflows (i.e., on `/connect` or other psql meta commands)

kyleconroy:
Fixed via https://github.com/sqlc-dev/sqlc/pull/4390

PR #4103, opened 2025-09-12
fix(postgresql): Support selecting from a function

wickwirew:
Fixes an issue in postgres when selecting from a function. It would silently fail and the output row type would just be a `NullString`. This was due to it not looking for a composite type from the return type. Also it was not generating structs for defined composite types.

Sort of addresses the following issues, both were closed but I dont think officially fixed, but this fix requires a composite return type:
https://github.com/sqlc-dev/sqlc/issues/3638
https://github.com/sqlc-dev/sqlc/issues/1322

Supporting `RETURNS TABLE(...)` may be a bit more complex. I don't think libpg_query even returns the data, [I don't see it at least](https://github.com/pganalyze/libpg_query/blob/9ac12d29f2ffc0e49de8f4239223d50aaccd1549/src/postgres/include/nodes/parsenodes.h#L3427) but I could be wrong. If it is going to be a known unsupported feature, I can emit a diagnostic saying to use a composite type at the call site if desired.

Supporting embedded structs is a bit odd, but possible. The function output needs to be aliased to the composite type name
❌ Does not work: `SELECT sqlc.embed(function_name) FROM function_name()`
✅ Works: `SELECT sqlc.embed(type_name) FROM function_name() AS type_name`


frozenjava:
Would love to see this merged!

kyleconroy:
**🤖 Automated message**

---

PR #4129, opened 2025-10-07
cross schema inherits

axlalmflower:
Fixes issues with inherit implementation which makes cross schema inherits not work. Fix is to ensure that we look for the inherited table in the inherited table's schema and not in the schema of the table that inherits.


axlalmflower:
(prob done something wrong setting up the test cases..)

axlalmflower:
issue: https://github.com/sqlc-dev/sqlc/issues/4128

PR #4161, opened 2025-10-30
feat(plugin): add configuration to run go package

frankli0324:
This PR adds "go_package" field under "plugin.process" for configurations
I'm trying to write plugin for a specific project, so I think it's reasonable to put plugin code alongside the project, and currently we need to install the plugin into $PATH before running sqlc generate, this simplifies the process.
I suspect the "cmd" field checks for existence for some reason, so I didn't add a "args" field which may be more universal but less maintainable. I'm not sure about that.

### DIFF
--- a/internal/cmd/generate.go
+++ b/internal/cmd/generate.go
@@ -349,6 +349,7 @@ func codegen(ctx context.Context, combo config.CombinedSettings, sql OutputPair,
 		switch {
 		case plug.Process != nil:
 			handler = &process.Runner{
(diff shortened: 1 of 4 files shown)


frankli0324:
I've being using this for a while and I didn't run into problems. rebased and squashed. any thoughts on this? @kyleconroy

frankli0324:
hi @kyleconroy should I or should I not assume that I can use this option later? I'm trying to decide the workflow of my project...

PR #4171, opened 2025-11-06
sqlite: support CTEs for INSERT / UPDATE / DELETE

PhilipTrauner:
CTE handling differs slighlty between `SELECT` and other kinds of statements, as `SELECT` grammar uses a
`common_table_stmt` token, while other DML statements use a `with_clause` token


PhilipTrauner:
inadvertently opened here, wanted to open it [here](https://github.com/OHF-Device-Database/sqlc) instead

PR #4172, opened 2025-11-07
Fix MySQL Mathematical Expression Type Inference

### LINKED ISSUE #4153: How can i get correct type
I want to generate `sql.NullFloat64` but got` interface{}`. How can I get the correct type?

Here is my `query.sql` and `schema.sql`.
query.sql
```sql
-- name: ListTest :many
SELECT 
    (a1 / 1024) a1_float, (a2 / 1024) a2_float, a3
FROM test;
```

scheam.sql
```sql
CREATE TABLE test (
    a1 float NULL,
    a2 float NULL,
    a3 float NULL
);
```

The generated code look like this:
```go
const listTest = `-- name: ListTest :many
SELECT 
    (a1 / 1024) a1_float, (a2 / 1024) a2_float, a3
FROM test
`

type ListTestRow struct {
	A1Float interface{}
	A2Float interface{}
	A3      sql.NullFloat64
}

func (q *Queries) ListTest(ctx context.Context) ([]ListTestRow, error) {
	rows, err := q.db.QueryContext(ctx, listTest)
	if err != nil {
		return nil, err
	}
	defer rows.Close()
	var items []ListTestRow
	for rows.Next() {
		var i ListTestRow
		if err := rows.Scan(&i.A1Float, &i.A2Float, &i.A3); err != nil {
			return nil, err
		}
		items = append(items, i)
	}
		return nil, err
	}
		return nil, err
	}
	return items, nil
}
```

The fields `A1Float` and `A2Float` in the struct `ListTestRow` are of type `interface{}`.

rubensantoniorosa2704:
This implements type inference for MySQL mathematical expressions. Before this change, queries with arithmetic operations would generate `interface{}` types, which meant no compile-time type safety. Now the compiler properly infers types based on the operands and operators.

**Before:**
```go
type ListTestRow struct {
    A1Float interface{} `json:"a1_float"`  // was interface{}
}
```

**After:**
```go
type ListTestRow struct {
    A1Float string `json:"a1_float"`  // now string (MySQL DECIMAL)
}
```

## What changed

- Added `inferExprType()` that recursively analyzes SQL expressions to determine their types
- Handles column references, constants, binary operations, and type casts
- MySQL-specific type rules: division returns decimal, float operations return float, nullability propagates correctly
- Enhanced COALESCE to use type inference instead of always returning `interface{}`
- Added "div" and "mod" to the mathematical operators list (MySQL uses symbolic names instead of `/` and `%`)

## Notes

The implementation is MySQL-specific for now. PostgreSQL and SQLite return `nil` from `combineGenericTypes()` to maintain current behavior. There's a TODO comment about refactoring `IsMathematicalOperator()` to be engine-specific.

Closes #4153

### DIFF
--- a/internal/compiler/infer_type.go
+++ b/internal/compiler/infer_type.go
@@ -25,6 +25,12 @@ func (c *Compiler) inferExprType(node ast.Node, tables []*Table) *Column {
 		return nil
 	}
 
+	// Only MySQL is supported for now - return nil for other engines
+	// to maintain existing behavior
+	if c.conf.Engine != config.EngineMySQL {
+		return nil
+	}
+
 	switch n := node.(type) {
 	case *ast.ColumnRef:
 		// Try to resolve the column reference


hts0000:
Hello, may I ask when this PR will be merged?

PR #4177, opened 2025-11-10
fix: robustly strip psql meta commands

### LINKED ISSUE #4065: SQLC fails for psql meta-commands like \restrict

With latest postgresql versions, they have added `--restrict-key=` and other meta-commands [here](https://www.postgresql.org/docs/current/app-pgdump.html)

When running `pg_dump` command, it adds a new line `\restrict <token>` at the top of `.sql` file, which sqlc fails to parse.

PostgreSQL

Go

ignatremizov:
## Summary

This PR makes PostgreSQL schema preprocessing safer for `pg_dump` / `psql`-flavored input without turning sqlc into a `psql` interpreter.

It preserves real PostgreSQL SQL, strips or warns on client-side commands, rejects script semantics that cannot be supported safely, and shares the apply-time schema loading path used by command setup and PostgreSQL sqltest seeding.

### PostgreSQL schema preprocessing

- Replace the old line-based PostgreSQL filter with a single-pass state machine that only strips top-level `psql` meta-commands outside SQL lexical context.
- Keep explicit coverage of documented `psql` backslash commands as a reference fixture. [See PostgreSQL `psql` docs](https://www.postgresql.org/docs/current/app-psql.html)
- Preserve backslashes that appear in valid PostgreSQL SQL, including:
  - string literals, including `E'...'`
  - quoted identifiers
  - dollar-quoted bodies
  - comments and nested block comments
- Support the valid `\meta ... \\ SQL...` separator form and keep trailing SQL intact, including glued and one-sided-whitespace separator forms observed in supported `psql` versions.
- Preserve invalid leading `\\` input instead of normalizing it into SQL.
- Handle Unicode dollar-quote tags, identifier-boundary checks, CR/LF normalization, and `\copy ... from stdin` blocks.

### Semantic command handling

- Strip semantic `psql` commands such as `\connect`, `\copy`, `\gexec`, `\i`, `\ir`, `\q`, `\quit`, and `\r` in parse/codegen paths with warnings so users get told when sqlc is ignoring client-side behavior.
- Reject those same semantic commands in apply/live-DB paths where silently flattening them would be unsafe.
- Reject `psql` conditionals (`\if`, `\elif`, `\else`, `\endif`) instead of pretending to support them.
- Keep `standard_conforming_strings` handling as a best-effort parse aid and warn in parse/codegen paths when preprocessing has to approximate session/script semantics rather than emulate `psql`.
- Suppress that parse-only approximation warning in apply/live-DB paths because the database executes transaction and session SQL directly there.

### Rollout across schema-loading paths

- Apply the shared PostgreSQL preprocessing behavior in:
  - compiler parsing / `generate`
  - `createdb`
  - `verify`
  - managed `vet`
  - PostgreSQL sqltest seeding
- Surface preprocessing warnings through the relevant callers instead of keeping them internal to the migration helper.
- Keep apply/live-DB paths strict so invalid SQL, including dumps that recreate already-present schema objects, still fails with PostgreSQL's normal errors.

### Follow-up refactor

- Extract the repeated apply-time glob/read/preprocess/warn loop into `schemautil.LoadSchemasForApply()`.
- Reuse that loader across command setup and PostgreSQL sqltest helpers so apply-time behavior stays aligned as preprocessing evolves.
- Preserve read-only PostgreSQL cache hashing on preprocessed DDL.
- Do not special-case or rewrite `pg_dump` `public` schema DDL; apply-mode callers continue to see PostgreSQL's native behavior.

## Behavioral impact

- sqlc is now much more tolerant of `pg_dump` / `psql`-flavored PostgreSQL schema files without corrupting valid SQL that happens to contain backslashes.
- Parse/codegen paths are intentionally permissive with warnings when users are in footgun territory.
- Apply/live-DB paths stay stricter and reject client-side script behavior that sqlc cannot reproduce safely.
- This does **not** try to become a full `psql` interpreter; full arbitrary client-side script semantics remain out of scope here.

## Testing

- `go test ./internal/migrations ./internal/compiler ./internal/schemautil ./internal/cmd ./internal/sqltest/...`
- `go test ./internal/migrations -coverprofile=/tmp/sqlc-migrations.cover` reports `100.0%` statement coverage for `internal/migrations`

## Related

Co-authored-by: Andrew Benton <andrew@sqlc.dev>

Addresses [gbarr's comment](https://github.com/sqlc-dev/sqlc/pull/4082#discussion_r2403920635) in https://github.com/sqlc-dev/sqlc/pull/4082 which closes https://github.com/sqlc-dev/sqlc/issues/4065

### DIFF
--- a/internal/cmd/createdb.go
+++ b/internal/cmd/createdb.go
@@ -85,7 +85,14 @@ func CreateDB(ctx context.Context, dir, filename, querySetName string, o *Option
 		if err != nil {
 			return fmt.Errorf("read file: %w", err)
 		}
-		ddl = append(ddl, migrations.RemoveRollbackStatements(string(contents)))
+		ddlText, warnings, err := migrations.PreprocessSchemaForApply(string(contents), string(queryset.Engine))
+		if err != nil {
+			return err
+		}
+		for _, warning := range warnings {
+			fmt.Fprintln(o.Stderr, warning)
+		}
+		ddl = append(ddl, ddlText)
 	}
 
 	now := time.Now().UTC().UnixNano()
--- a/internal/cmd/createdb.go
+++ b/internal/cmd/createdb.go
@@ -10,8 +10,7 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/sqlc-dev/sqlc/internal/config"
 	"github.com/sqlc-dev/sqlc/internal/dbmanager"
-	"github.com/sqlc-dev/sqlc/internal/migrations"
-	"github.com/sqlc-dev/sqlc/internal/sql/sqlpath"
+	"github.com/sqlc-dev/sqlc/internal/schemautil"
 )
 
 var createDBCmd = &cobra.Command{
@@ -76,23 +75,11 @@ func CreateDB(ctx context.Context, dir, filename, querySetName string, o *Option
 	}
 
 	var ddl []string
-	files, err := sqlpath.Glob(queryset.Schema)
+	ddl, err = schemautil.LoadSchemasForApply(queryset.Schema, string(queryset.Engine), func(warning string) {
+		fmt.Fprintln(o.Stderr, warning)
+	})
 	if err != nil {
 		return err
-	}
-	for _, schema := range files {
-		contents, err := os.ReadFile(schema)
-		if err != nil {
-			return fmt.Errorf("read file: %w", err)
-		}
-		ddlText, warnings, err := migrations.PreprocessSchemaForApply(string(contents), string(queryset.Engine))
-		if err != nil {
-			return err
-		}
-		for _, warning := range warnings {
-			fmt.Fprintln(o.Stderr, warning)
-		}
-		ddl = append(ddl, ddlText)
 	}
 
 	now := time.Now().UTC().UnixNano()
--- a/internal/cmd/generate.go
+++ b/internal/cmd/generate.go
@@ -306,6 +306,9 @@ func parse(ctx context.Context, name, dir string, sql config.SQL, combo config.C
 		return nil, true
 	}
 	if err := c.ParseCatalog(sql.Schema); err != nil {
+		for _, warning := range c.Warnings() {
+			fmt.Fprintln(stderr, warning)
+		}
 		fmt.Fprintf(stderr, "# package %s\n", name)
 		if parserErr, ok := err.(*multierr.Error); ok {
 			for _, fileErr := range parserErr.Errs() {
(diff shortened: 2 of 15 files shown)


ignatremizov:
@andrewmbenton please review

gbarr:
Thanks @ignat980  this looks a much more complete solution than I was expecting

kyleconroy:
@ignat980 if you open this against main I can get this merged. Just make sure to includes Andrew's commits.

kyleconroy (approved):

ignatremizov:
@kyleconroy Thanks! I rebased to latest sqlc/main and changed this PR's merge-into branch as sqlc/main. Just waiting on the test CI to finish

ignatremizov:
@kyleconroy bump for visibility - not sure why the CI tests haven't run, maybe you need to push something?

ignatremizov:
@kyleconroy bump again ⭐

atombender:
@kyleconroy Can you please merge this? It's been in approved state for 5 months now.

kyleconroy:
I've merged the simple fix in https://github.com/sqlc-dev/sqlc/pull/4390

If you rebase in that we can get this merged

ignatremizov:
@kyleconroy Rebased on top of #4390 (current main), but I didn’t just replay the old patch unchanged.

After checking the merged simple fix more closely, I found several correctness gaps if we wanted this to be safe against real-world `pg_dump` / `psql`-flavored schema input, not just the narrow `\restrict` / `\unrestrict` case.

So the branch now does more than just “robust stripping”, it does "correct stripping":

- keeps the single-pass state machine so we only strip top-level `psql` meta-commands outside real SQL literal/comment/body context
- preserves valid PostgreSQL SQL containing backslashes in:
  - string literals, including `E'...'`
  - quoted identifiers
  - dollar-quoted bodies
  - comments / nested block comments
- handles the valid `\meta ... \\ SQL...` separator form
- strips semantic `psql` commands in parse/codegen paths with warnings, so if users rely on psql-side behavior they get told they are now in footgun territory
- rejects those same semantic commands in apply/live-DB paths, where silently flattening them would be unsafe
- rejects psql conditionals (`\if`, `\elif`, `\else`, `\endif`) instead of pretending to support them

I did start going further down the path of "full correctness", but pretty quickly that becomes "reimplement psql", which is not the scope of this PR.

So I stopped at:
- lexically correct stripping
- semantically safer behavior for sqlc users
- warnings/errors where preprocessing becomes approximate

If we ever want to go further, the honest solution is probably not "keep accreting special cases here", but something closer to "run the SQL through a real PostgreSQL environment and reason about the resulting state after psql-side processing".

If users want full arbitrary-psql-script semantics in sqlc today, that is outside the scope of this PR and they can suck a lemon.

Also split out the repeated mechanical caller refactor as a separate follow-up commit so the actual behavior change stays readable.

kyleconroy:
**🤖 Automated message**

---

ignatremizov:
Opened https://github.com/sqlc-dev/sqlc/pull/4479

PR #4210, opened 2025-12-01
feat(codegen): Preserve casing of names by uppercasing the first character...

### LINKED ISSUE #4202: Preserve upper-case when mapping SQL identifiers to Go identifiers

It would be great if sqlc could leave upper-case letters as upper-case when mapping SQL identifiers to Go identifiers.

The existing behavior works pretty well for all-lowercase SQL identifiers using underscores:

```
(SQL)          (Go)
request_id  => RequestID
http_id     => HttpID
artwork_key => ArtworkKey
```

But some databases use camel-case naming convention. Currently, sqlc changes upper-case letters to lower-case, producing:

```
(SQL)         (Go)
RequestID  => Requestid
HTTPID     => Httpid
ArtworkKey => Artworkkey
```

My proposal: leave upper-case letters as upper-case, and otherwise keep the existing behavior. This wouldn't affect anyone who uses all-lowercase SQL names, and it would produce nicer results for mixed-case SQL names:

```
(SQL)         (Go)
RequestID  => RequestID
HTTPID     => HTTPID
ArtworkKey => ArtworkKey
```

If this is acceptable, I'd be happy to submit a patch.

How does it sound?

_No response_

Go

theAnuragMishra:
and leaving the rest in their original case.

closes #4202

### DIFF
--- a/internal/codegen/golang/struct.go
+++ b/internal/codegen/golang/struct.go
@@ -35,7 +35,9 @@ func StructName(name string, options *opts.Options) string {
 		if _, found := options.InitialismsMap[p]; found {
 			out += strings.ToUpper(p)
 		} else {
-			out += strings.Title(p)
+			if len(p) > 0 {
+				out += strings.ToUpper(p[:1]) + p[1:]
+			}
 		}
 	}
 


internal/codegen/golang/struct.go
```diff
@@ -35,7 +35,9 @@ func StructName(name string, options *opts.Options) string {
 		if _, found := options.InitialismsMap[p]; found {
 			out += strings.ToUpper(p)
 		} else {
-			out += strings.Title(p)
+			if len(p) > 0 {
+				out += strings.ToUpper(p[:1]) + p[1:]
+			}
```
em-ily-dev: There isn't any whitespace in `p` (see line 31 above), so this code will do the same thing as strings.Title.

em-ily-dev: strings.Title already does what we want here. https://github.com/sqlc-dev/sqlc/issues/4202#issuecomment-3618526408

PR #4234, opened 2025-12-15
feat(sqlite): Support NOT NULL constraints

ncruces:
You may have missed that [`sqlite3_table_column_metadata`](https://sqlite.org/c3ref/table_column_metadata.html) gets additional metadata for columns, in particular `NOT NULL` constraints.

Unfortunately because of [backwards compatibility](https://sqlite.org/lang_createtable.html#the_primary_key) a `PRIMARY KEY` column is not always `NOT NULL`, and it's really hard to _correctly_ detect `INTEGER PRIMARY KEY` columns which _are_ `NOT NULL`.

So this misses a few `NOT NULL` columns, but it's better than nothing, I guess.


ncruces:
@kyleconroy, polite bump.

I'm basically trying to point you to a feature you may have missed, since not many bindings wrap the `sqlite3_table_column_metadata` API. 

If that's inadequate, feel free to close.

Issue #4158, opened 2025-10-28
Refactor project for easy adding support of new database engine without sqlc core changes


asmyasnikov:
@kyleconroy What do you think of using an external database engine as a plugin, similar to a database/sql or GORM (for example using "_" import for register engine)?

_Originally posted by @asmyasnikov in https://github.com/sqlc-dev/sqlc/issues/4090#issuecomment-3265158841_

AI suggest next plan:
https://deepwiki.com/search/what-needs-to-be-done-with-the_bc38a550-304d-453f-99a0-071a46160042?mode=fast

Looks doable

PR #4220, opened 2025-12-06
Add ClickHouse Engine Support to sqlc

mgilbir:
## Overview

This PR introduces comprehensive support for ClickHouse as a database engine in sqlc. It enables users to write ClickHouse queries in `.sql` files and generate type-safe Go code using sqlc's familiar workflow.

### Key Features

- **ClickHouse SQL Parser**: Full SQL parsing support using the `github.com/AfterShip/clickhouse-sql-parser` library
- **Type Mapping**: ClickHouse types mapped to Go types with support for ClickHouse-specific types (Array, LowCardinality, Map, etc.)
- **Code Generation**: Generates Go database interfaces, model types, and query functions for ClickHouse operations
- **JOIN USING Support**: New SQL feature supporting `JOIN...USING` syntax, working with both ClickHouse and PostgreSQL
- **Comprehensive Testing**: 25+ end-to-end test cases covering basic operations, advanced SQL features, type handling, and functions

## Production Status

This implementation has been in production use for approximately one month. The core parser and code generation have proven stable for our queries, but I'm eager for community feedback on edge cases and other query patterns.

## Large PR Notice

⚠️ **This is a substantial PR** (~17,000 lines across 13 commits), but the impact on existing code is minimal:
- New engine implementation is isolated in `/internal/engine/clickhouse/`
- Code generation templates are isolated in `/internal/codegen/golang/templates/clickhouse/`
- Core changes to existing files are limited to:
  - Adding ClickHouse engine registration in the compiler
  - Minor type system extensions for ClickHouse types
  - Refactoring and improving `output_columns.go` (backward compatible)
  - Adding `JOIN...USING` support (benefits multiple engines)

**The majority of the PR size comes from test data and generated code examples**, which are valuable for validation and user documentation.

## Feedback Welcome

I'm committed to making this PR production-ready and welcome:
- **Architecture review** of the engine implementation
- **Bug reports** for queries that fail to parse or generate incorrectly
- **Feature requests** for missing ClickHouse capabilities
- **Code quality feedback** and suggestions for improvement
- **Test cases** for any edge cases you encounter

If you have ClickHouse queries that don't work with this implementation, please share them. I'm particularly interested in:
- Complex subqueries and CTEs
- Advanced window functions
- ClickHouse-specific features (PREWHERE, SAMPLE, array operations, etc.)
- Edge cases in type resolution

## Commit Structure

The 13 commits are organized to facilitate review, with test data grouped at the end:

### 1. **Core Engine Implementation** (2 commits)
[...]

PR #4247, opened 2025-12-28
Add database engine plugins (external engines)

asmyasnikov:
This PR resolves issue #4158 and adds support for **database engine plugins**: external processes that implement **`EngineService/Parse`** with the **same subprocess pattern as codegen plugins** — protobuf on stdin/stdout, no TCP gRPC, but the generated **`EngineServiceClient`** plus `process.Runner` (implementing `grpc.ClientConnInterface`) so the child’s argv ends with the **full gRPC method name** `/engine.EngineService/Parse`. That allows sqlc to work with databases that are not built-in (e.g. CockroachDB, TiDB, or custom SQL dialects). The plugin contract is deliberately minimal: no AST, no compiler in the middle, and a straight path from plugin output to codegen.

## Motivation

1. **sqlc is widely used**, and requests to add support for new databases come in regularly: #4244, #4243, #4090, #4220, #4009.

2. **Supporting new database engines** increases maintenance burden on core maintainers (parsers, catalogs, dialects, and compatibility for each engine).

3. **Exposing a way to add external database engines** (#4158) lets the community ship new backends without growing sqlc’s core, and lets users adopt sqlc for more SQL dialects.

4. **What should an engine plugin do?**
   - **4.1) Validate the query.** If the query is invalid, the plugin returns an error with a clear message. This is straightforward.
   - **4.2) Parse the query — but into what?** If the plugin were to return an AST in sqlc’s format (i.e. by making `internal/sql/ast` public), we risk that a new dialect cannot be expressed with the existing AST node types. This PR therefore **does not** have the plugin pass an AST into sqlc’s core.
   - **4.3)** Instead, the plugin **parses the SQL** and returns to the core **the query parameters** and **the types and names of the result columns**. That is enough for type-safe codegen.
   - **4.4)** **Optional** behavior (wildcard expansion, etc.) is left to the plugin. To support it, the core passes either `schema.sql` or database connection parameters into the plugin.
   - **4.5)** The core then feeds the plugin’s output **directly into the codegen plugin**. So an external database engine plugin **bypasses the compiler** and goes straight to code generation.

With this design, the engine plugin system can support many SQL dialects, while **validation and enrichment of SQL (e.g. `*` expansion) are the plugin’s responsibility**, not the core’s.

---

## Pipeline: built-in engine vs external plugin

The choice between "built-in engine" and "external plugin" is made **once per `sql[]` block** in **`internal/cmd/process.go`**, based on `engine:` in config: if it is `sqlite`, `mysql`, or `postgresql` → built-in path; if it is a name listed under top-level `engines:` → plugin path. (**Vet** has no plugin logic; for plugin-engine blocks it fails with "unknown engine".)

```mermaid
flowchart TB
  subgraph input["input"]
    schema[schema.sql]
    config[sqlc.yaml]
    queries[queries.sql]
  end
  subgraph sqlc_core["SQLC"]
    direction TB
    subgraph builtin_flow["Built-in Engine processing"]
      direction TB
      parser["Parse (postgresql, mysql, sqlite)"]
      ast[(AST)]
      catalog[(Catalog)]
      compiler[Compiler]
      convert_compiler_results_to_codegen[Queries + types]
      parser --> ast
      parser --> catalog
      schema --> parser
      queries --> parser
      ast --> compiler
      catalog --> compiler
      compiler --> convert_compiler_results_to_codegen
    end
    subgraph plugin_flow["Plugin Engine processing"]
      direction TB
      adapter["Engine process runner"]
      ext_parse["subprocess: argv … /engine.EngineService/Parse; stdin/out protobuf"]
      convert_plugin_response_to_codegen["Queries + types"]
      adapter --> |"Parse (generated client + Runner)"| ext_parse
      schema --> adapter
      queries --> adapter
      ext_parse --> |"ParseResponse"| convert_plugin_response_to_codegen
    end
    branch --> builtin_flow["Built-in"]
    branch --> plugin_flow
  end
  config --> branch{"Engine Type"}
  convert_plugin_response_to_codegen --> codegen
  convert_compiler_results_to_codegen --> codegen
  subgraph output["Codegen"]
    codegen[Codegen plugin]
  end
```

- **Built-in path:** Parser → AST, schema → Catalog; Compiler uses both and produces Queries + types. Codegen receives that as usual.
- **Plugin path:** Engine process runner sends **one ParseRequest per query file**: `sql` is the **entire contents** of that file, plus `schema_sql` or `connection_params`. The plugin returns **ParseResponse.statements** — one **Statement** per query block in the file (each with name, cmd, sql, parameters, columns). The plugin may also return **ParseResponse.catalog** (tables/columns derived from schema or DB); when present, sqlc passes that catalog to the codegen plugin so codegen can emit model types from the schema. Each statement is turned into a compiler query and passed into the same ProcessResult/codegen path, so N blocks → N codegen queries (N helpers). No AST, no compiler — validation and enrichment are the plugin's job.

- **No intermediate AST**: the plugin returns already “resolved” data (SQL text, parameters, columns).
- **No compiler** for the plugin path: type resolution, `*` expansion, and validation are the plugin’s job. sqlc does not run the built-in compiler on plugin output.
- Data from the plugin is passed through to the **codegen plugin** via a thin adapter that maps each **Statement** to a compiler query (name, cmd, sql, parameters, columns). N statements → N queries → N generated helpers.

So: for external engines, the pipeline is **schema + full query file → engine plugin (Parse) → N statements → N codegen queries**, with no AST and no compiler in between.

### Call flow (built-in vs plugin)

The split between built-in and plugin engines happens in **`internal/cmd/process.go`** inside **`processQuerySets()`**, which branches on **`config.IsBuiltinEngine(combo.Package.Engine)`**.

- **Built-in path:** For `engine: sqlite | mysql | postgresql`, `processQuerySets` calls **`parse()`** (defined in **`internal/cmd/generate.go`**). `parse()` calls **`compiler.NewCompiler(sql, combo, parserOpts)`**. **`NewCompiler`** in **`internal/compiler/engine.go`** has only three cases (SQLite, MySQL, PostgreSQL) and `default: return nil, fmt.Errorf("unknown engine: %s", conf.Engine)`; there is no plugin branch and no `FindEnginePlugin` in the compiler.

- **Plugin path:** For any other `engine:` (e.g. a name defined under `engines:` in config), `processQuerySets` calls **`runPluginQuerySet()`** in **`internal/cmd/plugin_engine.go`**. **`parse()`** and **`NewCompiler`** are not used; the engine process runner talks to the external plugin and passes the result into the same codegen pipeline.

- **Vet:** The vet command uses its own path and calls `parse()` / `NewCompiler()`, so vet fails for plugin engines with "unknown engine" (there is no plugin-specific branch in vet).

**Summary:** The branch and orchestration live in **`process.go`**. Built-in logic goes through **`generate.go`** → **`compiler/engine.go`**; plugin logic is in **`plugin_engine.go`**.

## No intermediate AST for external plugins

The plugin does **not** return an AST:

- **Request**: **entire query file content** (`sql`) + schema (or connection). One **Parse subprocess round-trip** **per query file** (argv tail `/engine.EngineService/Parse`, same Runner pattern as codegen).
- **Response**: **`repeated Statement statements`**. Each Statement has: `name`, `cmd` (enum: `:one` / `:many` / `:exec` / etc.), `sql`, `parameters`, `columns`. The plugin returns one Statement per query block in the file.

The plugin is the single place that defines how the query is interpreted. sqlc does not parse or analyze that SQL again; it maps each Statement to a codegen query and forwards them. N blocks in the file → N statements → N generated helpers.

---

## No compiler for external plugins

The built-in **compiler** (catalog, type resolution, validation, expansion of `*`) is **not** used for external engine plugins:

- The plugin is responsible for:
  - Resolving parameter and column types (using schema or DB).
  - Expanding `SELECT *` if desired.
  - Emitting whatever shape of `parameters` and `columns` the codegen expects.
- sqlc does not run the compiler on plugin output; it passes that output through to codegen. So “compiler” is only in the built-in-engine path.

---

## What is sent to and returned from the plugin

**Invocation**: one round-trip per query file: subprocess whose **last argv token** is the full gRPC method name **`/engine.EngineService/Parse`**, with **protobuf** `ParseRequest` on stdin and `ParseResponse` on stdout (no TCP gRPC or framing — same idea as `/plugin.CodegenService/Generate` for codegen plugins).  
Example: `sqlc-engine-mydb /engine.EngineService/Parse` (with optional flags before the method name) and `ParseRequest` on stdin / `ParseResponse` on stdout.

**Call pattern**: sqlc runs the engine plugin once **per query file** (generated `NewEngineServiceClient(process.Runner).Parse` → `Runner.Invoke` → child process). For each file (e.g. `query.sql`), the full file contents are sent in a single `ParseRequest`. The plugin returns one **Statement** per query block in that file.

### Sent to the plugin (`ParseRequest`)

| Field               | Description |
|---------------------|-------------|
| `sql`               | **Entire contents** of one query file (all blocks, including `-- name: X :cmd` comments). |
| `schema_sql`        | *(optional)* Contents of the schema file(s), e.g. concatenated `schema.sql`. |
| `connection_params` | *(optional)* DSN + options for “database-only” mode when schema is taken from the DB. |

Exactly one of `schema_sql` or `connection_params` is used per request, depending on project configuration.

### Proto schema (`protos/engine/engine.proto`)

The file defines **messages** (`ParseRequest`, `ParseResponse`, `Statement`, `Catalog`, …) and a **`service EngineService { rpc Parse … }`** for documentation and **generated Go client stubs** (`NewEngineServiceClient`, full method constant `EngineService_Parse_FullMethodName`). Transport remains **stdin/stdout protobuf** (no TCP): sqlc uses the same **`process.Runner` + `ClientConnInterface.Invoke`** pattern as codegen, not a network gRPC stack.

### Returned from the plugin (`ParseResponse`)

| Field         | Description |
|---------------|-------------|
| `statements`  | One **Statement** per query block in the request. |
| `catalog`     | *(optional)* **Catalog** derived from `schema_sql` or `connection_params`. When present, sqlc passes it to the codegen plugin so codegen can emit model structs from the schema (tables/columns) instead of only query-row types. The Catalog type is defined in `engine.proto` (Identifier, Catalog, CatalogSchema, CatalogTable, CatalogColumn) and is converted to the plugin codegen Catalog when building the GenerateRequest. |

Each **Statement** has:

| Field        | Description |
|--------------|-------------|
| `name`       | Query name (from `-- name: GetUser` etc.). |
| `cmd`        | Command/type: enum `Cmd` (`CMD_ONE`, `CMD_MANY`, `CMD_EXEC`, `CMD_EXEC_RESULT`, `CMD_EXEC_ROWS`, `CMD_EXEC_LAST_ID`, `CMD_COPY_FROM`, `CMD_BATCH_EXEC`, `CMD_BATCH_MANY`, `CMD_BATCH_ONE`). Matches sqlc’s `:one`, `:many`, `:exec`, etc. |
| `sql`        | Processed SQL for that block (as-is or with `*` expanded). |
| `parameters` | Parameters for this statement (name, position, data_type, nullable, is_array, array_dims). |
| `columns`    | Result columns for this statement (name, data_type, nullable, is_array, array_dims, optional table/schema). |

There is no top-level `sql` / `parameters` / `columns`; the response is `statements` and optionally `catalog`. N blocks in the file → N statements → N codegen queries (N helpers). If the plugin returns a non-empty **catalog**, sqlc forwards it to the codegen plugin so generated code can use table models (e.g. `type Author struct`) from the schema, not only per-query row types.

### Helpers in `pkg/engine` for plugin authors

To split the query file and parse `"-- name: X :cmd"` lines in the same way as built-in engines, the engine package provides (optional) helpers that delegate to `internal/metadata`:

- **`CommentSyntax`** — which comment styles to accept (Dash, SlashStar, Hash).
- **`ParseNameAndCmd(line, syntax)`** — parses a line like `"-- name: ListAuthors :many"` → `(name, cmd Cmd, ok)`.
- **`QueryBlocks(content, syntax)`** — splits file content into `[]QueryBlock` (Name, Cmd, SQL).
- **`StatementMeta(name, cmd, sql)`** — builds a `*Statement` with name/cmd/sql set; the plugin fills parameters and columns.
- **`CmdToString(c)`** / **`CmdFromString(s)`** — convert between enum `Cmd` and strings like `":one"`, `":many"`.

---

## How the schema is passed into the plugin

Schema is provided to the plugin in one of two ways, via `ParseRequest.schema_source`:

1. **Schema-based (files)**  
   - sqlc reads the configured schema files (e.g. `schema: "schema.sql"`) and passes their contents as **`schema_sql`** (a string) in `ParseRequest`.  
   - The plugin parses this SQL (e.g. `CREATE TABLE ...`) and uses it to resolve types, expand `*`, etc.

2. **Database-only**  
   - When schema is not from files, sqlc can pass **`connection_params`** (DSN + optional extra options) in `ParseRequest`.  
   - The plugin connects to the DB and uses live metadata (e.g. `INFORMATION_SCHEMA` / `pg_catalog`) to resolve types and columns.

So: **schema** is either “schema.sql as text” or “connection params to the database”; the plugin chooses how to use it.

---

### New top-level `engines`

Plugins are declared under `engines` and referenced by name in `sql[].engine`:

```yaml
version: "2"

engines:
  - name: mydb
    process:
      cmd: sqlc-engine-mydb
    env:
      - MYDB_DSN

sql:
  - engine: mydb
    schema: "schema.sql"
    queries: "queries.sql"
    codegen:
      - plugin: go
        out: db
```

- **`engines`**: list of named engines. Each has `name` and either `process.cmd` (and optionally `env`) or a WASM config.
- **`sql[].engine`**: for that SQL block, use the engine named `mydb` (which triggers the plugin) instead of `postgresql` / `mysql` / `sqlite`.

So the only new concept in config is “define engines (including plugins) by name, then point `sql[].engine` at them.” Schema and queries are still configured per `sql[]` block as today.

---

## Who handles sqlc placeholders in queries

Support for sqlc-style placeholders (`sqlc.arg()`, `sqlc.narg()`, `sqlc.slice()`, `sqlc.embed()`, etc.) is **entirely up to the plugin**:

- The plugin receives the full query file content (including those macros) in `ParseRequest.sql`.
- It can parse and interpret them and reflect the result in `parameters` (and, if needed, in `sql` or in how it uses schema). There is no separate “sqlc placeholder” pass in the core for the plugin path.
- If the plugin does not handle a placeholder, that placeholder will not be turned into proper parameters/columns by sqlc; the pipeline does not add a generic placeholder expander for external engines.

So: **the database engine plugin is responsible for understanding and handling sqlc placeholders** for its engine.

---

## Summary for maintainers

- **One `Parse` invocation per query file** (via `EngineServiceClient` + `process.Runner`): subprocess argv ends with `/engine.EngineService/Parse`; reads `ParseRequest` (full query file + `schema_sql` | `connection_params`) from stdin, writes `ParseResponse{ statements [, catalog ] }` to stdout. The Go SDK `engine.Run` still accepts the legacy argv `parse` for compatibility.
- **Response**: `repeated Statement statements` and optionally **`Catalog catalog`**. Each Statement has `name`, `cmd` (enum), `sql`, `parameters`, `columns`. If the plugin returns a non-empty catalog (tables/columns from schema or DB), sqlc converts it to the codegen Catalog and passes it in GenerateRequest so codegen can emit model structs from the schema.
- **N blocks → N statements → N codegen queries** (N helpers).
- **No AST, no compiler** on the plugin path; each Statement is mapped to a compiler query and passed to codegen.
- **Schema** is passed either as `schema_sql` (file contents) or as `connection_params` (DSN) in `ParseRequest`.
- **Config**: `engines[]` + `sql[].engine: <name>`; existing `schema` / `queries` / `codegen` stay as-is.
- **Placeholders**: handled inside the plugin; core does not add a generic layer for external engines.
- **Helpers** in `pkg/engine` (ParseNameAndCmd, QueryBlocks, StatementMeta, Cmd enum) reuse `internal/metadata` so plugin behavior matches built-in engines.

This keeps the plugin API small and leaves type resolution and dialect behavior inside the plugin, while still allowing sqlc to drive generation from a single, well-defined contract.


asmyasnikov:
YDB over engine plugin works fine (in my MVP) https://github.com/ydb-platform/sqlc-ydb/

asmyasnikov:
@kyleconroy soft ping ^^^

asmyasnikov:
Another approach - use https://pkg.go.dev/plugin
So database engine will be a library ("so" file) and sqlc app will load library and lookup export constructor for database engine
This approach decrease a transport difficulty - no need to think about request/response protos for meet sqlc app and database engine plugin
But standard plugin system works only on linux/macos/freebsd, windows is of of scope
And need to build plugin only with origin sqlc version (commit, equal dependencies). Different builds of plugin and sql are not compatible with other

internal/cmd/process.go
```diff
@@ -8,11 +8,14 @@ import (
 	"path/filepath"
 	"runtime"
 	"runtime/trace"
+	"strings"
 
 	"golang.org/x/sync/errgroup"
 
 	"github.com/sqlc-dev/sqlc/internal/compiler"
 	"github.com/sqlc-dev/sqlc/internal/config"
+	"github.com/sqlc-dev/sqlc/internal/debug"
```
asmyasnikov:
```suggestion
```

asmyasnikov:
@kyleconroy Can you see this PR? We are waiting YDB support in SQLC very much. Plugin engine can unblocked external RDBMs support

kyleconroy:
@asmyasnikov I don't have any plans to merge something like this. I'm currently rebuilding the analysis framework. It's going to look substantially different than this. You'll need to maintain a fork if you want plugins to work exactly like this.

PR #4299, opened 2026-02-16
fix: set Type field on INSERT parameter columns for correct schema resolution

rayakame:
I'm writing an external sqlc plugin ([sqlc-gen-better-python](https://github.com/rayakame/sqlc-gen-better-python)) and ran into a bug where `:copyfrom` queries targeting tables in custom schemas send incorrect type information to plugins.

When using a schema like this:

```sql
CREATE SCHEMA IF NOT EXISTS custom;
CREATE TYPE custom.mood AS ENUM ('sad', 'ok', 'happy');
CREATE TABLE custom.test_enum (
    id int PRIMARY KEY NOT NULL,
    m  custom.mood NOT NULL
);
```

And a query like:

```sql
-- name: InsertStuff :copyfrom
INSERT INTO custom.test_enum (id, m) VALUES ($1, $2);
```

The plugin receives the parameter type as `name: "custom.mood"` instead of `schema: "custom", name: "mood"`. This meant my plugin couldn't resolve the enum type to its definition in the catalog, since the schema was baked into the name string rather than being properly separated.

I noticed that `:many` SELECT queries on the exact same table returned correct type info, so the bug had to be specific to how INSERT parameter columns are built.

## Cause

The `*ast.ResTarget` case in `resolveCatalogRefs` (`internal/compiler/resolve.go`) builds `Column` structs with `DataType` (a string like `"custom.mood"`) but doesn't set the structured `Type` field (`*ast.TypeName` with separate `Schema` and `Name`). The shim in `internal/cmd/shim.go` checks `Type` first and falls back to putting `DataType` into `Name` with an empty schema when `Type` is nil.

## Previous fix / possible unnecessary code

In `/internal/codegen/golang/postgresql_type.go` there is the function `func parseIdentifierString(name string) (*plugin.Identifier, error)` which previously parsed the string generated by `sdk.DataType` into a new `*plugin.Identifier`. As far as I've tested this, this function can just be removed now and the `column.Type` can be used directly.

I could of course implement the same workaround in my own plugin, but I think fixing this at the source is the right approach.


rayakame:
As far as I can see the failing test is only something that went wrong with docker

PR #4302, opened 2026-02-18
Prioritize native database installations over Docker

kyleconroy:
## Summary
This change reorders the database server startup logic to prioritize native installations over Docker containers in the SQL test utilities for both MySQL and PostgreSQL.

## Key Changes
- **MySQL tests** (`internal/sqltest/local/mysql.go`):
  - Swapped the order of checks: now attempts native installation first, then falls back to Docker
  - Updated fallback comment to reflect the new priority

- **PostgreSQL tests** (`internal/sqltest/local/postgres.go`):
  - Applied the same reordering: native installation checked first, Docker as fallback
  - Updated fallback comment to reflect the new priority

- **Skip messages**: Updated error messages in both files to reflect the new priority order (native before Docker)

## Rationale
This change ensures that native database installations are preferred when available, as they typically offer better performance and lower overhead compared to Docker containers. Docker is now used as a fallback option only when native installations are not available.

https://claude.ai/code/session_01XbZNE4YfrtML4B6sLVf2E3

### DIFF
--- a/internal/sqltest/local/mysql.go
+++ b/internal/sqltest/local/mysql.go
@@ -26,21 +26,21 @@ func MySQL(t *testing.T, migrations []string) string {
 
 	dburi := os.Getenv("MYSQL_SERVER_URI")
 	if dburi == "" {
-		if ierr := docker.Installed(); ierr == nil {
-			u, err := docker.StartMySQLServer(ctx)
-			if err != nil {
-				t.Fatal(err)
-			}
-			dburi = u
-		} else if ierr := native.Supported(); ierr == nil {
-			// Fall back to native installation when Docker is not available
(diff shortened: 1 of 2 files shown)


rubensantoniorosa2704:
Hey, I was looking at the CI failures and noticed something interesting. Looking at the logs, MySQL tests are all failing with:
```
INFO native/mysql status="starting existing service"
MySQL is not installed or could not be started
```
And for PostgreSQL there's a different failure:
```
failed SASL auth: FATAL: password authentication failed for user "postgres"
```

I also took a look at `ci.yml` and didn't see any database setup steps. I'm curious — was the fallback to Docker tested in the CI environment after this change? The logs suggest it might not be kicking in as expected.

PR #4308, opened 2026-02-22
fix(sqlite): normalize column names in ALTER TABLE ADD/DROP COLUMN

### LINKED ISSUE #4307: ALTER TABLE ADD COLUMN statement with mixed case column name results in not found
### Version

1.30.0

### What happened?

An ALTER TABLE ... ADD COLUMN statement with a mixed-case column name results in column does not exist error.

If I modify the case in the ALTER TABLE statement to be all lowercase, it works.

In the playground it's using 1.25, but I have verified this happens with 1.30, too.

### Relevant log output

```shell
query.sql:2:17: column "barbaz" does not exist
```

### Database schema

```sql
CREATE TABLE foo (
  id INTEGER NOT NULL PRIMARY KEY,
  xYz INTEGER NOT NULL DEFAULT 0
);

ALTER TABLE foo ADD COLUMN barBaz TEXT NOT NULL DEFAULT '';
```

### SQL queries

```sql
-- name: GetOne :one
SELECT id, xYz, barBaz FROM foo WHERE id = ?;
```

### Configuration

```yaml
{
  "version": "2",
  "sql": [{
    "schema": "schema.sql",
    "queries": "query.sql",
    "engine": "sqlite",
    "gen": {
      "go": {
        "out": "db"
      }
    }
  }]
}
```

### Playground URL

https://play.sqlc.dev/p/1c10bdf06465cd0a1bf7237b01f4e869a9c174d21920a882a891d1114809a27a

Linux

SQLite

Go

stakeswky:
Fixes #4307

## Problem
When  or  statements use mixed-case column names (e.g., `barBaz`), the column name was stored in the catalog with its original case. However,  queries normalize all unquoted identifiers to lowercase via the `identifier()` function, causing a mismatch and "column does not exist" errors.

## Solution
This fix ensures  also normalize column names using `identifier()`, matching the behavior of  and  queries. This aligns with SQLite's case-insensitive identifier handling.

## Changes
- `internal/engine/sqlite/convert.go`: Apply `identifier()` to column names in ALTER TABLE ADD COLUMN and DROP COLUMN handlers (lines 70 and 91)
- `internal/engine/sqlite/catalog_test.go`: Add regression test for mixed-case column names in ALTER TABLE ADD COLUMN

## Testing
All existing tests pass, including the new regression test case that reproduces the issue from #4307.

```bash
$ go test ./internal/engine/sqlite/ -run TestUpdate -v
--- PASS: TestUpdate (0.08s)
    --- PASS: TestUpdate/3 (0.01s)  # New mixed-case test
```

### DIFF
--- a/internal/engine/sqlite/catalog_test.go
+++ b/internal/engine/sqlite/catalog_test.go
@@ -83,6 +83,38 @@ func TestUpdate(t *testing.T) {
 			},
 		},
 		{
+			// Regression test for #4307: mixed-case column name in ALTER TABLE ADD COLUMN
+			`
+			CREATE TABLE foo (id INTEGER NOT NULL, xYz INTEGER NOT NULL DEFAULT 0);
+			ALTER TABLE foo ADD COLUMN barBaz TEXT NOT NULL DEFAULT '';
+			`,
+			&catalog.Schema{
+				Name: "main",
+				Tables: []*catalog.Table{
+					{
+						Rel: &ast.TableName{Name: "foo"},
+						Columns: []*catalog.Column{
+							{
+								Name:      "id",
+								Type:      ast.TypeName{Name: "INTEGER"},
+								IsNotNull: true,
+							},
+							{
+								Name:      "xyz",
+								Type:      ast.TypeName{Name: "INTEGER"},
+								IsNotNull: true,
+							},
+							{
+								Name:      "barbaz",
+								Type:      ast.TypeName{Name: "TEXT"},
+								IsNotNull: true,
+							},
+						},
+					},
+				},
+			},
+		},
+		{
 			`
 			CREATE TABLE foo (bar text);
 			ALTER TABLE foo RENAME COLUMN bar TO baz;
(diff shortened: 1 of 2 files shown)


mem:
I have verified that this fix works.

Thanks!

rubensantoniorosa2704 (commented):
I've verified this pr locally. I confirmed that it correctly normalizes unquoted identifiers in ALTER TABLE statements.

I also ran a manual test with quoted identifiers (e.g., "PreserveMyCase") to ensure they are still handled correctly by the identifier() function, and they are.

@stakeswky, would you consider adding this test case to catalog_test.go to ensure we don't have regressions regarding quoted identifiers in the future?

```go
{
    `
    CREATE TABLE quoted_test (id INTEGER PRIMARY KEY);
    ALTER TABLE quoted_test ADD COLUMN "PreserveMyCase" TEXT NOT NULL DEFAULT '';
    `,
    &catalog.Schema{
        Name: "main",
        Tables: []*catalog.Table{
            {
                Rel: &ast.TableName{Name: "quoted_test"},
                Columns: []*catalog.Column{
                    {
                        Name: "id",
                        Type: ast.TypeName{Name: "INTEGER"},
                        IsNotNull: true,
                    },
                    {
                        Name:      "PreserveMyCase",
                        Type:      ast.TypeName{Name: "TEXT"},
                        IsNotNull: true,
                    },
                },
            },
        },
    },
},
```

Nice fix, thanks!

cod3rboy:
I encountered the same problem when using camel-case column names in queries.
Is there anything pending in this PR keeping it from being merged?

PR #4342, opened 2026-03-15
Support batching different queries together.

### LINKED ISSUE #4195: FR: Support batching different queries together
## Context
https://github.com/sqlc-dev/sqlc/issues/1275 was resolved by adding `:batchexec`, `:batchone`, and `:batchmany` commands generating code for submitting multiple instances of the same query in a batch. https://github.com/sqlc-dev/sqlc/issues/1275#issuecomment-2368589913 on that feature requests batching for different types of queries. This FR is for addressing that request.

Related FR: https://github.com/sqlc-dev/sqlc/issues/3752 requests "a query annotation to enable batching across multiple tables". This FR would address that usecase, though not through the addition of a new query annotation.

## Proposal
 - Generate a `QueryBatch` type which supports queueing and typesafe processing of any eligible query on the current generated Queries object.
   - "Eligible" here likely excludes `:batchexec`, `:batchone`, and `:batchmany`, `:copyfrom`, and possibly others that do not work in the context of an existing batch.
- For each supported method X on Queries, we create a corresponding method on QueryBatch which returns an XBatchResult object for retrieving batch results in a typesafe manner.
- A QueryBatch is submitted to get a QueryBatchResults type, which the XBatchResult objects can call to get and return typesafe results.
- Internally we will need to ensure the correct ordered processing of the batch results coming back from the database.

The goal is to support convenient batch execution of any already-defined queries, rather than requiring a separate query with a custom annotation and limiting usage to a single query type in a batch.

I can draft a PR proposing an implementation. In doing this, we should be careful not to conflict with the existing batch command support.

While a similar construct could be created in other languages and databases, I am interested in adding this for Go / pgx.

PostgreSQL

Go

jgiles:
Fix #4195 by adding an emit_query_batch codegen option that creates a QueryBatch type when using pgx v5.

When emit_query_batch: true is set in config (pgx/v5 only), sqlc generates:
- A QueryBatch struct wrapping pgx.Batch
- NewQueryBatch() constructor
- Queue* methods for each query (:one, :many, :exec, :execrows, :execresult)
- ExecuteBatch() on Queries to send all queued queries in one round-trip

This expands the batching support of existing :batchone/:batchmany/:batchexec annotations, which required separate query definitions and only supported baching the same query with different parameters. Batching multiple instances of the same query with different parameters is also supported via the new interface. While there is probably not a good reason to use both in the same package, you can generate and use both without error for backwards compatibility.

Key design decisions

- Follows pgx v5's recommended QueuedQuery callback pattern
- QueryBatch.Batch is exported so users can mix generated Queue* calls with custom pgx batch operations
- :exec queries have no callback (consistent with pgx - errors propagate via Close/ExecuteBatch)
- :one callbacks receive a bool indicating whether a row was found

Changes

- internal/codegen/golang/gen.go - Wire up EmitQueryBatch option, file generation, and validation
- internal/codegen/golang/imports.go - Add queryBatchImports() with import logic that skips struct field types (struct definitions live in query.sql.go, not the batch file). Extract queryUsesType() from inline closure for reuse.
- internal/codegen/golang/opts/options.go - Add EmitQueryBatch and OutputQueryBatchFileName options
- internal/codegen/golang/templates/pgx/queryBatchCode.tmpl - New template for all query batch methods
- internal/codegen/golang/templates/template.tmpl - Add queryBatchFile and queryBatchCode template definitions

Test coverage

- emit_query_batch - Full test with :one, :many, :exec, :execrows, :execresult
- emit_query_batch_db_arg - With emit_methods_with_db_argument: true
- emit_query_batch_minimal - Non-struct return types (verifies import handling for e.g. pgtype.Timestamptz)
- emit_query_batch_overrides - With custom type overrides (verifies batch file doesn't import types only used in struct fields)
- emit_query_batch_with_batch - Combines old-style :batchexec with emit_query_batch (verifies both batch.go and query_batch.sql.go coexist correctly)

Documentation

- docs/reference/config.md - Added emit_query_batch and output_query_batch_file_name
- docs/reference/query-annotations.md - New section with usage examples contrasting with :batch* annotations


jgiles:
Note that while the PR is sized XXL, a large majority of the change by file and line count is new end to end tests covering different scenarios related to the new batching functionality.

bmperrea (approved):

adrianotr:
Hi! Is there a plan for merging this PR? I’d really benefit from this feature. Thanks for your work on it!

PR #4388, opened 2026-04-19
Emit pointers for nullable enum columns when emit_pointers_for_null_types is set

### LINKED ISSUE #3276: Support generating pointers for nullable enum types (emit_pointers_for_null_types for enums)

Currently, when a field is nullable and an enum, sqlc always generates a new type with a prefix `Null`. This `Null___` type is a struct that wraps the enum's string alias type and adds a `Valid` boolean property to represent the `NULL` state.

sqlc has a setting called `emit_pointers_for_null_types` which replaces similar nullable wrapper types with pointers instead (`NullString` -> `*string`)

This feature request seeks to bring similar functionality to nullable enum fields. Instead of generating a `Null___` type for a nullable enum field, it should be a represented by a pointer (`NullFooEnum` -> `*FooEnum`)

## Questions

- If this feature were to be implemented, should it be enabled by the existing `emit_pointers_for_null_types` config setting? Or should it be a separate setting?

## Other notes

I couldn't find an existing issue for this, but there is an open PR #3093 which seemed to be working towards the same goal. I'm not sure if the approach there is sufficient to implement this feature, but it might be a 

_No response_

_No response_

kyleconroy:
Closes #3276. When emit_pointers_for_null_types is true (PostgreSQL with
pgx/v4 or pgx/v5), nullable enum columns are now generated as pointers
(*UserRole) instead of the NullUserRole wrapper struct, matching the
existing behavior for other nullable types.

To preserve the previous behavior when upgrading, set
emit_pointers_for_null_enum_types: false. When explicitly set, this flag
overrides emit_pointers_for_null_types for enums only, so it can also be
used to opt into the pointer treatment for enums without enabling it for
every other null type.

https://claude.ai/code/session_01K2xVtbacEbjbA9ymg6dPtr


winens:
@kyleconroy did you just reject my pull and then copied the same commit with your own name?

PR #4403, opened 2026-04-22
Validate unknown columns in SELECT WHERE clauses (fixes #4264)

### LINKED ISSUE #4264: No error raised about obvious typo in WHERE clause
### Version

1.30.0

### What happened?

This query should fail, since column `is_deleted` doesn't exist (should be`deleted`). But it compiles just fine and only errors when being called! Additional problem: The generated `query.sql.go` doesn't even select the users name, leading to broken SQL and empty user object:
```go
const getUsers = `-- name: GetUsers :many
select  from user where is_deleted = false
`

type GetUsersRow struct {
}
````

### Relevant log output

```shell
no logs, which is the problem :)
```

### Database schema

```sql
create table "user" ("name" text not null, deleted bool not null);
```

### SQL queries

```sql
-- name: GetUsers :many
select * from user where is_deleted = false;
```

### Configuration

```yaml
version: 2
sql:
  - schema: schema.sql
    queries: query.sql
    engine: postgresql
    gen:
      go:
        package: sqlc
        out: ./
```

### Playground URL

_No response_

Linux

PostgreSQL

Go

lawrence3699:
sqlc currently validates bad column references in SELECT targets, GROUP BY, and ORDER BY, but it lets unknown columns in SELECT WHERE clauses slip through to generated code. That means queries like `select * from "user" where is_deleted = false` pass `sqlc generate` even though PostgreSQL rejects them at execution time. Fixes #4264.

Run column-reference validation over SELECT WHERE clauses as well, and recurse into nested SELECTs so the same check also catches unknown columns inside EXISTS/scalar subqueries without breaking correlated references to outer scopes.

Add end-to-end fixtures for both the flat WHERE case and a nested EXISTS subquery case, with base and managed-db expectations.

Validation:
- `go test ./internal/endtoend -run 'TestReplay/base/invalid_where_unknown_column/postgresql/pgx$' -count=1`
- `go test ./internal/endtoend -run 'TestReplay/base/invalid_where_unknown_column_subquery/postgresql/pgx$' -count=1`
- `POSTGRESQL_SERVER_URI='postgresql://lawrence@/postgres?host=/tmp&sslmode=disable' go test ./internal/endtoend -run 'TestReplay/managed-db/invalid_where_unknown_column/postgresql/pgx$' -count=1`
- `POSTGRESQL_SERVER_URI='postgresql://lawrence@/postgres?host=/tmp&sslmode=disable' go test ./internal/endtoend -run 'TestReplay/managed-db/invalid_where_unknown_column_subquery/postgresql/pgx$' -count=1`
- `go test ./internal/endtoend -run 'TestReplay/base/select_exists/pgx/v5$' -count=1`
- `go test ./internal/endtoend -run 'TestReplay/base/select_subquery/postgresql/stdlib$' -count=1`


lawrence3699:
Closing this because the PR currently has a failing test check and I am not actively fixing it, so it should not remain open as a ready contribution.

PR #4411, opened 2026-04-23
fix: handle named params in SQLite upsert ON CONFLICT DO UPDATE SET

### LINKED ISSUE #3508: upset with arg or narg doesn't work in sqlite3
### Version

1.25.0

### What happened?

sqlc fails to generate query placeholders and leaves `@arg` `sqlc.arg` `sqlc.narg` in generated query when doing upsert.
Doesn't output any errors.
Possibly related to #1881 

Generated code from playground
```go
const upsertAuthor = `-- name: UpsertAuthor :one
insert into authors (
  "id",
  "name",
  "bio"
) values (
  ?1,
  ?2,
  ?3
) on conflict("id") do update set
  "name" = @name,
  "bio" = sqlc.narg('bio')
returning id, name, bio
`
```

### Relevant log output

_No response_

### Database schema

```sql
create table authors (
  id   text primary key,
  name text not null,
  bio  text
);
```

### SQL queries

```sql
-- name: UpsertAuthor :one
insert into authors (
  "id",
  "name",
  "bio"
) values (
  @id,
  @name,
  sqlc.narg('bio')
) on conflict("id") do update set
  "name" = @name,
  "bio" = sqlc.narg('bio')
returning *;
```

### Configuration

```yaml
{
  "version": "2",
  "sql": [{
    "schema": "schema.sql",
    "queries": "query.sql",
    "engine": "sqlite",
    "gen": {
      "go": {
        "out": "db"
      }
    }
  }]
}
```

### Playground URL

https://play.sqlc.dev/p/4b01c7bd73781a3aeaf17fa1c9fba826640a5907f97b2faaee51ccac9197d00e

Linux

SQLite

Go

Bhagwat-45:
## Problem
When using `@param`, `sqlc.arg()`, or `sqlc.narg()` in the
`ON CONFLICT DO UPDATE SET` clause of a SQLite upsert, sqlc leaves
the raw parameter references unrewritten in the generated SQL instead
of replacing them with `?N` placeholders.

## Root Cause
`convertInsert_stmtContext` never called `n.Upsert_clause()`, so the
`OnConflictClause` field was always nil and params inside it were
never processed by the rewriter.

## Fix
Added `convertUpsert_clauseContext` that properly builds the
`OnConflictClause` AST node with infer clause, target list, and
where clause — calling `c.convert()` on all expressions so named
params are correctly rewritten.

## Test
Added end-to-end test `upsert_named_params/sqlite` reproducing the
exact query from the issue.

Fixes #3508


rubensantoniorosa2704:
Tested this locally and the basic case works great. Nice work.

It would be great to see a few more test cases in upsert_named_params/sqlite covering the WHERE clause variations: a WHERE on the DO UPDATE side with a named param, a WHERE on the conflict target itself, and ideally both together.

Bhagwat-45:
Thanks for the feedback! Added three WHERE clause test cases to `upsert_named_params/sqlite` covering:
- WHERE on the DO UPDATE side with a named param (`@name`)
- WHERE on the DO UPDATE side with `sqlc.narg`
- Both together

Generated using a local build to confirm the fix handles all variations correctly.

PR #4419, opened 2026-04-29
fix: handle DROP TABLE CASCADE for dependent views in schema replay

### LINKED ISSUE #4416: PostgreSQL schema parser does not drop dependent views for DROP TABLE ... CASCADE
### Version

1.30.0

### What happened?

When `sqlc generate` replays a PostgreSQL schema containing `DROP TABLE ... CASCADE`, it appears to leave dependent views in its internal schema catalog.

In real Postgres, dropping a table with `CASCADE` drops views that depend on that table. In sqlc, the dependent view still appears to exist, so a later `CREATE VIEW` with the same name fails.

Minimal example ([Example on play.sqlc.dev](https://play.sqlc.dev/p/77fea362bc83d66c4b8439685849e22d2a7e472774e3a6a386074477ecbf080e)):

```sql
CREATE TABLE reference_rates (
    id BIGSERIAL PRIMARY KEY,
    data_ts TIMESTAMPTZ NOT NULL
);

CREATE VIEW vw_reference_rates AS
SELECT *
FROM reference_rates;

CREATE TABLE reference_rates_new (
    id BIGSERIAL PRIMARY KEY,
    data_ts TIMESTAMPTZ NOT NULL
);

DROP TABLE reference_rates CASCADE;

ALTER TABLE reference_rates_new RENAME TO reference_rates;

CREATE VIEW vw_reference_rates AS
SELECT *
FROM reference_rates;
```

I expected `sqlc generate` to succeed because this SQL is valid in Postgres. The `DROP TABLE reference_rates CASCADE` statement should remove `vw_reference_rates`, allowing the later `CREATE VIEW vw_reference_rates` to succeed.

Instead, sqlc fails with relation "vw_reference_rates" already exists.

This reproduces on `v1.31.1`. I also reproduced it on `v1.30.0`, and on `v1.31.1` with `SQLCEXPERIMENT=analyzerv2`.

### Relevant log output

```shell
# package db
schema.sql:1:1: relation "vw_reference_rates" already exists
```

### Database schema

```sql
CREATE TABLE reference_rates (
    id BIGSERIAL PRIMARY KEY,
    data_ts TIMESTAMPTZ NOT NULL
);

CREATE VIEW vw_reference_rates AS
SELECT *
FROM reference_rates;

CREATE TABLE reference_rates_new (
    id BIGSERIAL PRIMARY KEY,
    data_ts TIMESTAMPTZ NOT NULL
);

DROP TABLE reference_rates CASCADE;

ALTER TABLE reference_rates_new RENAME TO reference_rates;

CREATE VIEW vw_reference_rates AS
SELECT *
FROM reference_rates;
```

### SQL queries

```sql
-- name: ListReferenceRates :many
SELECT id, data_ts
FROM vw_reference_rates
ORDER BY id;
```

### Configuration

```yaml
{
  "version": "2",
  "sql": [
    {
      "engine": "postgresql",
      "schema": "schema.sql",
      "queries": "query.sql",
      "gen": {
        "go": {
          "package": "db",
          "out": "db"
        }
      }
    }
  ]
}
```

### Playground URL

https://play.sqlc.dev/p/77fea362bc83d66c4b8439685849e22d2a7e472774e3a6a386074477ecbf080e

macOS

PostgreSQL

Go

wucm667:
## Problem

When `sqlc generate` replays a PostgreSQL schema containing `DROP TABLE ... CASCADE`, dependent views remain in the internal catalog. This causes subsequent `CREATE VIEW` statements with the same name to fail with `relation "vw_reference_rates" already exists`.

Minimal repro ([Playground](https://play.sqlc.dev/p/77fea362bc83d66c4b8439685849e22d2a7e472774e3a6a386074477ecbf080e)):

```sql
CREATE TABLE reference_rates (id BIGSERIAL PRIMARY KEY);
CREATE VIEW vw_reference_rates AS SELECT * FROM reference_rates;
DROP TABLE reference_rates CASCADE;
ALTER TABLE reference_rates_new RENAME TO reference_rates;
CREATE VIEW vw_reference_rates AS SELECT * FROM reference_rates;  -- fails
```

In real Postgres, `DROP TABLE ... CASCADE` drops both the table and dependent views. In sqlc's replay, the view persists.

## Root Cause

1. `ast.DropTableStmt` lacked a `Behavior` field, so the parser never captured whether `CASCADE` or `RESTRICT` was specified.
2. `dropTable()` in the catalog only removed the table and linked types — it never checked for or removed dependent views.
3. Views stored in the catalog had no dependency tracking.

### 1. Capture CASCADE/RESTRICT in the AST (`internal/sql/ast/drop_table_stmt.go`)

Added `Behavior DropBehavior` field.

### 2. Parser extracts Behavior (`internal/engine/postgresql/parse.go`)

When converting `DropStmt` for `OBJECT_TABLE`, `OBJECT_VIEW`, or `OBJECT_MATVIEW`, pass through `n.Behavior`.

### 3. Track view dependencies (`internal/sql/catalog/view.go`)

Added `DependsOnTables []*ast.TableName` to `Table` struct. The `createView` function now walks the SELECT query AST to extract all `RangeVar` (table reference) nodes, populating dependencies at view creation time.

### 4. CASCADE drops dependent views (`internal/sql/catalog/table.go`)

When `Behavior == 2` (`DROP_CASCADE` in pg_query_go protobuf), `dropTable` scans all tables/views in the schema and removes any that depend on the dropped table.

### 5. Tests (`internal/engine/postgresql/catalog_test.go`)

- `TestDropTableCascadeViewRecreate`: verifies CASCADE removes dependent views.
- `TestDropTableCascadeWithoutCascadeFails`: verifies without CASCADE, views remain in catalog.

## Testing

```bash
go test -v -count=1 ./internal/engine/postgresql/... -run "TestDropTableCascade|TestUpdateErrors"
```

All existing `TestUpdateErrors` cases pass; new CASCADE tests pass.

Fixes #4416

### DIFF
--- a/internal/engine/postgresql/catalog_test.go
+++ b/internal/engine/postgresql/catalog_test.go
@@ -6,6 +6,8 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/sqlc-dev/sqlc/internal/sql/ast"
+	"github.com/sqlc-dev/sqlc/internal/sql/catalog"
 	"github.com/sqlc-dev/sqlc/internal/sql/sqlerr"
 
 	"github.com/google/go-cmp/cmp"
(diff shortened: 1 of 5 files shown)


0x4A756E65:
Would love for this to be considered - I believe it fixes my exact issue, so thank you @wucm667

luongs3:
I just closed #4456 in favor of this — I worked the same fix independently before finding yours. Same approach lands at the same place, so this is the one to merge. One thing I tried that's worth a quick consider:

In `internal/sql/ast`, the pg_query DropBehavior enum (UNDEFINED=0, RESTRICT=1, CASCADE=2) is checked as `Behavior == 2` in dropTable(). Adding named constants in `drop_behavior.go`:

```go
const (
    DropBehaviorUndefined DropBehavior = 0
    DropBehaviorRestrict  DropBehavior = 1
    DropBehaviorCascade   DropBehavior = 2
)
```

and using `stmt.Behavior == ast.DropBehaviorCascade` reads more intent-fully and would survive a future pg_query enum reshuffle. Tiny patch, no behavior change. Happy to PR it as a follow-up if you and the maintainers like it.

Otherwise this looks solid — the RangeVar walk for view DependsOn is the natural place to capture deps, and the transitive eviction handles the views-on-views case I was worried about.

wucm667:
@luongs3 Thanks for the kind words and for the suggestion! I agree the named constants 
would read better than == 2. If the maintainers want it, I'm happy to add it as part of 
this PR or you can do a follow-up — whichever works best.

@kyleconroy @jzelinskie Would love a review when you have a moment 🙂

luongs3:
Follow-up draft up at #4461 — just the named `DropBehavior` constants, no use-site changes. Tagging here so it's discoverable from this thread. Happy for it to land independently before this PR, or for the rename to be folded into this PR (or a sweep PR after this merges) — whichever the maintainers prefer.

Will keep #4461 in draft until there's a signal either way.

PR #4421, opened 2026-04-30
feat(golang): emit models to a different package

### LINKED ISSUE #835: Export models to a different package
Hi! Great work on this project!

I was wondering if it's currently possible to export Go model structs to a different folder package.

I think this could be a good feature for anyone who would like to make their code independent of the database backend but still use the automatically generated models.

For example:
```
internal/
  model/
    models.go
  db/
    postgres/
      db.go
      user.sql.go
      querier.go
```

This way you could e.g. also have a mongodb provider that uses these models, but without depending on the `postgres` package. I'm sure you'd agree that it would be odd for mongo to use `postgres.User` as a type.

Is this something that's possible to do? If not, would you possibly accept a PR that adds support for it? Or it it something you're not planning on supporting for now?

Thanks

kyleconroy:
Closes #835.

Adds four Go codegen options that let users place table model structs
and enums in a Go package distinct from the queries package, optionally
shared across multiple SQL blocks:

  - output_models_path: directory for the models file
  - output_models_package: Go package name for the models file
  - output_models_import: import path for the models package
  - output_models_emit: when false, skip emitting models.go and only
    reference an externally-emitted models package

When the models package differs from the queries package, query files
import the models package and reference types as `model.User` instead of
`User`. Embedded model structs in synthetic Row structs and enum types
in synthetic Params structs are qualified consistently.

End-to-end fixtures cover both shapes: a single block with models in a
sibling directory, and two query blocks that share one models package.

https://claude.ai/code/session_01Nj4LbGbqFBTU1EPKKNHtNn


justpranavrs:
Very Much Appreciated. Thanks for this Great Addition :). Helps a Lot

PR #4426, opened 2026-05-05
fix(golang): reject go_type specs with slash before type name

### LINKED ISSUE #4192: Type override, where new type has an underlying basic type, results in an invalid import
### Version

1.30.0

### What happened?

Given the following override definition:
```yaml
...
      go:
        package: "somepackage"
...
          - column: "table.column"
            go_type: "github.com/import/from/within/the/module/somepackage/SomeType"
```
Where the column is of type TEXT, and the type is defined within the module and package where the result is being generated. The type definition is:

```go
package somepackage

type SomeType string
```

The models.go generated file ends up with the following import line:
```go
package somepackage

import (
...
	"github"
```

Ideally, it must not attempt to import anything, since the type is defined from within the package that houses the generated files.

### Relevant log output

```shell

```

### Database schema

```sql

```

### SQL queries

```sql

```

### Configuration

```yaml

```

### Playground URL

_No response_

_No response_

_No response_

_No response_

SAY-5:
Fixes #4192.

A `go_type` spec like `github.com/foo/pkg/Type` (slash before the type name instead of the conventional dot) used to fall back to the last dot anywhere in the spec, which lands inside `github.com`. The generated `models.go` ended up with `import ("github")`, which does not compile.

This change rejects these specs in the parser with the same error already used for specs that contain no dot, pointing the user at the expected `'package.type'` format. Adds a regression case in `TestTypeOverrides`.

### DIFF
--- a/internal/codegen/golang/opts/go_type.go
+++ b/internal/codegen/golang/opts/go_type.go
@@ -146,7 +146,12 @@ func (gt GoType) parse() (*ParsedGoType, error) {
 		o.BasicType = true
 	} else {
 		// assume the type lives in a Go package
-		if lastDot == -1 {
+		if lastDot == -1 || lastDot < lastSlash {
+			// Either no dot at all, or every dot is part of the import path
+			// (e.g. "github.com/foo/pkg/Type"). Without a dot separating
+			// package from type after the last slash, we cannot tell where
+			// the import path ends and the type name begins, and silently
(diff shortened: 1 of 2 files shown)


SAY-5:
Bumping this one. It just makes a go_type like "pkg/Type" fail validation up front instead of silently generating code that won't compile.

PR #4430, opened 2026-05-08
fix: resolver fails to look up CTE columns

### LINKED ISSUE #4288: sqlc generate failed：column "new_streak" does not exist
### Version

1.30.0

### What happened?

Playground: 

This issue occurs when using sqlc version 1.30 both in the local environment and in the Playground. I have confirmed that the same error is reproducible in both environments.

### Relevant log output

```shell
sqlc generate failed.
```

### Database schema

```sql
CREATE TABLE users
(
    id                  BIGSERIAL PRIMARY KEY,
    credit              BIGINT      DEFAULT 0           NOT NULL CHECK ( credit >= 0 ),
    last_checkin_day    DATE,
    checkin_streak      INT         DEFAULT 0           NOT NULL CHECK ( checkin_streak >= 0 )
);

CREATE TABLE credit_log
(
    id          BIGSERIAL PRIMARY KEY,
    user_id     BIGINT                    NOT NULL REFERENCES users (id),
    amount      INT                       NOT NULL,
    type        TEXT                      NOT NULL,
    description TEXT,
    created_at  timestamptz DEFAULT NOW() NOT NULL
);
```

### SQL queries

```sql
-- name: CheckinUser :one
WITH target AS (
    SELECT id, last_checkin_day, checkin_streak
    FROM users
    WHERE users.id = $1
    FOR UPDATE
),
tz AS (
    SELECT (NOW() AT TIME ZONE 'Asia/Shanghai')::date AS today
),
abc (new_streak, id, applied) AS (
    SELECT id,
           (last_checkin_day IS NULL OR last_checkin_day <> (SELECT today FROM tz)) AS applied,
           CASE
               WHEN last_checkin_day = (SELECT today FROM tz) - INTERVAL '1 day' THEN checkin_streak + 1
               ELSE 1
               END AS new_streak
    FROM target
),
reward AS (
    SELECT id,
           applied,
           new_streak,
           CASE
               WHEN applied THEN $2 + ($2 * LEAST(GREATEST(new_streak - 1, 0), 5) / 5)
               ELSE 0
               END AS total_reward
    FROM abc 
),
updated AS (
    UPDATE users u
        SET last_checkin_day = CASE WHEN r.applied THEN (SELECT today FROM tz) ELSE u.last_checkin_day END,
            checkin_streak   = CASE WHEN r.applied THEN r.new_streak ELSE u.checkin_streak END,
            credit           = u.credit + r.total_reward
        FROM reward r
        WHERE u.id = r.id
        RETURNING r.applied, r.new_streak, r.total_reward, u.id, u.credit, u.last_checkin_day, u.checkin_streak
),
log AS (
    INSERT INTO credit_log (user_id, amount, type, description)
        SELECT $1, total_reward, 'checkin', $3
        FROM updated
        WHERE applied
        RETURNING id
)
SELECT applied::INT AS applied,
       new_streak,
       total_reward,
       id, credit, last_checkin_day, checkin_streak
FROM updated;

```

### Configuration

```yaml

```

### Playground URL

https://play.sqlc.dev/p/20c431276d4f5fd75aaaa81d1c9930307498080a4e0c02e2664f833297e0df1e

Windows11

postgresql

go

msaher:
Closes #4288 

The resolver fails to infer the type of a CTE column because it only searches normal tables. My solution is to have two passes, one where we search the CTE columns first (making them shadow regular tables) then if not found, we search through regular tables.

We cannot simply add the CTE column in the existing `typeMap` because a CTE and a regular table will have the same column name which will trigger the `found > 1` check and the column will be considered ambiguous. So we use a different type map for CTEs so we can separate the search

I have noticed three other cases where only regular tables are searched. They are `ast.BetweenExpr`, `ast.ResTarget`, and `ast.In`. I'm not sure if they should be changed so I left them as they are and they're not failing any tests, but I added a comment about them in case we need to update them. 

There is some duplication when populating the CTEs and searching through them. I have chosen not to create a helper function for searching as that seems to be the style. However, I'm happy to create one if the maintainers think its needed.


internal/endtoend/testdata/cte_left_join/postgresql/pgx/go/query.sql.go
```diff
@@ -28,7 +26,7 @@ FROM
 WHERE c1.name = $1
 `
 
-func (q *Queries) BadQuery(ctx context.Context, dollar_1 pgtype.Text) error {
-	_, err := q.db.Exec(ctx, badQuery, dollar_1)
+func (q *Queries) BadQuery(ctx context.Context, name string) error {
+	_, err := q.db.Exec(ctx, badQuery, name)
```
msaher, line 30: This change affects the existing `cte_left_join` test case. It now uses native Go types

PR #4444, opened 2026-05-14
[Feature] Add query metadata comments

### LINKED ISSUE #4489: Add configurable SQL metadata comments to generated queries
## What do you want to change?

I would like sqlc to support opt-in query metadata comments in generated SQL strings. The goal is to make SQL generated by sqlc easier to trace in database logs, APM tools, and database monitoring/tracing products that preserve or parse SQL comments.

Today, when a generated query appears as a slow query, sampled trace, or database log entry, it can be difficult to map it back to the originating sqlc query name and source file. Tools like Datadog Database Monitoring and ecosystems such as sqlcommenter and Marginalia use structured SQL comments for that correlation.

Proposed configuration:

```yaml
version: "2"
sql:
  - schema: "schema.sql"
    queries: "query.sql"
    engine: "postgresql"
    query_comments:
      enabled: true
      format: "sqlcommenter"
      tags:
        - name
        - cmd
        - filename
```

When enabled, sqlc would prepend a structured block comment to generated query text, for example:

```sql
/*sqlc_name='GetAuthor',sqlc_cmd='%3Aone',sqlc_filename='query.sql'*/
-- name: GetAuthor :one
SELECT id, name FROM authors WHERE id = $1
```

The feature should be disabled by default. A minimal first version could support:

- Formats: `sqlcommenter`, `marginalia`
- Tags: `name`, `cmd`, `filename`
- Default format: `sqlcommenter`
- Default tags: `["name"]`

This would help users correlate runtime SQL back to sqlc definitions without wrapper code or hand-editing individual queries. It would also make sqlc-generated queries friendlier to observability tooling while keeping generated SQL unchanged unless explicitly enabled.

Prior art and context:

- sqlcommenter-style comments are used by OpenTelemetry-related libraries and observability tools.
- Marginalia-style comments are common in Rails applications.
- Datadog Database Monitoring and similar tools preserve or display SQL comments for query attribution.
- I previously opened a prototype PR at https://github.com/sqlc-dev/sqlc/pull/4444, but it was closed because this feature request should be discussed first.

## What database engines need to be changed?

PostgreSQL, MySQL, SQLite

The mechanism is generated SQL text rather than engine-specific parsing, so this should be useful across all engines as long as comments are preserved by the driver/database path.

## What programming language backends need to be changed?

Go

The initial use case is Go-generated query constants. If the project prefers, the same metadata could be added before built-in/plugin code generation so additional backends or plugin consumers can opt in later.

DerekTBrown:
Fixes #4489

## Problem Statement

Database monitoring and tracing tools (ex. Datadog) rely on metadata comments to trace SQL statements. I am looking for an easy way to inject these into SQLC-generated statements, improving traceability and auditability of SQL queries.

## Proposal

This adds a `query_comments` config block that prepends generated SQL with structured metadata comments in either `sqlcommenter` or `marginalia` format based on the tags specified in the SQLC comment:

```yaml
version: "2"
sql:
  - schema: "schema.sql"
    query_comments:
      enabled: true
      format: "sqlcommenter"
      tags:
        - name
        - filename
```

All generated SQL statements will then contain the tracking comment:

```sql
/*sqlc_name='GetAuthor',sqlc_filename='query.sql'*/
-- name: GetAuthor :one
SELECT id, name FROM authors WHERE id = $1
```


kyleconroy:
@DerekTBrown please open up an issue or a discussion with your feature request before submitting any code. Thanks.

PR #4447, opened 2026-05-19
fix(sqlite): allow to use overrides for columns which were created using "alter table .. add column"

leonklingele:
This change adds a failing test which is then fixed by a follow-up commit.

Normalize sqlite `ALTER TABLE` column identifiers and preserve analyzer
origin metadata when merging query analysis. This allows exact column
overrides such as `chat_messages.updated_at` to match columns added via
`ALTER TABLE .. ADD COLUMN`.

Using the sqlc config:

    overrides:
      - column: "*.*_at"
        go_type:
          import: "time"
          type: "Time"

and the SQL schema:

    CREATE TABLE chat_messages (
      id TEXT PRIMARY KEY NOT NULL,
      body TEXT NOT NULL,
      created_at TEXT NOT NULL DEFAULT CURRENT_TIMESTAMP
    ) STRICT;
    
    ALTER TABLE chat_messages ADD COLUMN "updated_at" TEXT NOT NULL DEFAULT '';

sqlc would previously generate the following code:

    type ChatMessage struct {
    	ID        string
    	Body      string
    	CreatedAt time.Time
    	UpdatedAt string    // Bug: This should be a time.Time!
    }

_With_ this fix, the following code is generated:

    type ChatMessage struct {
    	ID        string
    	Body      string
    	CreatedAt time.Time
    	UpdatedAt time.Time // Fixed, now a time.Time, even for retroactively added columns
    }


leonklingele:
Fixed the failing test. Can you have a look @kyleconroy? :)

PR #4471, opened 2026-06-07
fix(sqlite): correct parameter binding when mixing sqlc.arg() with bare "?"

abgoyal:
### What happened?

When a SQLite query mixes `sqlc.arg()` named parameters with bare positional `?` placeholders, sqlc generates SQL with mismatched parameter numbering that silently binds values to wrong columns.

Given this query:
```sql
-- name: RenameTemplate :one
UPDATE templates SET name = sqlc.arg(new_name), updated_at = datetime('now')
WHERE org_id = ? AND name = sqlc.arg(old_name) AND deleted_at IS NULL
RETURNING name, content, valid, updated_at;
```

**Before (broken):** sqlc generates `SET name = ?2 ... WHERE org_id = ? AND name = ?3` — the bare `?` gets auto-assigned index 1 by SQLite, but the Go function passes `NewName` as the 1st positional arg. Result: `SET name = OrgID`, `WHERE org_id = NewName`.

**After (fixed):** all placeholders are numbered in text order: `SET name = ?1 ... WHERE org_id = ?2 AND name = ?3`, matching the positional argument order.

### Root cause

Commit c2dcd56a5 ("Allow for mixed parameters types ($1 or ?) and sqlc.arg()") added mixed-param support for PostgreSQL and MySQL but did not cover SQLite. PostgreSQL doesn't have this issue because `$N` is inherently numbered and args are sorted by number. MySQL doesn't have it because all params emit bare `?` (no numbering, purely positional). SQLite is the only engine that generates numbered `?N` for named params but also accepts unnumbered `?` — and those two styles have incompatible binding semantics when mixed in the same query.

The `NamedParameters` rewriter assigns numbered `?N` to `sqlc.arg()` params while skipping positions pre-reserved for bare `?` — but it never renumbers the bare `?` itself. Since SQLite's auto-numbering for bare `?` is independent of explicit `?N`, the two schemes conflict.

### Fix

When SQLite has both named params and bare `?` in the same query:
1. Clear the pre-reserved positions so `ParamSet` doesn't skip them
2. Handle bare `?` during the Apply walk alongside named params, assigning all numbers in text order
3. Emit edits to replace bare `?` with explicit `?N`

This ensures the numbered placeholders in the output SQL match the positional argument order in the generated Go function.

### Test plan

- Added `sqlite` variant to existing `mix_param_types` end-to-end test (matching the pg/mysql variants from the original mixed-param commit)
- Verified all existing SQLite tests produce identical output (`sqlc diff` shows no changes)
- `go test -run TestReplay/base ./internal/endtoend/` passes


abgoyal:
can i please get a review for this?

PR #4472, opened 2026-06-07
Expose source_tables on the plugin Query message

Prateeks16:
## What

Adds a `source_tables` field to the plugin `Query` message: the list of base tables a query reads from, deduplicated and sorted ascending.

## Why

Plugins currently learn a query's tables only through `Query.columns[].table`, which covers only tables that contribute a column to the output. A table referenced solely in a `JOIN`, a subquery, or a common table expression body is invisible to plugins even though the query depends on it. Plugins that build reactive/refetch logic (for example keyed off SQLite's `update_hook`) need a query's full read set, not just its output columns.

## What's included

- `source_tables` field on the `Query` proto (and regenerated `codegen.pb.go`).
- The compiler resolves the read set from the statement AST: every base table in a `FROM`, a `JOIN`, or a subquery in any clause, including CTE bodies. Common table expression names and the target relations of `INSERT`/`UPDATE`/`DELETE`/`TRUNCATE` are excluded.
- A unit test (`internal/compiler`) covering the example below plus deduplication, write-target exclusion, and the no-tables case. The existing JSON codegen goldens are regenerated to include the new field.

For this query:

```sql
WITH filtered_accounts AS (
    SELECT account_id FROM accounts WHERE accounts.space_id = $1
    AND NOT EXISTS (SELECT 1 FROM account_tags t WHERE t.account_id = accounts.account_id)
)
SELECT acc.* FROM accounts acc
JOIN filtered_accounts fa ON acc.account_id = fa.account_id
LEFT JOIN transactions t ON t.debit_account_id = acc.account_id;
```

`source_tables` is `["account_tags", "accounts", "transactions"]`: `accounts` appears once (read in the CTE body and the outer query), `account_tags` appears though it is read only inside a `NOT EXISTS` subquery, `transactions` appears though it projects no column, and `filtered_accounts` is excluded as a CTE name.


Prateeks16:
Replaced by a clean branch.

PR #4473, opened 2026-06-07
Expose source_tables on the plugin Query message

### LINKED ISSUE #4434: Feature: expose source_tables on Query proto for plugin access to full query table dependencies
## Problem

Plugins receive `Query.columns[].table` which only covers tables contributing
to SELECT output. Tables referenced only in subqueries, CTEs, or JOINs are
invisible to plugins.

Example — plugin only sees `accounts`, misses `account_tags` and `transactions`:

```sql
-- name: ListSpaceAccounts :many
WITH filtered_accounts AS (
    SELECT account_id FROM accounts WHERE space_id = ?
    AND NOT EXISTS (
        SELECT 1 FROM account_tags t WHERE t.account_id = accounts.account_id
    )
)
SELECT acc.* FROM accounts acc
JOIN filtered_accounts fa ON acc.account_id = fa.account_id
LEFT JOIN transactions t ON t.debit_account_id = acc.account_id
```

## Use case

SQLite's `update_hook` fires with the name of the written table. Frontend plugins
generating reactive query stores need every table a query depends on — not just
output columns — to know when to refetch. Currently the only workaround is running
`sqlc parse` as a separate post-codegen script outside the plugin WASM sandbox.

## Proposed change

Add `source_tables` to the `Query` proto:

```protobuf
message Query {
  // ... existing fields ...
  repeated string source_tables = 7; // all tables this query reads from, CTEs excluded
}
```

sqlc's analyzer already resolves this internally when validating joins and column
references — this only surfaces existing data through the proto, no new analysis needed.

## Expected result
```
source_tables: ["account_tags", "accounts", "transactions"]
```
## Alternatives considered

| Approach | Problem |
|---|---|
| Regex on `Query.text` inside plugin | Fragile, false positives |
| `sqlc parse` as post-codegen script | Can't run from inside WASM sandbox |
| `Column.table` from output columns | Misses subquery/CTE/JOIN-only dependencies |

_No response_

_No response_

Prateeks16:
## What

Adds a `source_tables` field to the plugin `Query` message: the list of base tables a query reads from, deduplicated and sorted ascending.

## Why

Plugins currently learn a query's tables only through `Query.columns[].table`, which covers only tables that contribute a column to the output. A table referenced solely in a `JOIN`, a subquery, or a common table expression body is invisible to plugins even though the query depends on it. Plugins that build reactive/refetch logic (for example keyed off SQLite's `update_hook`) need a query's full read set, not just its output columns.

## What's included

- `source_tables` field on the `Query` proto (and regenerated `codegen.pb.go`).
- The compiler resolves the read set from the statement AST: every base table in a `FROM`, a `JOIN`, or a subquery in any clause, including CTE bodies. Common table expression names and the target relations of `INSERT`/`UPDATE`/`DELETE`/`TRUNCATE` are excluded.
- A unit test (`internal/compiler`) covering the example below plus deduplication, write-target exclusion, and the no-tables case. The existing JSON codegen goldens are regenerated to include the new field.

For this query:

```sql
WITH filtered_accounts AS (
    SELECT account_id FROM accounts WHERE accounts.space_id = $1
    AND NOT EXISTS (SELECT 1 FROM account_tags t WHERE t.account_id = accounts.account_id)
)
SELECT acc.* FROM accounts acc
JOIN filtered_accounts fa ON acc.account_id = fa.account_id
LEFT JOIN transactions t ON t.debit_account_id = acc.account_id;
```

`source_tables` is `["account_tags", "accounts", "transactions"]`: `accounts` appears once (read in the CTE body and the outer query), `account_tags` appears though it is read only inside a `NOT EXISTS` subquery, `transactions` appears though it projects no column, and `filtered_accounts` is excluded as a CTE name.

Closes #4434.


Prateeks16:
Thanks for the review. Addressed in the follow-up commit:

- **Schema qualification:** `source_tables` now reports schema-qualified names when a schema is present and deduplicates on the qualified form, so tables of the same name in different schemas (e.g. `audit.accounts` vs `accounts`) stay distinct. Unqualified references keep their bare name, matching the issue's expected output. Added a unit test for this.
- **nil vs empty slice:** the function now returns an empty slice instead of `nil`, so the field serializes as `[]` rather than `null`. Updated the no-tables test case accordingly.
- **Field documentation:** added a doc comment on the proto field (and the compiler model) describing what is included/excluded, the ordering, and the qualification rule.
- **`ParseTableName` error handling:** `ParseTableName` is total for a `*ast.RangeVar` (`parseRelation` reads the relation name directly and cannot return an error for that node type), so a fallback there would be unreachable. I left it as a guard rather than adding dead defensive code, but happy to plumb an error upward instead if you'd prefer.

PR #4479, opened 2026-06-10
fix: robustly strip psql meta commands

### LINKED ISSUE #4065: SQLC fails for psql meta-commands like \restrict

ignatremizov:
## Summary

This PR makes PostgreSQL schema preprocessing safer for `pg_dump` / `psql`-flavored input without turning sqlc into a `psql` interpreter.

It preserves valid PostgreSQL SQL, strips non-semantic client commands, warns when ignored commands may affect results, and rejects script semantics that cannot be reproduced safely.

### PostgreSQL schema preprocessing

- Replace the line-based PostgreSQL filter with a single-pass state machine that recognizes top-level `psql` meta-commands only outside SQL lexical contexts.
- Preserve backslashes in valid PostgreSQL SQL, including:
  - regular and escape string literals
  - quoted identifiers
  - dollar-quoted bodies
  - line comments and nested block comments
- Support the valid `\meta ... \\ SQL...` separator form while preserving trailing SQL, including glued and one-sided-whitespace forms accepted by supported `psql` versions.
- Preserve invalid leading `\\` input instead of normalizing it into SQL.
- Handle Unicode dollar-quote tags, identifier boundaries, CR/LF line endings, and `\copy ... from stdin` data blocks.

### Safe handling by execution mode

- In parse/codegen paths, strip semantic client commands such as `\connect`, `\copy`, `\gexec`, `\i`, `\ir`, `\q`, `\quit`, and `\r`, and surface warnings when ignoring them may change results.
- In apply/live-database paths, reject commands whose effects sqlc cannot reproduce safely.
- Reject `psql` conditionals (`\if`, `\elif`, `\else`, `\endif`) instead of flattening branches and changing SQL semantics.
- Treat `standard_conforming_strings` changes as a best-effort parsing aid and warn when parse/codegen preprocessing must approximate session behavior.
- Keep apply/live-database paths strict while allowing PostgreSQL to execute ordinary transaction and session SQL directly.

### Schema-loading integration

- Add engine-aware `PreprocessSchema()` and `PreprocessSchemaForApply()` entry points.
- Apply preprocessing consistently in:
  - compiler parsing and `generate`
  - `createdb`
  - `verify`
  - managed `vet`
  - PostgreSQL test database seeding
- Propagate non-fatal preprocessing warnings through compiler and command callers.
- Pass apply-safe schema text to managed analyzers while retaining parse-safe schema text for catalog analysis.

### Shared apply-time loading

- Add `schemautil.LoadSchemasForApply()` to centralize schema glob expansion, file loading, apply-mode preprocessing, and warning delivery.
- Reuse the loader across command setup and PostgreSQL test database seeding so apply-time behavior remains consistent.
- Preserve read-only PostgreSQL fixture caching by hashing the preprocessed DDL.
- Leave ordinary PostgreSQL errors unchanged; for example, schema DDL that conflicts with an existing database object still fails normally.

## Behavioral impact

- sqlc accepts more `pg_dump` / `psql`-flavored schema files without corrupting valid SQL containing backslashes.
- Parse/codegen paths remain permissive where safe and report ignored client behavior.
- Apply/live-database paths reject client-side semantics that cannot be reproduced reliably.
- Arbitrary `psql` script execution remains intentionally out of scope.

## Testing

- `go test ./internal/migrations ./internal/compiler ./internal/schemautil ./internal/cmd ./internal/sqltest/local`
- Regression coverage includes documented and unknown meta-commands, literals, comments, dollar quotes, inline separators, semantic warnings, apply-mode rejections, copy data, line endings, compiler warning propagation, and PostgreSQL apply-time loading.
- `go test ./internal/migrations -coverprofile=/tmp/sqlc-migrations.cover` reports `100.0%` statement coverage for `internal/migrations`.

## Related

Replaces https://github.com/sqlc-dev/sqlc/pull/4177.

Addresses [gbarr's review comment](https://github.com/sqlc-dev/sqlc/pull/4082#discussion_r2403920635) on https://github.com/sqlc-dev/sqlc/pull/4082, which closes https://github.com/sqlc-dev/sqlc/issues/4065.

Co-authored-by: Andrew Benton <andrew@sqlc.dev>

### DIFF
--- a/internal/cmd/createdb.go
+++ b/internal/cmd/createdb.go
@@ -84,7 +84,14 @@ func CreateDB(ctx context.Context, dir, filename, querySetName string, o *Option
 		if err != nil {
 			return fmt.Errorf("read file: %w", err)
 		}
-		ddl = append(ddl, migrations.RemoveRollbackStatements(string(contents)))
+		ddlText, warnings, err := migrations.PreprocessSchemaForApply(string(contents), string(queryset.Engine))
+		if err != nil {
+			return err
+		}
+		for _, warning := range warnings {
+			fmt.Fprintln(o.Stderr, warning)
+		}
+		ddl = append(ddl, ddlText)
 	}
 
 	now := time.Now().UTC().UnixNano()
--- a/internal/cmd/createdb.go
+++ b/internal/cmd/createdb.go
@@ -10,8 +10,7 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/sqlc-dev/sqlc/internal/config"
 	"github.com/sqlc-dev/sqlc/internal/dbmanager"
-	"github.com/sqlc-dev/sqlc/internal/migrations"
-	"github.com/sqlc-dev/sqlc/internal/sql/sqlpath"
+	"github.com/sqlc-dev/sqlc/internal/schemautil"
 )
 
 var createDBCmd = &cobra.Command{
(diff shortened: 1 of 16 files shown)


ignatremizov:
@kyleconroy FYI

ignatremizov:
@kyleconroy rebased onto the latest main again. Is there anything else needed before this can be merged?

PR #4484, opened 2026-06-12
docs: add example for directory-per-migration glob pattern

KhotSaprem:


KhotSaprem:
Hi, I'd like to work on this. I'll add a documentation example showing the migrations/*/up.sql glob pattern to the DDL howto page.

Issue #4464, opened 2026-06-01
add :stream keyword`


borissmidt:
### What do you want to change?

Earlier requests have been made to return an iterator, but we have to be careful not to break the 'RETURNING' statements that don't work well with a cursor. Also the developer kind of knows it is a 'big query' or not. 

So instead of changing existing behavior i would suggest adding a new keyword `:stream` for returning an iterator and using a cursor. Then it is up to the developer himself to decide can i return an iterator or array.

Related issue in the python generator:
https://github.com/sqlc-dev/sqlc-gen-python/issues/101

Previous issues:
https://github.com/sqlc-dev/sqlc/issues/4108
https://github.com/sqlc-dev/sqlc/issues/720

### What database engines need to be changed?

PostgreSQL, SQLite, MySQL

### What programming language backends need to be changed?

Go, Python, Kotlin

borissmidt:
I don't mind picking up the python generator changes needed for this.

Issue #4108, opened 2025-09-15
Proposal: generate low‑level query prepare and row‑binding helpers for streaming `:many` results


benjaco:
### What do you want to change?

`sqlc` generates methods for `:many` queries that read all rows into a slice. For example, the parameterised `ListAuthorsByIDs` query executes:

```sql
-- name: ListAuthorsByIDs :many
SELECT id, bio, birth_year FROM authors
WHERE id = ANY($1::int[]);
```

and the generated Go method allocates a `[]Author`, iterates over `rows.Next()`, scans each row into a struct and appends it to the slice before returning. This is convenient but can be inefficient for large result sets because it requires storing all rows in memory.

<details>
<summary>Standard :many generated code</summary>

```go
const listAuthors = `-- name: ListAuthors :many
SELECT id, name, bio FROM authors
ORDER BY name
`

func (q *Queries) ListAuthors(ctx context.Context) ([]Author, error) {
	rows, err := q.db.QueryContext(ctx, listAuthors)
	if err != nil {
		return nil, err
	}
	defer rows.Close()
	var items []Author
	for rows.Next() {
		var i Author
		if err := rows.Scan(&i.ID, &i.Name, &i.Bio); err != nil {
			return nil, err
		}
		items = append(items, i)
	}
	if err := rows.Close(); err != nil {
		return nil, err
	}
	if err := rows.Err(); err != nil {
		return nil, err
	}
	return items, nil
}
```
</details>

There have been discussions about adding streaming APIs (e.g. via a `:stream` or `:callback` annotation) [1](https://github.com/sqlc-dev/sqlc/discussions/1229) and a long‑running PR to support iterators [2](https://github.com/sqlc-dev/sqlc/pull/3631). A maintainer noted that the forthcoming Go `iter` package may influence how such a feature should be implemented, and no built‑in solution has been decided on yet.

### Proposal

To avoid committing to a particular iterator/channel API, I propose generating two low‑level helper functions per `:many` query when a new `emit_stream_helpers` option is enabled. These helpers decouple executing the query from scanning rows, enabling callers to implement their own streaming or pooling logic while relying on `sqlc`’s type‑safe scanning.

#### Example using `ListAuthorsByIDs`

Today `sqlc` generates a `ListAuthorsByIDs` method that returns `([]Author, error)` and reads all rows into a slice. With `emit_stream_helpers: true`, `sqlc` could additionally generate helpers like these:

```go
[...]

PR #4488, opened 2026-06-14
feat(codegen/go): PoC for opt-in iter.Seq2 companions on :many queries

Tilzen:
## Why

Revives the long-standing request for streaming/lazy `:many` query results ([#720](https://github.com/sqlc-dev/sqlc/issues/720)). Today, sqlc always materializes `:many` into `[]T`, which is fine for small result sets but costly for exports, sync jobs, and backfills.

Go 1.23+ shipped native iterators (`iter.Seq2`). Kyle [noted](https://github.com/sqlc-dev/sqlc/issues/720#issuecomment-1821962820) that this unblocks native iterator generation; [PR #3631](https://github.com/sqlc-dev/sqlc/pull/3631) explored an implementation but closed without merge after API design remained open.

This PR is **not a merge-ready feature**. It is a **proof of concept (PoC)** posted for design feedback before investing in a production implementation. See the discussion proposal in [#720](https://github.com/sqlc-dev/sqlc/issues/720#issuecomment-4702731093) and related [#4464](https://github.com/sqlc-dev/sqlc/issues/4464).

**PoC reference (fork):** https://github.com/Tilzen/sqlc/tree/feat/emit-iterators-poc  
**PRD in PoC branch:** [`docs/emit-iterators-prd.md`](https://github.com/Tilzen/sqlc/blob/feat/emit-iterators-poc/docs/emit-iterators-prd.md)

## What Changed (PoC scope)

Opt-in config (zero breaking changes to default `:many` → `[]T`):

```yaml
gen:
  go:
    emit_iterators: true
    iterator_scope: global         # or explicit_only
    iterator_method_prefix: "Iter"
    iterator_style: seq2           # MVP: seq2 only
    iterator_start: lazy
```

Generates a lazy companion alongside existing slice methods:

```go
func (q *Queries) ListAuthors(ctx context.Context) ([]Author, error) { /* unchanged */ }

func (q *Queries) IterAuthors(ctx context.Context) iter.Seq2[Author, error] {
    return func(yield func(Author, error) bool) {
        rows, err := q.db.QueryContext(ctx, listAuthors)
        // defer rows.Close(), yield rows one at a time
    }
}
```

Also supports query annotations for `explicit_only` scope:

- `-- name: StreamAuthors :stream`
- `-- name: ListAuthors :many:stream`

**PoC includes:** stdlib + pgx/v5 templates, metadata parsing, config wiring, PRD doc.

**Explicitly out of PoC scope:** `callback`/`rows` styles, eager mode, `emit_interface` iter methods, Python/Kotlin generators, full driver matrix.

## Open Design Questions (feedback welcome)

1. Method prefix: `Iter` vs `Stream`?
2. Default lazy start — agree?
3. Ship both `emit_iterators: global` and `:stream` annotations?
4. Should `:stream` remain a separate query kind or only `:many:stream`?
5. Min Go version bump when enabled?

## Test Plan

PoC test coverage (all passing on the fork branch):

- [x] **Unit:** iterator naming, scope logic, stream annotation parsing, opts validation
- [x] **Codegen matrix:** global, explicit_only (`:stream`, `:many:stream`), `emit_iterators: false` regression, parameterized `:many`, pgx/v5
- [x] **Golden replay:** `TestReplay/emit_iterators*` fixtures
- [x] **Runtime (SQLite):** iter/slice parity, lazy start (error only on first `range`), early `break` + connection reuse, parameterized queries

```bash
go test ./internal/codegen/golang/... ./internal/metadata/... \
  ./internal/endtoend/ -run 'EmitIterators|TestReplay/emit_iterators|ValidateOptsEmitIterators|Iterator'

cd internal/endtoend/testdata
go test ./emit_iterators/... ./emit_iterators_params/... -v
```

## References

- Proposal comment: [#720#issuecomment-4702731093](https://github.com/sqlc-dev/sqlc/issues/720#issuecomment-4702731093)
- Related: [#4464](https://github.com/sqlc-dev/sqlc/issues/4464), [#4108](https://github.com/sqlc-dev/sqlc/issues/4108)
- Prior art (closed): [#3631](https://github.com/sqlc-dev/sqlc/pull/3631)
- **PoC fork branch:** https://github.com/Tilzen/sqlc/tree/feat/emit-iterators-poc

## Notes

- **This is a PoC for discussion**, not a request to merge as-is. Happy to close this PR and iterate on design in #720 if that is preferred.
- If maintainers sign off on the config surface + default API (`seq2` + lazy + `Iter` prefix), I can refine the PoC toward a production-ready PR in follow-up commits.


benjaco:
Thanks for working on this. One limitation I think is worth considering is allocation control. The iterator avoids retaining the full result set, but each row is still scanned into a fresh result value. Dynamic fields such as strings, byte slices, or driver-specific values may require additional allocations, which are only reclaimed when the GC runs. The iterator therefore removes the growing `[]T`, but does not by itself guarantee a constant memory footprint or eliminate allocation and GC pressure.

It would be valuable for the streaming API to give callers some way to control reuse of the scan destination. A synchronous consumer could reuse a single destination, while a consumer with multiple results in flight could obtain destinations from a `sync.Pool`. The important part is that ownership and lifetime remain explicit, so a destination is not reused while the caller may still retain it.

If that can be supported cleanly by the iterator API, it may also cover the allocation-control use case discussed in #4108, without requiring separate low-level query and scan helpers. If it cannot be supported safely within the iterator abstraction, then the lower-level approach from #4108 would still be useful alongside the iterator.

I also noticed that `:many:stream` is implemented by adding a dedicated `Stream` boolean and threading it through the parser and metadata system. Since `stream` is effectively a modifier on `:many`, would it make sense to represent modifiers generically, for example as a map or set, instead of adding a new field for each one? That would allow future modifiers to reuse the existing `:many` path without requiring another hard-coded command variant and boolean throughout the system.

james-chroma:
> Thanks for working on this. One limitation I think is worth considering is allocation control. The iterator avoids retaining the full result set, but each row is still scanned into a fresh result value. Dynamic fields such as strings, byte slices, or driver-specific values may require additional allocations, which are only reclaimed when the GC runs. The iterator therefore removes the growing `[]T`, but does not by itself guarantee a constant memory footprint or eliminate allocation and GC pressure.
> 
I don't see that as an issue. I prefer this approach to the lower-level approach that you have proposed. That approach results in needing to create more boiler-plate code, which can be done more simply just by being able to call `QueryContext` directly. I agree with the [this comment](https://github.com/sqlc-dev/sqlc/issues/4108#issuecomment-3545008023), that the lower-level approach could easily lead to implementation errors.

kyleconroy:
If we're going to discuss the iterator approach, please keep the discussion on the #720 issue. I have no plans on merging this specific PR, as it was mainly a proof-of-concept

PR #4373, opened 2026-04-14
fix(sqlite): non-ASCII in SQL comments corrupts generated queries

exploded:
## Summary

- **Root cause:** ANTLR4's `InputStream` operates on characters (runes), so token positions are character indices. `source.Pluck()` slices Go strings with byte offsets. When multi-byte UTF-8 characters (e.g. em-dash U+2014, 3 bytes) appear in SQL comments, the offset mismatch causes query text to be extracted at wrong positions -- truncating `?` parameter placeholders and leaking comment text into generated Go code.
- **Fix:** Build a rune-to-byte offset lookup table in the SQLite parser and convert ANTLR character indices to byte offsets before storing `StmtLocation`/`StmtLen`.
- **Regression test:** New end-to-end test `sqlite_unicode_comment` reproducing the exact scenario from the issue (em-dash in a comment between two queries).

Fixes #4372

## Test plan

- [x] New end-to-end test `sqlite_unicode_comment` passes
- [x] All existing SQLite end-to-end tests pass (`sqlite_skip_todo`, `sqlite_table_options`)
- [x] SQLite engine unit tests pass
- [x] Full `TestReplay/base` suite passes (only pre-existing `wasm_plugin_sqlc_gen_unsafe_paths` failure, unrelated)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4492, opened 2026-06-22
fix(sqlite): translate antlr rune offsets to byte offsets

abgoyal:
## Summary

Fixes a bug where SQLite code generation produced truncated or corrupt
output whenever the source `.sql` file contained a non-ASCII character
(em-dash, accented letters, etc.).

## The bug

Given a query with a multi-byte character in a comment (here the `é` in
"café", which is 2 bytes in UTF-8):

```sql
-- name: GetItem :one
-- Lookup used by the café menu screen
SELECT id, name FROM items WHERE id = ?;
```

sqlc generated a truncated SQL constant, dropping the trailing `?`:

```go
const getItem = `-- name: GetItem :one
SELECT id, name FROM items WHERE id =
`
```

In queries that use `sqlc.arg()` or `SELECT *`, the misaligned offsets
caused edits to land at the wrong byte positions and produced SQL that
fails to parse (e.g. `SEid ...`). There was no error at generate time —
the corruption was silent.

## Root cause

antlr's `InputStream` stores the input as a `[]rune` and reports all
token positions as **rune indices**. The rest of sqlc, shared with the
Postgres and MySQL engines, treats query offsets as **byte offsets**
into the original source:

- `source.Pluck` slices the raw SQL for each statement
- `source.Mutate` applies parameter / star-expansion edits

For ASCII, rune index == byte offset, so the bug never surfaced. A
multi-byte rune (em-dash = 3 bytes, 1 rune) makes every later rune
index lag its true byte offset, truncating the statement slice and
shifting every edit. Postgres and MySQL are unaffected because they get
byte offsets from libpg_query and TiDB.

## The fix

Confined to the SQLite engine. We build a rune-index → byte-offset table
for the source and translate offsets at the point they are produced, so
a rune-based offset never escapes the parser:

- `parse.go` converts the statement's `StmtLocation` / `StmtLen` and
  threads the table into the converter (`cc.convertPos`).
- `convert.go` routes every node `Location` through a single `cc.pos`
  helper instead of using the raw antlr token offset.

By the time the AST leaves the parser, all offsets are byte offsets,
matching the invariant the other engines already satisfy. The
translation lives where the offsets originate, so there is no separate
post-processing pass and no reflection.

## Testing

- New regression test `internal/engine/sqlite/parse_test.go` (fails
  before the fix, passes after).
- Verified `sqlc generate` end-to-end on the query above and on a harder
  case (multi-byte string literal before `sqlc.arg()` params +
  `SELECT *`); both are correct after the fix, and the latter produced
  corrupt SQL before it.

### DIFF
--- a/internal/engine/sqlite/convert.go
+++ b/internal/engine/sqlite/convert.go
@@ -14,6 +14,25 @@ import (
 
 type cc struct {
 	paramCount int
+
+	// convertPos maps a rune index (as reported by antlr) to a byte offset in
+	// the original source. The compiler treats all node locations as byte
+	// offsets, so every Location must be routed through pos. May be nil in
+	// tests, in which case offsets are passed through unchanged.
+	convertPos func(int) int
+}
+
+// pos translates an antlr rune offset into the byte offset the rest of sqlc
+// expects when slicing the source SQL (see source.Pluck / source.Mutate).
+func (c *cc) pos(tok antlr.Token) int {
+	if tok == nil {
+		return 0
+	}
+	start := tok.GetStart()
+	if c.convertPos == nil {
+		return start
+	}
+	return c.convertPos(start)
 }
 
 type node interface {
@@ -286,7 +305,7 @@ func (c *cc) convertFuncContext(n *parser.Expr_functionContext) ast.Node {
 		if funcName == "coalesce" {
 			return &ast.CoalesceExpr{
 				Args:     args,
-				Location: name.GetStart().GetStart(),
+				Location: c.pos(name.GetStart()),
 			}
 		} else {
 			return &ast.FuncCall{
@@ -303,7 +322,7 @@ func (c *cc) convertFuncContext(n *parser.Expr_functionContext) ast.Node {
 				Args:        args,
 				AggOrder:    &ast.List{},
 				AggDistinct: n.DISTINCT_() != nil,
-				Location:    name.GetStart().GetStart(),
+				Location:    c.pos(name.GetStart()),
 			}
 		}
 	}
@@ -334,7 +353,7 @@ func (c *cc) convertColumnNameExpr(n *parser.Expr_qualified_column_nameContext) 
 		Fields: &ast.List{
 			Items: items,
 		},
-		Location: n.GetStart().GetStart(),
+		Location: c.pos(n.GetStart()),
 	}
 }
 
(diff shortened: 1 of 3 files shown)


Nthalk:
Carrying the same fix in a fork, and this PR matches what we ended up with. Two
notes from reviewing it.

The test doesn't cover the convert.go half. Revert just this line in parse.go:

    converter := &cc{convertPos: runeToByte}   ->   converter := &cc{}

and TestParseNonASCIIOffsets still passes. It only asserts through source.Pluck,
which exercises StmtLocation/StmtLen. So it can't tell this PR apart from #4314
and #4373, which patch parse.go only.

Those two are incomplete, which is the other note: node Locations stay rune
offsets while StmtLocation becomes bytes, and expand.go:195 does
res.Location - raw.StmtLocation. Mixed units.

The convert.go half is also where the silent failure lives. Truncation breaks
loudly. A misplaced star expansion can still parse, so sqlc exits 0 and the
generated Go only fails at runtime:

    SELECT 'café' AS tag, * FROM item;
    SELECT 'café' AS tag,id, name* FROM ite

That reads ResTarget.Location.

I have a test that catches it. It checks each location by slicing the source at
it and comparing the token, over six cases including the one above and a
multibyte char in an earlier statement. Passes on your branch, fails 5 subtests
with convertPos nil. Want it as a PR against your branch?

MarkRosemaker:
This is great, thank you for working on this! As another suggested, I tweaked the test you wrote to catch truncation in a second query because the shift accumulates (so its window begins inside the previous one on main).


```go
func TestParseNonASCIIOffsets(t *testing.T) {
	const query = "-- name: GetItem :one\n" +
		"-- Returns an item — looks up by id\n" +
		"SELECT id, name FROM items WHERE id = ?;\n" +
		"\n" +
		"-- name: ListItems :many\n" +
		"SELECT id, name FROM items ORDER BY name;\n"

	p := NewParser()
	stmts, err := p.Parse(strings.NewReader(query))
	if err != nil {
		t.Fatal(err)
	}
	if len(stmts) != 2 {
		t.Fatalf("expected 2 statements, got %d", len(stmts))
	}

	want := []string{"WHERE id = ?", "ORDER BY name"}
	for i, stmt := range stmts {
		raw := stmt.Raw
		plucked, err := source.Pluck(query, raw.StmtLocation, raw.StmtLen)
		if err != nil {
			t.Fatalf("statement %d: %v", i, err)
		}
		if !strings.Contains(plucked, want[i]) {
			t.Errorf("statement %d plucked query was truncated:\n%q\nmissing %q", i, plucked, want[i])
		}
	}
}
```

Nthalk:
Nice. Why oh why can't we have nice things?

PR #4511, opened 2026-07-12
docs: fix SQLite tutorial CreateAuthor usage with RETURNING *

### LINKED ISSUE #4468: SQLite getting-started tutorial uses LastInsertId() on RETURNING * result
## Description 
                              
  The SQLite getting-started tutorial appears to have an inconsistency between the SQL query and the Go usage example.

  Page:                                                     

  https://docs.sqlc.dev/en/latest/tutorials/getting-started-sqlite.html

  The tutorial defines `CreateAuthor` as:                                                                                

  ```sql            
  -- name: CreateAuthor :one
  INSERT INTO authors (                                     
    name, bio
  ) VALUES (           
    ?, ?
  )                                                                                                                      
  RETURNING *;
```
                              
  With :one and RETURNING *, sqlc generates a method returning an Author:

```go
                                                                                                                         
  func (q *Queries) CreateAuthor(ctx context.Context, arg CreateAuthorParams) (Author, error)
```

  However, the later Go example uses the returned value as if it were a sql.Result:                                      
                               
```go                                                                                          
  result, err := queries.CreateAuthor(ctx, tutorial.CreateAuthorParams{                                                  
      Name: "Brian Kernighan",                              
      Bio:  sql.NullString{String: "Co-author of The C Programming Language and The Go Programming Language", Valid:
      true},
  })

  insertedAuthorID, err := result.LastInsertId()
```
  This does not compile because result is an Author, not a sql.Result.

 ## Expected behavior

  The tutorial should use the returned Author directly, for example:
```go
  createdAuthor, err := queries.CreateAuthor(ctx, tutorial.CreateAuthorParams{
      Name: "Brian Kernighan",
      Bio:  sql.NullString{String: "Co-author of The C Programming Language and The Go Programming Language", Valid:
      true},
  })
  if err != nil {
      return err
  }
  log.Println(createdAuthor.ID)
```

  Then the later fetch could use:

```go
  fetchedAuthor, err := queries.GetAuthor(ctx, createdAuthor.ID)
```
  ## Actual behavior

  The tutorial currently calls:
```go
  result.LastInsertId()
```

  on the value returned by CreateAuthor, but that value is generated as an Author.

  ## Version

  I reproduced this with sqlc v1.31.1, which is also the version shown in generated files from the current tutorial.

Abhineshhh:
## Summary

- The SQLite getting-started tutorial defines `CreateAuthor` as `:one` with `RETURNING *`, so sqlc generates `(Author, error)`, not `sql.Result`.
- The Go example incorrectly treated the return value as `sql.Result` and called `LastInsertId()`, which does not compile.
- Update the example to use the returned `Author` (same pattern as the PostgreSQL tutorial) and compare full structs after `GetAuthor`.

Fixes #4468

## Verification

1. Generated code from the tutorial `schema.sql` / `query.sql` with a local `sqlc` build:
   - `CreateAuthor(ctx, CreateAuthorParams) (Author, error)`
   - `GetAuthor(ctx, id int64) (Author, error)`
   - `Author{ID int64; Name string; Bio sql.NullString}`
2. Compiled and ran the fixed `tutorial.go` end-to-end with `modernc.org/sqlite` and an in-memory DB.
3. Program output included `true` from `reflect.DeepEqual(insertedAuthor, fetchedAuthor)`.

## Test plan

- [x] Confirmed generated signatures match the fixed example
- [x] `go build` and run of the full tutorial program succeeds
- [ ] Docs render as expected after merge

### DIFF
--- a/docs/tutorials/getting-started-sqlite.md
+++ b/docs/tutorials/getting-started-sqlite.md
@@ -159,28 +159,23 @@ func run() error {
 	log.Println(authors)
 
 	// create an author
-	result, err := queries.CreateAuthor(ctx, tutorial.CreateAuthorParams{
+	insertedAuthor, err := queries.CreateAuthor(ctx, tutorial.CreateAuthorParams{
 		Name: "Brian Kernighan",
 		Bio:  sql.NullString{String: "Co-author of The C Programming Language and The Go Programming Language", Valid: true},
 	})
 	if err != nil {
 		return err
 	}
-
-	insertedAuthorID, err := result.LastInsertId()
-	if err != nil {
-		return err
-	}
-	log.Println(insertedAuthorID)
+	log.Println(insertedAuthor)
 
 	// get the author we just inserted
-	fetchedAuthor, err := queries.GetAuthor(ctx, insertedAuthorID)
+	fetchedAuthor, err := queries.GetAuthor(ctx, insertedAuthor.ID)
 	if err != nil {
 		return err
 	}
 
 	// prints true
-	log.Println(reflect.DeepEqual(insertedAuthorID, fetchedAuthor.ID))
+	log.Println(reflect.DeepEqual(insertedAuthor, fetchedAuthor))
 	return nil
 }
 


Abhineshhh:
have a review @kyleconroy

kyleconroy (approved):

kyleconroy:
Thanks!

PR #4514, opened 2026-07-13
Local PostgreSQL container for query analysis / validation

nikolayk812:
Add a testcontainers_image field to the database config block. When set, sqlc automatically starts a local PostgreSQL Docker container, applies schema migrations, and uses it for analysis — then tears it down when done.

```
database:
      testcontainers_image: "postgres:18-alpine"
```

https://github.com/sqlc-dev/sqlc/issues/4513


davideimola:
I really like the idea of using containers in development!

internal/config/config.go
```diff
@@ -74,8 +74,9 @@ type Server struct {
 }
 
 type Database struct {
-	URI     string `json:"uri" yaml:"uri"`
-	Managed bool   `json:"managed" yaml:"managed"`
+	URI                 string `json:"uri" yaml:"uri"`
+	Managed             bool   `json:"managed" yaml:"managed"`
+	TestcontainersImage string `json:"testcontainers_image" yaml:"testcontainers_image"`
```
ayakovlenko, line 79: Does it have to be called `testcontainers_image`? Why not just `image` or `container_image` to not leak implementation details about the library used to the public API?

kyleconroy:
@nikolayk812 sorry, I'm not interested in adding Docker / container support at this time.

nikolayk812:
@kyleconroy fair enough. I got a feeling that in-memory mode of sqlc should be deprecated or discouraged. In-memory supports results in reimplementing lots of semantical validations in the generator, which DB can do automatically. 

But then current DB mode is limited, in a way that it requires migrations to be applied separately, outside of the generator if I understand correctly. Is this to give advantage to SQLc Cloud?

PR #4518, opened 2026-07-18
feat(pgx-v5): Add static typing support for jsonb_build_object

m-ali-akbay:
Adds a `sqlc.jsonb_build_object."Name"(key, value, ...)` directive for returning JSON-shaped results that decode straight into a Go struct. It's handy for pulling a one-to-many relationship into a single query instead of running one query per parent row.

```sql
-- name: GetAuthors :many
SELECT
  authors.id,
  ARRAY(
    SELECT sqlc.jsonb_build_object."Book"('id', books.id, 'title', books.title)
    FROM books WHERE books.author_id = authors.id
  ) AS books
FROM authors;
```

gives you:

```go
type GetAuthorsRow struct {
	ID    int64
	Books []Book
}

type Book struct {
	ID    int64  `json:"id"`
	Title string `json:"title"`
}
```

A few notes:

- **pgx/v5 only** — `generate` fails with a clear error for any other `sql_package`. pgx decodes the `jsonb` value into the struct (or `[]struct` inside `ARRAY(...)`) directly, so there's no `Scan`/`Value` plumbing.
- The struct name is required and read off the qualified call — Postgres parses `sqlc.jsonb_build_object."Book"` as `catalog.schema.name`, so no new parser work.
- Nesting works: a value can be another `jsonb_build_object` or an `ARRAY(...)` of one.
- Two queries that use the same name share one Go type when their shapes match; a shape mismatch or a collision with a model fails generation. Names and fields can be overridden with `rename`.
- Internally the call is just rewritten to Postgres's `jsonb_build_object` before analysis.

Docs in `docs/howto/embedding.md`; covered by unit tests plus a `sqlc_json` golden fixture.


m-ali-akbay:
@kyleconroy, may I ask for a review? 🙏

PR #4521, opened 2026-07-22
clickhouse: type inference via the new analysis core (analyze only)

kyleconroy:
## Motivation

sqlc already has a working ClickHouse **parser** (`internal/engine/clickhouse`, backed by `doubleclick`), but no **analysis** for it: the ClickHouse catalog was a stub, the engine was not wired into the compiler, and there was no `EngineClickHouse`. The missing layer is a typed catalog + a query analyzer.

This PR lands a new analysis core in sqlc and makes **ClickHouse the first engine to run on it**. ClickHouse is a natural first target because it has a parser and zero existing analysis — this is net-new, with no regression surface for the other engines.

Scope is deliberately limited to **`sqlc analyze`**. There is no code generation for ClickHouse in this PR: the goal is to exercise the catalog and analyzer end to end, and prove the type inference, before committing to a generated-code surface. This follows the same shape as #4522, which wires GoogleSQL into `analyze` only.

## What's here

`sqlc analyze --dialect clickhouse` resolves result column types entirely through the new core.

```console
$ sqlc analyze --dialect clickhouse --schema schema.sql query.sql
```

Given `id UInt64, name String, tag Nullable(String), amount Float64, created DateTime`:

```json
[
  {
    "name": "ListEvents",
    "cmd": ":many",
    "columns": [
      { "name": "id",      "data_type": "uint64",   "not_null": true,  "table": "events" },
      { "name": "name",    "data_type": "string",   "not_null": true,  "table": "events" },
      { "name": "tag",     "data_type": "string",   "not_null": false, "table": "events" },
      { "name": "amount",  "data_type": "float64",  "not_null": true,  "table": "events" },
      { "name": "created", "data_type": "datetime", "not_null": true,  "table": "events" }
    ],
    "params": []
  }
]
```

Note `tag`: `Nullable(String)` correctly unwraps to a nullable `string`.

**The core** (`internal/core/`)
- A SQLite-backed catalog whose `sql_*` tables are modeled on the Postgres system catalogs (namespaces, classes, attributes, types, casts, operators, procs).
- `internal/core/analyzer/` — a dialect-neutral analyzer that resolves projections and expression types against that catalog.
- `internal/core/schema/` — engine-neutral DDL application (`CREATE TABLE` / `DROP TABLE` → catalog).

**sqlc builds sqlc.** Every query the catalog runs against its own `sql_*` tables is generated by sqlc's SQLite engine from `internal/core/catalogdef/{schema.sql,query.sql}` into `internal/core/catalogdb`, via `go:generate`. It is idempotent, offline, and introduces no build-time cycle — the SQLite engine uses the legacy compiler, so there is no analysis recursion.

**ClickHouse**
- `seed.go` — the ClickHouse dialect and its built-in types.
- The converter now renders nested type parameters (`Nullable(String)`, `Array(UInt64)`, `Decimal(18,4)`) instead of dropping them as `TODO`, and unwraps them to a canonical scalar plus array/nullable flags so the core catalog only ever sees scalars.
- The parser computes statement byte-spans (doubleclick reports statement starts but not ends), so `-- name:` annotations fall inside each statement.

**Compiler**
- `config`: the `clickhouse` engine constant.
- `NewCompiler` builds a core catalog seeded with the ClickHouse dialect; `parseCatalog` applies schema DDL to it; `parse_core.go` resolves columns via `core/analyzer` and assembles `*compiler.Query`, reusing only the shared query-metadata parsing.

**Tests** — `analyze_basic/clickhouse` covers the pipeline end to end, matching how the other engines are tested. Per review feedback, there are no unit tests; coverage is end-to-end only.

## Design decisions

- **One AST** — sqlc's `internal/sql/ast`; the analyzer was written against it directly, not a second copy with a converter.
- **One catalog** — the core catalog is the single source of truth for ClickHouse analysis; the in-memory `internal/sql/catalog` is not used on this path.
- **No analyzer seam, no legacy analyze step** — ClickHouse never enters `analyzeQuery`/`inferQuery`/`outputColumns` or the `analyzer.Analyzer` interface.
- **Nothing engine-specific in `compiler`** beyond parser construction, which matches how the other engines are wired.

## Follow-ups

- `WHERE` clauses and bind parameters. The analyzer currently handles projecting `SELECT`s; `SELECT ... WHERE id = {id:UInt64}` still reports `a_expr: unsupported kind 0`, so the `analyze` case is scoped to what actually works.
- Query coverage beyond `SELECT` (INSERT/UPDATE/DELETE).
- Code generation: a core-catalog → `plugin.Catalog` projection and a ClickHouse → Go type map. Both existed in an earlier revision of this branch and were removed to keep this PR analyze-only; richer types (`decimal.Decimal`, `uuid.UUID`, `netip.Addr`, `json.RawMessage`) need import wiring.
- Array-ness is parsed but not yet modeled on catalog attributes.
- A `testgen`-style conformance suite against a real ClickHouse.

This is the first engine on the new core. When other engines migrate, they follow the same path and the legacy analyze machinery can be retired.

`go build ./...` and `go vet` pass; `go generate ./internal/core/...` is idempotent; the `analyze_basic/clickhouse` and `parse_basic/clickhouse` end-to-end cases are green.

https://claude.ai/code/session_01XTGxNHW6v1S1YyC9FDSgrK


internal/cmd/shim.go
```diff
@@ -224,10 +225,55 @@ func pluginQueryParam(p compiler.Parameter) *plugin.Parameter {
...
 	return &plugin.GenerateRequest{
 		Settings:    pluginSettings(r, settings),
-		Catalog:     pluginCatalog(r.Catalog),
+		Catalog:     cat,
 		Queries:     pluginQueries(r),
 		SqlcVersion: info.Version,
 	}
 }
+
+func pluginCatalogFromCore(cc *core.Catalog) *plugin.Catalog {
```
kyleconroy: Let's move this into a new core/shim package

internal/compiler/compile.go
```diff
@@ -55,6 +56,16 @@ func (c *Compiler) parseCatalog(schemas []string) error {
 			continue
 		}
 
+		if c.coreCatalog != nil {
+			for i := range stmts {
+				if err := clickhouse.Apply(c.coreCatalog, stmts[i].Raw); err != nil {
```
kyleconroy: We're eventually going to support more engines here. I don't think this should be in the clickhouse package.

internal/compiler/parse_core.go
```diff
@@ -0,0 +1,114 @@
+package compiler
```
kyleconroy, line 1: This entire file is not right. We shouldn't have any engine-specific files inside the compiler.

internal/core/analyzer/analyzer_test.go
```diff
@@ -0,0 +1,100 @@
+package analyzer_test
```
kyleconroy: Delete this entire file. We only test using end-to-end tests

internal/core/cast_test.go
```diff
@@ -0,0 +1,51 @@
+package core
```
kyleconroy: No tests like this, only end-to-end tests

internal/core/dialect_test.go
```diff
@@ -0,0 +1,42 @@
+package core
```
kyleconroy: No tests like this, only end-to-end tests

internal/core/operator_test.go
```diff
@@ -0,0 +1,46 @@
+package core
```
kyleconroy: No tests like this, only end-to-end tests

internal/core/proc_test.go
```diff
@@ -0,0 +1,67 @@
+package core
```
kyleconroy: No tests like this, only end-to-end tests

internal/engine/clickhouse/analyze_test.go
```diff
@@ -0,0 +1,94 @@
+package clickhouse_test
```
kyleconroy: No tests like this, only end-to-end. Remove

internal/cmd/shim.go
kyleconroy: Moved to a new `internal/core/shim` package as `PluginCatalog`.

---

internal/compiler/compile.go
kyleconroy: Agreed. The DDL apply now lives in a new engine-neutral `internal/core/schema` package (`schema.Apply`), and the loop here calls that. The ClickHouse-specific type unwrapping moved into the engine's column conversion, so the schema apply only sees canonical scalar types.

---

internal/compiler/parse_core.go
kyleconroy: This is no longer engine-specific — renamed to `parse_core.go`. It only handles the generic core analysis path (it builds a `compiler.Query`, so it has to live in this package to avoid an import cycle). All ClickHouse-specific logic is out of the compiler now.

---

PR #3961, opened 2025-05-13
fix: escape constant names in queries

snprajwal:
Constant names in queries were not escaped, leading to keywords being emitted as constant names in the generated output. This has been updated to be escaped, and `any` has been added as a reserved keyword as of Go 1.18.

This also addresses the lint raised by gopls on sqlc's generated Go files for modernizing the usage of an empty interface for the `any` type.

PR #4525, opened 2026-07-26
Escape constant names in queries

snprajwal:
The `any` keyword was introduced in Go 1.18, and can be used in place of `interface{}`. This was not in the list of reserved keywords, leading to identifiers using this term in generated code.

Update the logic to escape constant names, and also add `any` to the list of reserved keywords.

This is a minimal rebased version of #3961, which I'm unable to reopen. It only contains the changes that fix invalid code generation.


kyleconroy:
Please open an issue! I'm not currently accepting pull requests.

Issue #4512, opened 2026-07-13
Missing sqlc.arg in SQL UPDATE query with EXISTS subquery


paulkoehlerdev:
### Version

1.31.1

### What happened?

When running sqlc 1.31.1 with `go tool sqlc generate` I dont get any error log, but the output does not include the specified parameter. I would expect a timestamp parameter in this example.

Expectation:
```
// Code generated by sqlc. DO NOT EDIT.
// versions:
//   sqlc v1.31.1
// source: query.sql

package db

import (
	"context"
	"time"
)

const updateEntities = `-- name: UpdateEntities :execrows
UPDATE target_table
SET active = 0
WHERE EXISTS (
  SELECT 1
  FROM source_table s
  WHERE s.updated_at >= ?
)
`

func (q *Queries) UpdateEntities(ctx context.Context, updatedAfter time.Time) (int64, error) {
	result, err := q.db.ExecContext(ctx, updateEntities, updatedAfter)
	if err != nil {
		return 0, err
	}
	return result.RowsAffected()
}
```

Actual output:
```
// Code generated by sqlc. DO NOT EDIT.
// versions:
//   sqlc v1.25.0
// source: query.sql

package db

import (
	"context"
)

const updateEntities = `-- name: UpdateEntities :execrows
UPDATE target_table
SET active = 0
WHERE NOT (
    EXISTS (
        SELECT 1
        FROM source_table s
        WHERE s.updated_at >= sqlc.arg(updated_after)
    )
)
`

func (q *Queries) UpdateEntities(ctx context.Context) (int64, error) {
	result, err := q.db.ExecContext(ctx, updateEntities)
	if err != nil {
		return 0, err
	}
	return result.RowsAffected()
}
```

### Relevant log output

```shell
-- no output --
```

### Database schema

```sql
CREATE TABLE target_table
(
    id     BIGINT PRIMARY KEY,
    active BOOLEAN NOT NULL
);

CREATE TABLE source_table
(
    id         BIGINT PRIMARY KEY,
    updated_at TIMESTAMP NOT NULL
);
```

### SQL queries

```sql
-- name: UpdateEntities :execrows
UPDATE target_table
SET active = 0
WHERE NOT (
    EXISTS (
        SELECT 1
        FROM source_table s
        WHERE s.updated_at >= sqlc.arg(updated_after)
    )
);
```

### Configuration

```yaml
{
  "version": "2",
  "sql": [{
    "schema": "schema.sql",
[...]

PR #4526, opened 2026-07-27
fix(dolphin): preserve NOT EXISTS expressions

muratmirgun:
Fix MySQL AST conversion dropping NOT from NOT EXISTS expressions.

the affected syntax in two different ways:
- NOT (EXISTS (...)) as a unary NOT operation
- NOT EXISTS (...) through ExistsSubqueryExpr.Not

The Dolphin converter did not handle either representation correctly. In the parenthesized case, the entire WHERE expression could be dropped, including nested sqlc.arg parameters. In the direct case, the query was preserved but the NOT operator was lost.

This change:

- converts unary logical NOT expressions to ast.BoolExpr
- preserves ExistsSubqueryExpr.Not
- adds regression coverage for both forms
- verifies that nested named parameters remain present
#4512

### DIFF
--- a/internal/engine/dolphin/convert.go
+++ b/internal/engine/dolphin/convert.go
@@ -397,12 +397,21 @@ func (c *cc) convertRenameTableStmt(n *pcast.RenameTableStmt) ast.Node {
 	return list
 }
 
-func (c *cc) convertExistsSubqueryExpr(n *pcast.ExistsSubqueryExpr) *ast.SubLink {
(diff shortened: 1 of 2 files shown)


kyleconroy:
Can you please open an issue instead? I'm not accepting pull requests at the moment. Include the full schema and SQL queries that don't work.

muratmirgun:
Hello @kyleconroy reference issue here https://github.com/sqlc-dev/sqlc/issues/4512

PR #4541, opened 2026-08-02
docs: add uv/pip installation instructions

rayakame:
Adds a short section to the install docs for installing sqlc from PyPI via [sqlc-bin](https://pypi.org/project/sqlc-bin/). The package ships the unmodified official release binaries, so Python projects can pin sqlc as a dev dependency without needing a Go toolchain. I maintain the sqlc-bin package.

### DIFF
--- a/docs/overview/install.md
+++ b/docs/overview/install.md
@@ -22,6 +22,18 @@ Installing recent versions of sqlc requires Go 1.21+.
 go install github.com/sqlc-dev/sqlc/cmd/sqlc@latest
 ```
 
+## uv / pip
+
+sqlc is available on PyPI as [sqlc-bin](https://pypi.org/project/sqlc-bin/), which packages the official release binaries. No Go toolchain is required.
+
+```
+uv tool install sqlc-bin
+```
+
+```
+pip install sqlc-bin
+```
+
 ## Docker
 
 ```


kyleconroy:
I don't support pypi as an official distribution channel. Are you the maintainer of sqlc-bin? Could you either remove it or change the name? `sqlc-bin` makes it sounds like it's affiliated with the project.

rayakame:
Yes, I maintain `sqlc-bin`. I wanted to have a way for people to install sqlc without needing the go toolchain installed, and being able to pin a version of sqlc.

I can understand that you don't support Pypi as an official distribution channel, but I don't think `sqlc-bin` implies direct affiliation with the project. I intentionally didn't name it just `sqlc` because of that. Its already a widespread precedent for community-provided binaries to have the same name, purely to avoid confusion. https://github.com/njzjz/nodejs-wheel for example. 

Ive updated the README or any other documentation to make it clear that its an unofficial project and not affiliated with sqlc and its developers. If you have a name recommendation with which you are more comfortable, please let me know, I'm still open to it.

kyleconroy:
Thanks for updating the README. I think you're right that `sqlc-bin` is different enough. The nodejs-wheel package is a good example.

rayakame:
Yeah of course no problem, and if there ever is anything else, you can always contact me.

Just so you know, while changing the README i came across this:

https://pypi.org/project/sqlc/
https://github.com/Flamefork/sqlc-py

Which is a similar package to mine, instead of just downloading the release binaries it builds them locally. 

It is installed by doing `pip install sqlc`

PR #4544, opened 2026-08-06
Add coreanalyzer experiment routing generate through the analysis core

kyleconroy:
## Summary

`SQLCEXPERIMENT=coreanalyzer` routes `sqlc generate` through the core catalog and analyzer — the same path `sqlc analyze` and the ClickHouse and GoogleSQL engines already use — instead of each engine&#39;s own analysis path.

- **New experiment** in `internal/opts`: `coreanalyzer` (`no` prefix to disable, case-insensitive). When set, `parse()` builds each query set&#39;s compiler with `compiler.WithCoreAnalysis()`.
- **Experiment wiring**: `Env.Experiment` was parsed from the environment but never reached the compiler — `processQuerySets` always built empty parser options. It now threads the experiment through.
- **Codegen through the core**: a core-mode `Result` carried a nil catalog, which `pluginCatalog` would dereference — generate had never run on this path (ClickHouse/GoogleSQL only have `parse`/`analyze` cases). `catalog_core.go` now dumps the core catalog into the legacy catalog shape after the schema is applied, so codegen builds models as usual. Seeded `pg_catalog`/`information_schema` relations come along but codegen already skips those schemas.
- **Array fix**: codegen renders one `[]` per `ArrayDims`, which the core path never set — array columns silently generated as scalars (`Tags string` instead of `Tags []string`, no `pq.Array` scan). Both the catalog dump and core query columns now set `ArrayDims: 1` when the core reports an array.

Output through the core matches the legacy path exactly for the covered cases, except `SELECT *` / `RETURNING *` stay literal in the emitted SQL: the core takes columns from analysis rather than rewriting the query text.

Also documents the experiment in `docs/reference/environment-variables.md`, replacing the stale &#34;no experiments are defined&#34; note.

Rebased on main through #4543 (extension catalog seed files). #4545 removed the analyzerv2 experiment, so `coreanalyzer` is now the only one and the wiring passes `Env.Experiment` through unscoped.

## Testing

- New end-to-end case `internal/endtoend/testdata/experiment_coreanalyzer/` with committed goldens for postgresql (stdlib), sqlite, and mysql, run via `exec.json` env `SQLCEXPERIMENT=coreanalyzer`.
- Experiment parsing cases added to the existing `internal/opts` tests.
- `go test ./...` green on the rebased branch with PostgreSQL and MySQL live, including the full `endtoend` package and the three `experiment_coreanalyzer` replay cases.

https://claude.ai/code/session_01ANVHEs41RJbybS3ja9N1SA


internal/cmd/process.go
```diff
@@ -86,7 +86,15 @@ func processQuerySets(ctx context.Context, rp ResultProcessor, conf *config.Conf
 			sql.Queries = joined
 
 			var name, lang string
-			parseOpts := opts.Parser{}
+			parseOpts := opts.Parser{
+				// Only the coreanalyzer experiment reaches the compiler.
+				// analyzerv2 is intentionally left behind: the goldens of the
+				// tests that name it predate this wiring and pin the ordinary
+				// analyzer path.
```
kyleconroy: Remove this comment

kyleconroy:
The `buf / build` failure on 96b5207 is a transient rate limit, not something in this diff. It failed during setup, before any buf command ran:

```
Setting up buf version "1.50.0"
Resolving the download URL for the current platform...
##[error]API rate limit exceeded for 172.183.131.23.
```

`bufbuild/buf-setup-action@v1` resolves the buf release URL through the GitHub API, and `.github/workflows/buf.yml` passes it no token, so the request is unauthenticated and subject to per-IP limiting on shared runners — the job logs the action's own warning, `No github_token supplied, API requests will be subject to stricter rate limiting`. This diff touches no `.proto` or buf files at all, and run 2271 is the only buf failure in the last ten runs; run 2268, on this same branch, passed.

It needs a re-run, which I can't trigger — the app token here has no `actions: write`. The durable fix would be adding `github_token: ${{ github.token }}` to the buf-setup step, but that's an unrelated CI change I'd rather not land inside this PR. Happy to send it separately if you want it.

---

Issue #3585, opened 2024-09-06
Resolving references to CTE Columns fails


simonklee:
### Version

1.27.0

### What happened?

Resolving references to multiple CTE's with arguments fails.

This is a duplicate of #3219, but 3219 has been closed even thought the issue still persists and the relevant PR is not merged.

PR to fix this issue: https://github.com/sqlc-dev/sqlc/pull/3220

### Relevant log output

```shell
sqlc generate failed.
# package 
query.sql:5:28: table alias "t1_ids" does not exist
```

### Database schema

```sql
CREATE TABLE t1
(
    id SERIAL NOT NULL PRIMARY KEY
);
```

### SQL queries

```sql
-- name: CTERef :one
2
WITH t1_ids AS (
3
    SELECT id FROM t1
4
)
5
SELECT * FROM t1_ids WHERE t1_ids.id = sqlc.arg('id');
```

### Configuration

```yaml
{
2
  "version": "2",
3
  "sql": [{
4
    "schema": "schema.sql",
5
    "queries": "query.sql",
6
    "engine": "postgresql",
7
    "gen": {
8
      "go": {
9
        "out": "db"
10
      }
11
    }
12
  }]
13
}
14
​
```

### Playground URL

https://play.sqlc.dev/p/9048b486f822c300301650867612e42bef8f8d512dfd12d84ad3a4cd210f4ef7

### What operating system are you using?

Linux

### What database engines are you using?

PostgreSQL

### What type of code are you generating?

Go

michaelgrijalva-resim:
If anyone else runs into this issue and lands on this page, I found it usually works if you explicitly annotate the appropriate type on the sqlc argument:

```sql
-- name: CTERef :one
WITH t1_ids AS (
    SELECT id FROM t1
)
SELECT * FROM t1_ids WHERE t1_ids.id = sqlc.arg('id')::uuid;
--                                                   ^^^^^^
```

This makes sqlc happy for now, until the above fix is merged.

mem:
I have a similar issue, but the workaround suggested in this case does not work.

https://play.sqlc.dev/p/3df2d3f54eff7ef74e32ffca38be69ddd9a3dc7058cbe7ae9274007e05909236

The SQL in the link is silly, as I have reduced the case as much as possible.

The actual query is trying to retrieve a single record with some conditions and insert a modified version of the same record using input provided by the user.

ednmechrock:
[...]

PR #4586, opened 2026-08-27
Fix parameter inference for aliased CTE columns

riordanpawley:
## Summary

Parameter inference indexed catalog tables but skipped common-table expressions. As a result, a bind parameter compared with an aliased CTE column failed compilation with `column "activity_at" does not exist`, even though output-column analysis had already resolved the CTE.

This change indexes the resolved CTE columns in the same parameter-inference map used for catalog tables. The end-to-end fixture is the minimal reproduction: one base column renamed in a CTE and referenced beside `$1` in an aggregate `FILTER`.

## User-facing changes

sqlc now infers parameters compared with aliased CTE columns instead of rejecting the query as an unknown column.

Existing CTE update parameters now use their resolved non-null column types instead of nullable fallback wrappers; the corresponding golden output is updated.

## Validation

- Confirmed the fixture fails on unmodified `main` with `column "activity_at" does not exist`.
- `go test ./internal/endtoend -run 'TestReplay/base/cte_filter_alias' -count=1`
- `go test ./internal/compiler ./internal/endtoend -run 'TestReplay/base/cte_filter_alias' -count=1`
- Regenerated the existing `cte_update` fixture to record its newly inferred non-null parameter types.

## Risk

Low. The new indexing path runs only when a range variable resolves as a CTE rather than a catalog table. It carries the CTE's already-resolved names and types into existing parameter inference. Callers generated from affected CTE updates may move from nullable wrappers to scalar parameters when the source columns are non-null.


riordanpawley:
Closing as a duplicate. #4430 already fixes CTE-column lookup in parameter inference and tracks the ambiguity concern with sharing the regular-table type map; #3585 is the existing issue. I should have found those before opening this PR.

PR #4587, opened 2026-08-27
Upgrade to oliphant 0.2.0 and wire up sqlc fmt for PostgreSQL

kyleconroy:
Upgrades oliphant from v0.1.0 to v0.2.0 and turns on `sqlc fmt` for the `postgresql` engine.

## What's in oliphant 0.2.0

The release adds `parser.ParseFile`, which returns the parse tree together with the comment tokens the scanner already produces — from the same pass the parse makes over the input. That is exactly the contract `sqlc fmt` requires from an engine: statements and comments in one lexer pass, positioned in the same byte offsets the tree's location fields carry.

## Changes

- **`internal/engine/postgresql/parse.go`**: the PostgreSQL parser now implements `ParseFile(io.Reader) (*ast.File, error)`, mirroring the SQLite engine's meyer-based template. oliphant's comment tokens become `ast.Comment` values (text, byte offsets, own-line flag), and the existing statement-span conventions — leading comments covered, pre-PG18 location semantics — are untouched. `Parse` now delegates to `ParseFile`.
- **`internal/cmd/fmt.go`**: `newQueryFormatter` returns the PostgreSQL parser for `postgresql` configs. The engine already implemented `format.Dialect`, so `ParseFile` was the only missing piece.
- **Fingerprint belt**: PostgreSQL brings a proof the other engines cannot offer — oliphant's pg_query-compatible fingerprint. fmt now recognizes an optional `fingerprinter` interface: where an engine implements `Fingerprint`, a formatted statement is only accepted when its fingerprint matches the original's, and falls back to the text as written otherwise. This sits on top of the existing reparse / comment-multiset / fixed-point checks and makes the documented "proven to survive formatting unchanged" contract literal for PostgreSQL.
- **Test data**: the `fmt/postgresql` end-to-end case drops its "does not yet support the postgresql engine" `stderr.txt`; `stdout.txt` now commits the real `fmt --diff` golden output. Trailing and interior comments stay anchored (`-- the primary lookup` stays on the `LIMIT 1;` line, `-- the primary key` stays with its select-list item), and the deliberately long one-line `SearchAuthors` stays on one line, gofmt-style.

## Testing

- `TestReplay/base/fmt/{postgresql,mysql,sqlite}` pass.
- `TestFormat` passes — it round-trips every PostgreSQL testdata case through the printer and compares pg_query fingerprints.
- Full `go test --tags=examples -timeout 20m ./...` passes against live PostgreSQL and MySQL.

https://claude.ai/code/session_01K8CZxjsrAjJsvGfZNrR7Gd

---
_Generated by [Claude Code](https://claude.ai/code/session_01K8CZxjsrAjJsvGfZNrR7Gd)_

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -22,7 +22,7 @@ require (
 	github.com/sqlc-dev/doubleclick v1.0.0
 	github.com/sqlc-dev/marino v0.3.0
 	github.com/sqlc-dev/meyer v0.1.2
-	github.com/sqlc-dev/oliphant v0.1.0
+	github.com/sqlc-dev/oliphant v0.2.0
 	github.com/sqlc-dev/teesql v1.1.0
 	github.com/sqlc-dev/zetajones v0.1.0
 	github.com/tetratelabs/wazero v1.12.0
--- a/go.sum
+++ b/go.sum
@@ -73,8 +73,8 @@ github.com/sqlc-dev/marino v0.3.0 h1:e9cinBXJFFa3yRpokYNNinWvkewgd6XVDgnlG0bOmTw
 github.com/sqlc-dev/marino v0.3.0/go.mod h1:mQxC2dgDE0DWHMb2B5jZNk7KToJuS6wnxnffBfYnq08=
 github.com/sqlc-dev/meyer v0.1.2 h1:40Ng9Glnx7CTf3yOYV7jfvDIN7voIFjrStkIULS+uws=
 github.com/sqlc-dev/meyer v0.1.2/go.mod h1:pS4USCRf/SLjWtaMcnTo4YrEEFKBj8CyyqlxcVUJQH8=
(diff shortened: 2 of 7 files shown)


kyleconroy:
CI status on 71a6e6f: the only red check is **build windows/amd64**, and the failure is not this PR's — the job died during `go mod download` with a transient Go module proxy error (`proxy.golang.org … stream ID 129; INTERNAL_ERROR; received from peer` fetching `github.com/ncruces/go-sqlite3-wasm/v3`, a dependency this PR doesn't touch), before any code was built. The same module downloaded fine in the five sibling build jobs of the same run, and **test** (the full `--tags=examples` suite against live PostgreSQL and MySQL), **vuln_check**, and the other platform builds are all green. **build linux/arm64** shows as cancelled only because fail-fast stopped it during post-job cleanup; its `go build ./...` step had already succeeded.

I don't have permission to re-run the failed jobs from here (the API returns 403), so this needs a "Re-run failed jobs" click on the [workflow run](https://github.com/sqlc-dev/sqlc/actions/runs/33096935911) from a maintainer. No code change is needed.

---

PR #4593, opened 2026-08-28
docs(transactions): note the lost-update risk in read-modify-write

### LINKED ISSUE #3485: Error in the documentation for transactions?

shaikn6:
## What

The **Using transactions** how-to's `bumpCounter` example reads a row and then writes `counter + 1` inside a transaction:

```go
r, err := qtx.GetRecord(ctx, id)
...
qtx.UpdateRecord(ctx, tutorial.UpdateRecordParams{ID: r.ID, Counter: r.Counter + 1})
```

Under the default isolation level (`READ COMMITTED`, on both PostgreSQL and MySQL/InnoDB), two transactions running this concurrently can both read the same `counter` and both write the same `counter + 1` — a lost update. The page presents this as the standard pattern with no caveat, which is what #3485 flags.

## Change

Docs only. Adds a short **Concurrent read-modify-write** section after the examples:

- explains the lost-update window
- shows a `GetRecordForUpdate` query using `SELECT ... FOR UPDATE`
- mentions `SERIALIZABLE` + retry as the alternative, noting sqlc doesn't manage isolation levels or retries

Existing examples are unchanged — they still demonstrate `WithTx`, which is the page's purpose.

Closes #3485

### DIFF
--- a/docs/howto/transactions.md
+++ b/docs/howto/transactions.md
@@ -99,3 +99,29 @@ func bumpCounter(ctx context.Context, db *pgx.Conn, queries *tutorial.Queries, i
 	return tx.Commit(ctx)
 }
 ```
+
+## Concurrent read-modify-write
+
+The `bumpCounter` example reads a row and then writes a value derived from it.
+Under the default isolation level (`READ COMMITTED`), two transactions running
+`bumpCounter` concurrently can both read the same `counter`, both write
+`counter + 1`, and one update is lost.
+
+When a write depends on a value read earlier in the same transaction, lock the
+row on read with `SELECT ... FOR UPDATE` (supported by PostgreSQL and
+MySQL/InnoDB):
+
+```sql
+-- name: GetRecordForUpdate :one
+SELECT * FROM records
+WHERE id = $1
+FOR UPDATE;
+```
+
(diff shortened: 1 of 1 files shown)


kyleconroy:
The transaction documentation shows users how to use transactions with sqlc. It's not intended for a general discussion about the correct ways to use database transactions.

PR #4600, opened 2026-09-02
Use renamed cel.dev/cel-go package & update to 0.32.0

der-eismann:
The cel-go guys moved their repository and also changed the import path for their package: https://github.com/cel-expr/cel-go/issues/1329

I'm a bit confused by the `vendor` folder though, since it seems to be checked in, but after running `go mod tidy && go mod vendor` it created an untracked `modules.txt` and also added a bunch of packages that weren't there before... let's see how the build goes.


kyleconroy:
Thanks for the heads up. In the future, please open an issue instead of a pull request.

PR #4613, opened 2026-09-08
fix(mysql): encode copyfrom enum values as strings

### LINKED ISSUE #4324: MySQL copyfrom throws "can't encode type models.ExperienceLocationsType to TSV" if models.ExperienceLocationsType is a string enum
### Version

1.30.0

### What happened?

When trying to run a query with LOAD DATA LOCAL INFILE, mySQL throws an error "can't encode type models.ExperienceLocationsType to TSV". Said models.ExperienceLocationsType is defined as:
```
type ExperienceLocationsType string

const (
	ExperienceLocationsTypeStartPoint      ExperienceLocationsType = "start_point"
	ExperienceLocationsTypePickupPoint     ExperienceLocationsType = "pickup_point"
	ExperienceLocationsTypeRedemptionPoint ExperienceLocationsType = "redemption_point"
	ExperienceLocationsTypeEndPoint        ExperienceLocationsType = "end_point"
)
```

but encoded as 
`e.AppendValue(row.Type)`
in corresponding `convertRowsForUpsertExperienceLocationsFixed()` function

Manually changing `AppendValue()` function above to `AppendString()` fixes the issue, so I guess this should be done always in case LOAD DATA INFILE has to insert a string enum value.

### Relevant log output

```shell
can't encode type models.ExperienceLocationsType to TSV
```

### Database schema

```sql
create table experience_locations
(
    location_id   varchar(512)                                                          not null,
    type          enum ('start_point', 'pickup_point', 'redemption_point', 'end_point') not null
)
```

### SQL queries

```sql
-- name: UpsertExperienceLocations :copyfrom
REPLACE INTO experience_locations (location_id, type)
VALUES (?, ?);
```

### Configuration

```yaml
version: "2"

sql:
  - engine: mysql
    schema: migrations
    queries: queries
    gen:
      go:
        package: models
        sql_package: "database/sql"
        sql_driver: "github.com/go-sql-driver/mysql"
        out: models
```

### Playground URL

https://play.sqlc.dev/p/f8a7a9571876a2c4956bfed6b81df52da25073cc2975700512372529e8b5502b

macOS

MySQL

Go

fzlzjerry:
Fixes #4324.

Generated non-null enums are defined string types. Passing them to `mysqltsv.Encoder.AppendValue` fails at runtime because they are neither a built-in string nor a `driver.Valuer`. The generated code compiles, but LOAD DATA cannot consume its rows.

Recognize the generated Go enum names, including the qualifier used for split models, and emit `AppendString(string(value))`. Nullable enum wrappers already implement `driver.Valuer`, so they keep the generic path. Type overrides are checked by their resolved Go type, not merely their SQL column type.

The production change is 11 added lines. The remaining files are CLI inputs and generated goldens for default arguments, parameter-struct pointers and split models. They cover multi-column and scalar enums, nullable values, an overridden enum, ordinary strings and INSERT/REPLACE conversion. Existing goldens, models, dependencies and public APIs are unchanged.

Validation:
- The final CLI golden fails on the clean base and passes in both base and managed-db contexts, also with `-race`.
- Real MySQL 9.7.2: the original generated code fails nine enum cases and passes twelve controls. All 21 cases pass after the fix, including exact stored values/counts and zero `SHOW WARNINGS`. Checked with `-race` on Go 1.26.8 and 1.27.0.
- Full tagged suite with the CI's Go 1.26.5, `CGO_ENABLED=0`, live PostgreSQL 16.14/MySQL 9.7.2 and all required process plugins: 6,889 passing test observations, zero failures. The 153 declared skips are unsupported replay contexts and one statement with no faithful rendering; none are missing tools.
- Full generated-code corpus builds. All six CI build targets (Linux/macOS/Windows, amd64/arm64) build with CGO disabled. Native build and vet pass on Go 1.26.8 and 1.27.0; generator race tests and formatting pass.
- `govulncheck` passes on Go 1.26.8. An additional scan on the older build/test compiler, Go 1.26.5, reports five standard-library advisories that also reproduce on the clean base. No toolchain/dependency changes are included.

REPLACE duplicate-key behavior stays with #4559. The runtime REPLACE cases use fresh keys to isolate the enum encoding problem; this PR does not claim to fix that separate semantic issue.


maglicheev-check24:
@fzlzjerry Can confirm that the fix indeed works for my issue, the generated function is the same as the workaround I've developed in my code as a temporary fix

fzlzjerry:
Thanks for confirming this against your affected query, @maglicheev-check24. Good to know the generated function matches your working workaround.

